PR #44, opened 2024-05-24
feat: Add common types used by storage provider

### LINKED ISSUE #42: feat: Add common types library
There are common types that are used by different pallets. Create a library holding these common types. Common types include things such as `Cid`, `Address` and `ActorID`. More common types can be found in the research documents in the [polka-disk repo](https://github.com/eigerco/polka-disk/). These common types will probably need to be extended at some point. For now, implement the types needed to complete #13

aidan46:
# Description

Add `common` lib for types used by different pallets. The `common` library will be extended when needed. This commit adds types that the storage provider pallet needs.

# Implementation

All types have been documented in the code.
Types added:
- `ActorID` -> Identifier for Actors.
- `Cid` -> Identifier for a CID.
- `MinerId` -> Tuple struct holding u32 identifying a miner.
- `Address` -> Defines the protocol and data payload conversion from either a public key or a value.
- `Payload` -> The data of the `Address`.
- `DelegatedAddress` -> A "delegated" (f4) address.

The `AccountIdConversion` trait has been added to implement conversion between `AccountId` and other types. This trait has been implemented for `MinerID`.

Fixes #42

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1488,6 +1488,13 @@ dependencies = [
 [[package]]
 name = "common"
 version = "0.1.0"
+dependencies = [
+ "parity-scale-codec",
+]
+
+[[package]]
+name = "common"
+version = "0.1.0"
 source = "git+https://github.com/w3f/ring-proof#b273d33f9981e2bb3375ab45faeb537f7ee35224"
 dependencies = [
  "ark-ec",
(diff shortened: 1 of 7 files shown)


th7nder (changes requested):
I just don't understand a lot about this PR, honestly :(

Issue #12, opened 2024-05-14
Payment Channel Pallet Implementation


serg-temchenko:
As described in our guidelines in the [`Payment Channel Pallet`](https://github.com/eigerco/polka-disk/blob/main/doc/research/parachain/parachain-implementation.md#payment-channel-pallet) section, the [PaymentChannelActor](https://spec.filecoin.io/#section-glossary.payment-channel) should be ported as a Substrate pallet.

TODO: @aidan46 needs to define the order in which pallets should be implemented and extend the description accordingly.

PR #104, opened 2024-06-28
Payment channel pallet implementation

serg-temchenko:
### Description

This PR introduces the implementation of a Payment Channel pallet, inspired by the [Filecoin Payment Channel Actor](https://github.com/filecoin-project/builtin-actors/blob/master/actors/paych/src/lib.rs). The pallet enables secure off-chain microtransactions between two parties, with on-chain settlement for final balances.

Design page and extrinsics.

Closes Issue #12.

### Checklist

  - [X] If yes, which ones?
    - This PR still requires benchmarks/weights, tests implementation. Also, when off-chain protocol will be implemented, this pallet might require some adjustments.
- [ ] If there are follow-ups, have you created issues for them?
  - [ ] If yes, which ones? / List them here
- [ ] Did you document new (or modified) APIs?


jmg-duarte (changes requested):

th7nder (changes requested):
I didn't go through all of the functions yet, leaving some feedback, thanks for that!

Issue #486, opened 2024-10-29
Runtime: harmonize extrinsic calls emitting events


neutrinoks:
Currently, our pallets do not uniquely emit events. Some emit a single event per extrsinic call, and some emit multiple events per extrinsic call.

**Extrsinsics emitting a single event per call**
- `StorageProvider::register_storage_provider()` -> `StorageProviderRegistered`
- `StorageProvider::slash_sector()` -> `SectorSlashed`
- `StorageProvider::submit_windowed_post()` -> `ValidPoStSubmitted`
- `Market::add_balance()` -> `BalanceAdded`
- `Market::withdraw_balance()` -> `BalanceWithdrawn`
- `Market::settle_deal_payments()` -> `DealsSettled`

**Extrinsics emitting multiple events per call**
- ~`StorageProvider::pre_commit_sectors()` -> `SectorsPreComitted`~ (not the case)
- ~`StorageProvider::prove_commit_sectors()` -> `SectorsProven`~
- ~`StorageProvider::declare_faults()` -> `FaultsDeclared`~
- ~`StorageProvider::declare_faults_recovered()` -> `FaultsRecovered`~
- ~`StorageProvider::terminate_sectors()` -> `SectorsTerminated`~
- ~`Market::publish_storage_deals()` -> `DealPublished`~

So, in my opinion, the extrinsic emitting multiple events should be modified to emit only one event with collected/combined information.

PR #579, opened 2024-11-15
fix(market): batch published deals event

jmg-duarte:
### Description

Batch the published deals events into a single one, saving gas.

Partially addresses #486 

### Checklist


th7nder (approved):

cernicc (approved):

maat/tests/real_world.rs
```diff
@@ -191,14 +191,15 @@ async fn publish_storage_deals<Keypair>(
 
     for event in deal_result
         .events
-        .find::<storagext::runtime::market::events::DealPublished>()
+        .find::<storagext::runtime::market::events::DealsPublished>()
```
aidan46, line 194: Why not use `find_first` here, since it is only looking for a single occurrence?

aidan46, line 194: Why not use `find_first` here? There is only a single deal being passed in.

maat/tests/real_world.rs
```diff
@@ -191,14 +191,15 @@ async fn publish_storage_deals<Keypair>(
...
     {
         let event = event.unwrap();
         tracing::debug!(?event);
 
-        assert_eq!(event.client, alice.account_id().clone().into());
         assert_eq!(event.provider, charlie.account_id().clone().into());
-        assert_eq!(event.deal_id, 0); // first deal ever
+        assert_eq!(event.deals.0.len(), 1);
+        assert_eq!(event.deals.0[0].client, alice.account_id().clone().into());
```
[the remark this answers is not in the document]
jmg-duarte: BoundedVec -> Vec is `.0`

Issue #592, opened 2024-11-19
feat: handle recovered and declares faults in Windowed PoSt


th7nder:
When submitting PoSt currently we're just taking all of the sectors that were assigned to the partition, don't take into the account recovered faults and all.

- https://github.com/filecoin-project/lotus/blob/e7db584232113ce70dbb011dc3d598014e02fff0/storage/wdpost/wdpost_run.go#L351

4. Add support in pipeline generate proofs to generate multiple partitions.
     - Run full PoSt cycle: https://github.com/filecoin-project/lotus/blob/e7db584232113ce70dbb011dc3d598014e02fff0/storage/wdpost/wdpost_run.go#L270 
     - https://github.com/filecoin-project/lotus/blob/e7db584232113ce70dbb011dc3d598014e02fff0/storage/wdpost/wdpost_run.go#L596
     - they need to be ordered by sector number, because the partitions will end up being different!

PR #593, opened 2024-11-19
feat: integrate submit_windowed_post with pallet proofs

### LINKED ISSUE #449: Integrate `verify_submit_windowed_post` with `pallet-proofs` verify PoSt
- https://github.com/filecoin-project/curio/blob/a4886790ce74c77678456b4f6e114c679966e847/tasks/window/compute_do.go#L52 -> this how randomness is selected from beacon from given challenge?
- https://github.com/filecoin-project/builtin-actors/blob/master/actors/miner/src/deadline_info.rs#L27
- https://github.com/filecoin-project/builtin-actors/blob/master/actors/miner/src/types.rs#L118
- https://github.com/filecoin-project/builtin-actors/blob/ca31c6099ccb31fc48e1707df19ace6f12e5429b/actors/miner/src/lib.rs#L4897
- https://github.com/filecoin-project/curio/blob/main/tasks/seal/task_porep.go#L101
- https://github.com/filecoin-project/builtin-actors/blob/ca31c6099ccb31fc48e1707df19ace6f12e5429b/actors/miner/src/lib.rs#L502
- https://github.com/filecoin-project/builtin-actors/blob/ca31c6099ccb31fc48e1707df19ace6f12e5429b/actors/miner/src/lib.rs#L719C21-L719C41
- https://github.com/filecoin-project/builtin-actors/blob/ca31c6099ccb31fc48e1707df19ace6f12e5429b/actors/miner/src/lib.rs#L4738
- https://github.com/filecoin-project/ref-fvm/blob/1e313301b050b9daf99ba0f9121ca4c219a12539/fvm/src/kernel/filecoin.rs#L309
-  https://github.com/filecoin-project/rust-fil-proofs/blob/3c83391d8ad8f64764fe67625f727a19300ee2c1/filecoin-proofs/src/api/window_post.rs#L181

th7nder:
### Description

Fixes #449.

1. It implements the logic in `submit_windowed_post` of Pallet Storage Provider, so it calls Pallet Proofs to verify PoSt. 
2. Exposes Runtime API to fetch the current deadline, so `current_deadline.challenge` will be able to be called in the pipeline #596.

I tested it by running:
1. `examples/rpc_publish.sh examples/test-data-big.car`
2. `export CACHE=<directory outputed in rpc_publish>`
3. `export COMMR=<commr outputed by rpc publish>`
4. `export CHALLENGE_BLOCK=<runtime api query call for current_deadline>` 
5. `./target/release/polka-storage-provider-client proofs post --sr25519-key "//Charlie" --cache-directory $CACHE --proof-parameters-path 2KiB.post.params --sector-number 1 $CACHE/../sealed/1 --challenge_block $CHALLENGE_BLOCK $COMMR`
6. `submit_windowed_post`  via `polkadot.js.org` with the file `1.sector.proof.post.scale`.

image+

```bash
CACHE_FOLDER="/tmp/F8JAt9d/deals_storage/cache"

PROVIDER="//Charlie"
CAR_FILE="../examples/test-data-big.car"
SECTOR_ID=1

COMMR="bagboea4b5abcauws63b3gksuxld4zxk7jlzx776kmh66rokcngcvhtwsuksxpklm"
CHALLENGE_BLOCK="108"
POST_PARAMS_PATH="../2KiB.post.params"

polka-storage-provider-client proofs post \
--sr25519-key "$PROVIDER" \
--proof-parameters-path "$POST_PARAMS_PATH" \
--cache-directory "$CACHE_FOLDER" \
--sector-number "$SECTOR_ID" \
--challenge-block "$CHALLENGE_BLOCK" \
"$CACHE_FOLDER/../sealed/$SECTOR_ID" \
"$COMMR"
```

### Important things

- maat test is broken! I need to fix it, but it's out of the scope as I'm planning bigger changes to it. 

### Checklist

- [X] If there are follow-ups, have you created issues for them?
  - [X] If yes, which ones? / List them here
      - [ ] #592 
      - [x] #595 
      - [x] #596 
      - [x] #615 


pallets/storage-provider/src/lib.rs
```diff
@@ -316,6 +316,7 @@ pub mod pallet {
         /// extrinsic but is not registered as one.
         StorageProviderNotFound,
         /// Emitted when trying to access an invalid sector.
+        InvalidPartition,
         InvalidSector,
```
aidan46: Docs are for the enum member below it.

pallets/storage-provider/src/lib.rs
```diff
@@ -714,6 +720,52 @@ pub mod pallet {
...
+                    let comm_r = Commitment::<CommR>::from_cid_bytes(&sector_info.sealed_cid)
+                        .expect("CommR to be validated on pre-commit");
+                    let _ = replicas
+                        .try_insert(
+                            *sector_number,
+                            PublicReplicaInfo {
+                                comm_r: comm_r.raw(),
+                            },
+                        )
+                        .map_err(|_| Error::<T>::TooManyReplicas);
```
aidan46: I am a bit confused by this. Why are you capturing the result in `_` instead of unwrapping with `?`?

pallets/storage-provider/src/lib.rs
```diff
@@ -1015,6 +1065,36 @@ pub mod pallet {
...
 
     impl<T: Config> Pallet<T> {
+        /// Gets the current deadline of the storage provider.
+        ///
+        /// If there is no Storage Provider of given AccountId returns [`Option::None`].
+        /// May exceptionally return [`Option::None`] when
+        /// conversion between BlockNumbers fails, but technically should not ever happen.
+        pub fn current_deadline(
+            storage_provider: &T::AccountId,
+        ) -> core::option::Option<primitives_proofs::CurrentDeadline<BlockNumberFor<T>>> {
```
aidan46: I'm curious as to why we need to specify `core::option::Option` here. Is it because the `StorageProviderApi` is defined in a crate that supports `std`?

pallets/storage-provider/src/lib.rs
th7nder: `try_insert()` returns the old value if it was there, I don't mind the old value, so I'm discarding it.

Ref: https://paritytech.github.io/polkadot-sdk/master/frame_support/struct.BoundedBTreeMap.html#method.try_insert

aidan46: You are not unwrapping tho, just mapping `Result<T, E>` to `Result<T, F>` and discarding the `Result<T, F>`. Why not return an error like this:
```suggestion
                    replicas
                        .try_insert(
                            *sector_number,
                            PublicReplicaInfo {
                                comm_r: comm_r.raw(),
                            },
                        )
                        .map_err(|_| Error::<T>::TooManyReplicas)?;
```

What you have right now is barely different from:
```suggestion
                    let _ = replicas
                        .try_insert(
                            *sector_number,
                            PublicReplicaInfo {
                                comm_r: comm_r.raw(),
                            },
                        );
```

th7nder: You're right! Thanks for spotting this one, good catch. I need to add `?`.

pallets/storage-provider/src/lib.rs
jmg-duarte: I can't speak for Konrad, but in general, yes. Even though all `std` members that have a `core` counterpart are implemented as basically a rename if your crate supports `no_std` you probably want to use `core` wherever you can instead of `std`. Otherwise, when you try to compile without `std` compilation fails with "missing std"

jmg-duarte: Thinking about is further, I'm not sure the fully qualified name is required here. `Option` is usually imported as part of the prelude (if you run `cargo expand` you'll see what the prelude is) and as such, I would hope the compiler to be smart enough to import `Option` from the right prelude.

th7nder: I don't remember why I added it, but it looks like you're right guys. Works without it, removing.

pallets/storage-provider/src/lib.rs
```diff
@@ -686,10 +688,14 @@ pub mod pallet {
             // Ensure a valid proof size
             // TODO(@jmg-duarte,#91,19/8/24): correctly check the length
```
cernicc: Is this comment still valid?

storagext/lib/src/clients/storage_provider.rs
```diff
@@ -98,9 +101,30 @@ pub trait StorageProviderClientExt {
...
+        account_id: &AccountId32,
+    ) -> impl Future<Output = Result<Option<CurrentDeadline<BlockNumber>>, subxt::Error>>;
 }
 
 impl StorageProviderClientExt for crate::runtime::client::Client {
+    async fn current_deadline(
+        &self,
+        account_id: &AccountId32,
+    ) -> Result<Option<CurrentDeadline<BlockNumber>>, subxt::Error> {
+        let payload = runtime::apis()
```
cernicc, line 116: Good job on figuring this out. I didn't know you can also call the pallet that way.

pallets/storage-provider/src/lib.rs
th7nder: good catch!

storagext/lib/src/clients/storage_provider.rs
th7nder: me neither, it took some time and lots of issues to get this to work :D

cernicc (approved):
Good job 💪

aidan46 (approved):

PR #599, opened 2024-11-22
refactor: Publish storage deals

### LINKED ISSUE #121: Market Pallet: reconsider publish_storage_deals design
https://github.com/eigerco/polka-storage/pull/105#discussion_r1666865574

The issue is that the function fails silently in a lot of cases, by just not emitting the event for the respective deal.
We need to decided if we want to keep this design decision or if all deals must be valid.

Related:
* https://github.com/eigerco/polka-storage/issues/492

### LINKED ISSUE #492: Return better error for insufficient client balance
Return a better error from the `market` pallet when the client's balance is too low. Currently we return `AllProposalsInvalid` but we should be able to return a better one, we have all the info we need, even in the logs:
```
2024-10-31 13:15:18.042 ERROR tokio-runtime-worker runtime::market: [Parachain] invalid deal: client 2c035cee15ad7b9deda3563c9acfdc7b523a4f2bbb084f14424b684b76aceb48 (5D4Qxp9r...) not enough free balance 0 < 650000000000 to cover deal idx: 0
```

Screenshot from 2024-10-31 13-16-03

aidan46:
### Description

Refactor publish storage deals to fail if any deals are invalid and report the error to the user. Publishing storage deals now fails if a single deal is invalid and the error is relayed to the user. The tests have been updated to reflect the change. The success test for publish storage deals needed to be updated as there were not enough funds allocated, making 1 out of 3 deals fail. As a result of this we need to increase the initial balance and adapt the tests to take this into consideration.


pallets/market/src/lib.rs
```diff
@@ -268,9 +272,9 @@ pub mod pallet {
             )
         }
 
-        fn cid(&self) -> Result<Cid, ProposalError> {
+        fn cid(&self) -> Result<Cid, PaddedPieceSizeError> {
             let cid = Cid::try_from(&self.piece_cid[..])
-                .map_err(|e| ProposalError::InvalidPieceCid(e))?;
+                .map_err(|_| PaddedPieceSizeError::InvalidPieceCid)?;
```
jmg-duarte: Why can't this one be 
```suggestion
                .map_err(|_| Error::<T>::InvalidPieceCid)?;
```

pallets/market/src/lib.rs
```diff
@@ -1002,15 +834,15 @@ pub mod pallet {
         ) -> Result<Cid, DispatchError> {
             let pieces = proposals
                 .map(|p| {
-                    let cid = p.cid()?;
-                    let commitment =
-                        Commitment::from_cid(&cid).map_err(|err| ProposalError::CommD(err))?;
+                    let cid = p.cid().map_err(|_| PaddedPieceSizeError::InvalidPieceCid)?;
+                    let commitment = Commitment::from_cid(&cid)
+                        .map_err(|_| PaddedPieceSizeError::UnableToCreateCommD)?;
```
jmg-duarte: Same as the previous one, is it because of Display?

pallets/market/src/error.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+            DealSettlementError::ExpiredDeal => {
+                write!(f, "DealSettlementError: Expired Deal")
+            }
+            DealSettlementError::DealNotActive => {
+                write!(f, "DealSettlementError: Deal Not Active")
+            }
+        }
+    }
+}
```
jmg-duarte: Use Display to implement Debug

pallets/market/src/error.rs
```diff
@@ -0,0 +1,77 @@
...
+    fn from(value: PrimitivesPieceSizeError) -> Self {
+        match value {
+            PrimitivesPieceSizeError::SizeTooSmall => PaddedPieceSizeError::SizeTooSmall,
+            PrimitivesPieceSizeError::SizeNotPowerOfTwo => PaddedPieceSizeError::SizeNotPowerOfTwo,
+            PrimitivesPieceSizeError::NotAMultipleOfNodeSize => {
+                PaddedPieceSizeError::NotAMultipleOfNodeSize
+            }
+        }
+    }
+}
```
jmg-duarte: We can just use the original and convert as needed

primitives/commitment/src/piece.rs
```diff
@@ -197,6 +197,14 @@ impl Into<filecoin_proofs::PaddedBytesAmount> for PaddedPieceSize {
         filecoin_proofs::PaddedBytesAmount(self.0)
     }
 }
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum PaddedPieceSizeError {
+    SizeTooSmall,
+    SizeNotPowerOfTwo,
+    NotAMultipleOfNodeSize,
+}
```
jmg-duarte: Move to the top of the file (maybe)

pallets/market/src/lib.rs
aidan46: We are collecting errors from the commitment here, any error that happens will be logged and returned as an invalid commD error. I would add the error to the error enum in but I need `PalletError` implemented for `PaddedPieceSizeError`. I think logging invalid commd + docs should be enough for the user.

pallets/market/src/error.rs
jmg-duarte, line 1: Much cleaner, nice job!

th7nder (dismissed):

pallets/market/src/lib.rs
th7nder: This is actually not a CommD, it's a Piece CID aka. CommP (it's on a deal, multiple CommPs == CommD)

jmg-duarte (commented):
Fix Konrad's suggestion please

pallets/market/src/lib.rs
aidan46: Good catch! Updated in the latest commit

th7nder (dismissed):

pallets/market/src/lib.rs
```diff
@@ -268,9 +272,9 @@ pub mod pallet {
             )
         }
 
-        fn cid(&self) -> Result<Cid, ProposalError> {
+        fn cid(&self) -> Result<Cid, PaddedPieceSizeError> {
```
cernicc: There is no need for this function. Also the error type is confusing. Use `Commitment::from_cid_bytes`. It does the conversion and it also checks if cid is correct. Hmm. If you want to have a function on the struct then it would be better to have something like `fn piece_commitment(&self) -> Commitment<CommP>`. That is because the piece cid actually encodes the commitment.

Now that I think about it. The `piece_commitment` function is better. That way we are ready if/when we implement TypeInfo for the `Commitment` struct and use that directly on the runtime struct.

pallets/market/src/lib.rs
```diff
@@ -1002,15 +834,15 @@ pub mod pallet {
         ) -> Result<Cid, DispatchError> {
             let pieces = proposals
                 .map(|p| {
-                    let cid = p.cid()?;
-                    let commitment =
-                        Commitment::from_cid(&cid).map_err(|err| ProposalError::CommD(err))?;
+                    let cid = p.cid().map_err(|_| PaddedPieceSizeError::InvalidPieceCid)?;
+                    let commitment = Commitment::from_cid(&cid)
```
cernicc: If you'll add a `piece_commitment` to the proposal you can just use that.

pallets/market/src/lib.rs
```diff
@@ -1041,8 +873,8 @@ pub mod pallet {
                 Self::validate_deal_can_activate(deal, provider, sector_expiry, sector_activation)
                     .map_err(|e| {
                         log::error!(target: LOG_TARGET, "deal {} cannot be activated, because: {:?}", *deal_id, e);
-                        Error::<T>::DealActivationError }
-                    )?;
+                        e
```
cernicc, line 878: There is `inspect_err`

pallets/market/src/lib.rs
```diff
@@ -1150,45 +979,48 @@ pub mod pallet {
...
             current_block: BlockNumberFor<T>,
-        ) -> Result<(), ProposalError> {
+        ) -> Result<(), Error<T>> {
             let encoded = Encode::encode(&deal.proposal);
             log::trace!(target: LOG_TARGET, "sanity_check: encoded proposal: {}", hex::encode(&encoded));
             Self::validate_signature(&encoded, &deal.client_signature, &deal.proposal.client)?;
 
             // Ensure the Piece's Cid is parsable and valid
-            let _ = deal.proposal.cid()?;
+            let _ = deal
```
cernicc: You can also use `piece_commitment` here.

primitives/commitment/src/piece.rs
```diff
@@ -100,6 +100,15 @@ impl Into<filecoin_proofs::UnpaddedBytesAmount> for UnpaddedPieceSize {
     }
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub enum PaddedPieceSizeError {
+    SizeTooSmall,
+    SizeNotPowerOfTwo,
+    NotAMultipleOfNodeSize,
+    InvalidPieceCid,
```
cernicc: The InvalidPieceCid and UnableToCreateCommP should be removed. The piece size has no relevance with the piece commitment.

cernicc (changes requested):

pallets/market/src/lib.rs
jmg-duarte: Adding `inspect_err` here makes it more confusing IMO.

primitives/commitment/src/piece.rs
jmg-duarte: You'll need to be more specific 
image

th7nder (dismissed):

th7nder (dismissed):

cernicc (approved):

th7nder (approved):

PR #600, opened 2024-11-22
fix(maat): proofs

cernicc:
### Description

Fixes maat proofs and adds a script for future use.


runtime/src/configs/mod.rs
```diff
@@ -446,20 +446,27 @@ impl pallet_randomness::Config for Runtime {
...
-        fn random(_: &[u8]) -> (T::Hash, BlockNumberFor<T>) {
+    impl<Output, T> Randomness<Output, BlockNumberFor<T>> for PredictableRandomnessSource<T>
+    where
+        Output: Decode + Default,
+        T: Config,
+    {
+        fn random(subject: &[u8]) -> (Output, BlockNumberFor<T>) {
             (
-                T::Hashing::hash(&[]),
+                Output::decode(&mut TrailingZeroInput::new(subject)).unwrap_or_default(),
```
th7nder, line 467: What's going on here?

cli/polka-storage-provider/client/src/commands/proofs.rs
```diff
@@ -203,17 +217,40 @@ impl ProofsCommand {
...
+                    DomainSeparationTag::SealRandomness,
+                    seal_randomness_height,
+                    &entropy,
+                );
+                println!(
+                    "[{seal_randomness_height}] Ticket randomness: {}",
+                    hex::encode(ticket)
+                );
+
+                let interactive_block_number = pre_commit_block_number + 10;
```
th7nder, line 246: This number is magical, can we reference the pipeline code it's drawing from?

cli/polka-storage-provider/client/src/commands/proofs.rs
```diff
@@ -465,3 +509,14 @@ fn file_with_extension(
...
 }
+
+fn get_randomness(
+    personalization: DomainSeparationTag,
+    block_number: u64,
+    entropy: &[u8],
+) -> [u8; 32] {
+    // This randomness digest is hardcoded because it's always same on testnet.
+    // Check `PredictableRandomnessSource` struct for more details.
+    let digest = [0u8; 32];
```
th7nder, line 521: But it's not [0u8; 32] on the testnet, right?

cli/polka-storage-provider/client/src/commands/proofs.rs
```diff
@@ -72,6 +77,15 @@ pub enum ProofsCommand {
         /// Directory where the proof files and the sector will be put. Defaults to the current directory.
         #[arg(short, long)]
         output_path: Option<PathBuf>,
+        /// Sector number
+        #[arg(long)]
+        sector_id: u32,
+        /// The height at which we draw the randomness for deriving a sealed cid.
```
th7nder, line 83: We should propose the defaults that were hardcoded here, If I were a new user, I wouldn't know where to put it and which number needs to be after which one.

th7nder (changes requested):

runtime/src/configs/mod.rs
cernicc: TrailingZeroInput adds zeros to subject and then decodes that. In our case we are using empty slice for subject. That is why we get [0; 32] randomness when called. Before the slice was hashed. Now it's only decoded.

While fixing the test I found [TestRandomness](https://paritytech.github.io/polkadot-sdk/master/frame_support_test/struct.TestRandomness.html). The latest commit uses that directly. So that we don't need to support our implementation.

cli/polka-storage-provider/client/src/commands/proofs.rs
cernicc: Answered [here](https://github.com/eigerco/polka-storage/pull/600#discussion_r1853982716). It is now :D

cli/polka-storage-provider/client/src/commands/proofs.rs
cernicc: I thought that we will eventually remove those commands? The defaults were also specific for a single use case

jmg-duarte:
Looks good to me, but lets wait until #599 gets merged

th7nder (approved):

runtime/Cargo.toml
```diff
@@ -146,7 +147,7 @@ std = [
   "xcm-executor/std",
   "xcm/std",
 ]
-testnet = []
+testnet = ["frame-support-test"]
```
th7nder: `"dep:frame-support-test"`?

th7nder (changes requested):

runtime/Cargo.toml
cernicc: Only if testnet flag is enabled. The frame_support_test dependency is used. It is used because of [TestRandomness](https://paritytech.github.io/polkadot-sdk/master/frame_support_test/struct.TestRandomness.html)

cli/polka-storage-provider/client/src/commands/proofs.rs
```diff
@@ -72,6 +77,15 @@ pub enum ProofsCommand {
         /// Directory where the proof files and the sector will be put. Defaults to the current directory.
         #[arg(short, long)]
         output_path: Option<PathBuf>,
+        /// Sector number
+        #[arg(long)]
+        sector_id: u32,
```
aidan46, line 82: Is sector number and sector id used interchangeably? I am asking cause we use `SectorId` from [storage_proofs_core](https://docs.rs/storage-proofs-core/latest/storage_proofs_core/sector/struct.SectorId.html) and `SectorNumber` as our internal type.

cernicc: Currently, they are used interchangeably. A quick search over the workspace shows 38 results where `sector_id` is used. We should probably unify that.

th7nder (approved):

jmg-duarte (approved):
LGTM, good job!

PR #611, opened 2024-11-25
fix(storagext): use subxt retry mechanism instead of hand written

jmg-duarte:
### Description

As written in the title.

If we didn't release already I'd consider making this fixed and use the exponential backoff only.

### DIFF
--- a/cli/polka-storage/storagext/src/runtime/client.rs
+++ b/cli/polka-storage/storagext/src/runtime/client.rs
@@ -36,9 +36,6 @@ pub struct Client {
 
 impl Client {
     /// Create a new [`RuntimeClient`] from a target `rpc_address`.
-    ///
-    /// By default, this function does not support insecure URLs,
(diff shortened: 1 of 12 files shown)


cli/polka-storage/storagext/src/runtime/client.rs
```diff
@@ -41,30 +47,18 @@ impl Client {
     ) -> Result<Self, subxt::Error> {
         let rpc_address = rpc_address.as_ref();
 
-        let mut current_retries = 0;
-        loop {
-            let client = if cfg!(feature = "insecure_url") {
```
cernicc, line 46: I think this was the only place where we used `insecure_url` feature. Maybe we can also remove it entirely

jmg-duarte: Totally forgot about that, agreed

th7nder (approved):

cernicc (approved):

PR #612, opened 2024-11-26
feat(sp): Add batch events

### LINKED ISSUE #597: Batch extrinsic events if possible
**Market**
* `publish_deals` — https://github.com/eigerco/polka-storage/pull/579

**Storage Provider**
* `check_deadlines` — may emit multiple `PartititionFaulty`, however, since partitions contain sectors this may be acceptable, furthermore, it's used by the `on_finalize` hook which does not consume "gas" so batching this may not be worth the effort
* `check_precommited_sectors` — may emit multiple `SectorSlashed`, while it is also called from the `on_finalize` hook, here the sector is a single unit, so if multiple sectors for a single owner are slashed the same amount of events will be emitted. I think that batching here will help downstream users when consuming events and reduce "spamming" the network, plus the change should be simple enough

aidan46:
### Description

This PR adds batch events in the SP pallet where applicable.

`check_deadlines` - This hook previously emitted an event for every faulty partition for every storage provider. It has been refactored to be a single event, `PartitionsFaulty` for every storage provider, collecting the faulty partition data along the way. 
`check_precommited_sectors` - This hook previously emitted an event for every sector slashed for every storage provider. It has been refactored to be a single event, `SectorsSlashed`, for every storage provider, collection slashing data along the way.

The tests, documentation, and CLI have been updated to reflect this change.
The SCALE file with the metadata has been updated after this change.


storagext/lib/src/runtime/display/storage_provider.rs
```diff
@@ -87,12 +87,17 @@ impl std::fmt::Display for Event {
...
+            Event::SectorsSlashed {
                 owner,
-                sector_number,
+                sector_numbers,
             } => f.write_fmt(format_args!(
                 "Sector Slashed: {{ owner: {}, sector_number: {} }}",
-                owner, sector_number,
+                owner,
+                itertools::Itertools::intersperse(
+                    sector_numbers.iter().map(|sector| format!("{}", sector)),
```
jmg-duarte: `format!("{}", sector)` implies that `type(sector): Display` which implies that `type(sector): ToString` as such you may find that the following is clearer
```suggestion
                    sector_numbers.iter().map(ToString::to_string),
```
If you like it, apply to the others

pallets/storage-provider/src/lib.rs
```diff
@@ -293,10 +298,10 @@ pub mod pallet {
             recoveries: BoundedVec<RecoveryDeclaration, ConstU32<DECLARATIONS_MAX>>,
         },
         /// Emitted when an SP doesn't submit Windowed PoSt in time and PoSt hook marks partitions as faulty
-        PartitionFaulty {
+        PartitionsFaulty {
             owner: T::AccountId,
-            partition: PartitionNumber,
-            sectors: BoundedBTreeSet<SectorNumber, ConstU32<MAX_SECTORS>>,
+            // No need to be bounded as we are creating the sets ourselves
+            faulty_partitions: BTreeMap<PartitionNumber, BTreeSet<SectorNumber>>,
```
jmg-duarte: I don't get this comment + the bound has important information. Can you elaborate?

aidan46: The data we are returning is retrieved from data that is bounded. I didn't see a reason to have an upper limit on these and enforce more error handling because of the bound.

jmg-duarte: I get the point but the bound also conveys information to the user, even if little since the bound is "all sectors".

I found a counter example to your suggestion from the official SDK:
https://github.com/paritytech/polkadot-sdk/blob/8216235b480630c5a69636b172f5146711c88c9a/substrate/frame/uniques/src/lib.rs#L352-L357

In general, Polkadot states that best practice is to have the bound so I'm not quite sure if we should keep it or not.

cernicc: In this case, I would also probably use BoundedVec. It's a substrate way. But If I am being honest, I am not sure exactly what are the benefits for using the bounded collections in events. Except conveying information about the max size.

docs/src/architecture/pallets/storage-provider.md
```diff
@@ -326,9 +326,9 @@ The Storage Provider Pallet emits the following events:
     - `sector_number` - The sector number that is proven.
     - `partition_number` - The partition number the proven sector is in.
     - `deadline_idx` - The deadline index assigned to the proven sector.
-- `SectorSlashed` - A previously pre-committed sector, but not proven, has been slashed by the system because it has expired.
+- `SectorsSlashed` - Previously pre-committed sectors, but not proven, have been slashed by the system because it has expired.
   - `owner` - SS58 address of the storage provider.
-  - `sector_number` - The sector number that has been slashed because of expiry.
+  - `sector_numbers` - The sector numbers that hav been slashed because of expiry.
```
th7nder:
```suggestion
  - `sector_numbers` - The sector numbers that have been slashed because of expiry.
```

pallets/storage-provider/src/lib.rs
th7nder: If polka has it bounded, let's make it bounded as well. Let's be consistent (I guess the tradeoff is the code with this not-funny errors to be handled when inserting into bounded collection)

pallets/storage-provider/src/lib.rs
```diff
@@ -1231,9 +1238,9 @@ pub mod pallet {
                 }
 
                 log::info!(target: LOG_TARGET, "block: {:?}, checking storage provider {:?} deadline: {:?}",
-                   current_block,
-                   storage_provider,
-                   current_deadline.idx,
+                    current_block,
```
th7nder, line 1240: formatting or what heppened here?

th7nder (changes requested):

pallets/storage-provider/src/lib.rs
aidan46: Yeah, this is formatting. Not sure why this wasn't formatted before 🤷

th7nder (approved):

cernicc (approved):

PR #613, opened 2024-11-26
chore: Replace static str errors

aidan46:
### Description

When I was refactoring the errors for the market pallet, I noticed there we some `Result`'s that returned a static str as the error. This small change adds proper errors for those.  static str's cannot be returned as errors in pallet, this future-proofs the error were we to use them in pallets.

### DIFF
--- a/primitives/proofs/Cargo.toml
+++ b/primitives/proofs/Cargo.toml
@@ -9,16 +9,17 @@ version = "0.1.0"
 
 [dependencies]
 cid = { workspace = true, default-features = false, features = ["alloc"] }
-clap = { workspace = true, features = ["derive"], optional = true }
 codec = { workspace = true, default-features = false, features = ["derive"] }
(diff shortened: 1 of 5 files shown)


primitives/proofs/Cargo.toml
jmg-duarte, line 1: Nit: separate the optional from the non-optional with a newline so taplo doesn't bundle them
Reasoning: becomes easier to understand what is always available vs what isn't

jmg-duarte (commented):
LGTM but if you changed the primitives I'm fairly sure you need to update the SCALE

aidan46:
> LGTM but if you changed the primitives I'm fairly sure you need to update the SCALE

I thought the same thing, but after running generate-scale against the testnet, no changes to the SCALE file. I think this is because these errors are not used in pallets (yet)

th7nder (approved):

jmg-duarte (approved):

PR #616, opened 2024-11-28
chore: combine primitives

cernicc:
### Description

Combine primitives to a single create. Move `storage-provider` to the root of the project.


th7nder (dismissed):
I don't have a strong opinion on `primitives` as not sure we splitted them in first place, I guess consolidation now makes sense. If we need to, we'll split it later. That's fine.

About getting rid of the `cli` directory, 100% on-board. I didn't make much sense to have it there.

th7nder (approved):

aidan46 (approved):

PR #619, opened 2024-11-29
docs: add a point about ubuntu 24.04

th7nder:
### Description

Addresses some [feedback](https://forum.polkadot.network/t/polkadot-native-storage-updates/7021/51?u=th7nder)
from polkadot forum.

### DIFF
--- a/docs/src/getting-started/building/source.md
+++ b/docs/src/getting-started/building/source.md
@@ -3,11 +3,17 @@
 This guide will outline how to setup your environment to build the Polka Storage parachain,
 we cover how to build the binaries directly on your system or using [Nix](https://nixos.org/download/) to ease the process.
 
+<div class="warning">
+Not all of the binaries can be built from polka-storage repository!
+We depend on the <a href="https://github.com/paritytech/zombienet/releases/tag/v1.3.116">zombienet</a> and <a href="https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-stable2409-2">polkadot, polkadot-prepare-worker, polkadot-execute-worker</a> binaries which need to be downloaded regardless.
+</div>
+
 * [Get the code](#get-the-code)
 * [System dependencies](#system-dependencies)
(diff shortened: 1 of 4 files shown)


cernicc (dismissed):

aidan46 (dismissed):

docs/src/getting-started/building/source.md
```diff
@@ -3,12 +3,18 @@
 This guide will outline how to setup your environment to build the Polka Storage parachain,
 we cover how to build the binaries directly on your system or using [Nix](https://nixos.org/download/) to ease the process.
 
+<div class="warning">
+Not all of the binaries can be built from polka-storage repository!
```
jmg-duarte:
```suggestion
Not all of the binaries can be built from the polka-storage repository!
```

docs/src/getting-started/building/source.md
```diff
@@ -3,12 +3,18 @@
...
+We depend on the <a href="https://github.com/paritytech/zombienet/releases/tag/v1.3.116">zombienet</a> and <a href="https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-stable2409-2">polkadot, polkadot-prepare-worker, polkadot-execute-worker</a> binaries which need to be downloaded regardless.
```
jmg-duarte: Doesn't nix solve this? If yes, we need to mention that so that users don't go and download everything and then hit conflicts

jmg-duarte (commented):
Should we also add the glibc thing to the releases?

docs/src/getting-started/building/source.md
th7nder: Nope, if someone doesn't use nix, it does not. Looks like the feedback we received was not using nix.

th7nder: added a (if not using nix)

jmg-duarte (approved):

cernicc (approved):

PR #620, opened 2024-12-02
docs: Polka-index research rfc

aidan46:
### Description

This PR adds an RFC for the libp2p Peer ID to multiaddr resolver. This application will be used by clients to find the multiaddr associated with a Peer ID of a deal to fetch the file from the storage provider (using the polka-fetch component) I followed the guidelines described in [rfc7322](https://www.rfc-editor.org/rfc/rfc7322). 

A follow-up PR will be created for the scraper that collects on-chain information about deals (both historical and current).


jmg-duarte:
Note, we should keep the RFCs in the same place instead of scattered around the workspace

polka-index/RFC.md
```diff
@@ -0,0 +1,134 @@
...
+Polka-index is designed to scale across a growing number of storage providers and users:
+
+- mDNS Discovery: Enables seamless detection of storage providers as they join or leave the network.
+- Dialing: Ensures efficient, direct communication between the Polka-Indexer and storage providers.
+- Caching: Polka-Indexer caches CID mappings for faster responses to frequent queries.
+- HTTP API: Allows users to interact with the system easily, with minimal overhead.
+
+### 7. References
+
+[RFC 7322: Guidelines for RFC Style](https://www.rfc-editor.org/rfc/rfc7322)
```
jmg-duarte: This isn't really a reference

polka-index/RFC.md
```diff
@@ -0,0 +1,132 @@
+# RFC: Polka-index
+
+## Abstract
+
+This document describes Polka-index, a indexing application that maps Content Identifiers (CIDs) to the Peer IDs of storage providers in a peer-to-peer (p2p) network. Users interact with Polka-index through a simple HTTP API, while Polka-Indexer communicates with storage providers over the p2p network using [libp2p](https://docs.libp2p.io/)'s dialing mechanism. This document outlines the system's architecture, key components, and communication protocols.
+
+## 1. Introduction
+
+In the polka-storage system it can be hard to find which CID is owned by which storage provider. Polka-index solves this problem by connecting CIDs (unique identifiers for files) to the Peer IDs of storage providers who store them.
```
jmg-duarte: Is it only `CID -> PeerId`? Or are we also indexing deals and offering search on other fields?

polka-index/RFC.md
```diff
@@ -0,0 +1,132 @@
...
+- **Dialing**: The process in [libp2p](https://docs.libp2p.io/) where one peer establishes a direct connection to another peer using its Peer ID.
+
+## 3. System Overview
+
+### 3.1 Components
+
+#### Storage Providers Component
+
+- Connect to the p2p network using unique Peer IDs.
+- Broadcast their Peer IDs using mDNS so they can be discovered by the Polka-Indexer.
```
jmg-duarte: This works locally, but how would it work between nodes in two different networks?

polka-index/RFC.md
```diff
@@ -0,0 +1,132 @@
...
+- Maintain a Deal Database containing the CIDs and metadata for the files they store.
+- Respond to Polka-Indexer queries via dialing.
```
jmg-duarte: I don't understand what this means. 

I know that "dial" is the connection mechanism/metaphor for libp2p, but the answer isn't done on connection, is it? It *requires a connection* but just establishing a connection (dialing) doesn't answer queries.

polka-index/RFC.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### Polka-Index Component
+
+- Connects to the p2p network using its own unique Peer ID.
+- Detects storage providers through mDNS broadcasts.
+- Dials storage providers directly to query for CID mappings.
+- Maintains a CID-to-Peer ID mapping database.
```
jmg-duarte: Are we using RocksDB, SQLite, Postgres, etc?

We can also make it somewhat backend agnostic, but its important that we at least recognize that need.

polka-index/RFC.md
```diff
@@ -0,0 +1,132 @@
...
+
+- A communication layer powered by [libp2p](https://docs.libp2p.io/) that supports Peer ID discovery (via mDNS) and direct connections (via dialing).
+Users
+- Connect to the Polka-Indexer via HTTP to find out which Peer ID is associated with a specific CID.
+
+### 3.2 How It Works
+
+1. Storage providers connect to the p2p network and broadcast their Peer IDs using mDNS.
+2. Polka-Indexer listens for these broadcasts and keeps track of active storage providers.
+3. Users send an HTTP request to the Polka-Indexer, asking for the Peer ID associated with a specific CID.
```
jmg-duarte: Are we providing a JSON or JSON-RPC API?

polka-index/RFC.md
```diff
@@ -0,0 +1,132 @@
...
+4. If the CID isn’t already in its database, the Polka-Indexer dials storage providers directly to ask for the mapping.
```
jmg-duarte: This seems like *a lot* of requests going out. Shouldn't we rather "subscribe" to the storage providers and get notified of new deals?

polka-index/RFC.md
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+### 6. Scalability and Performance
+
+Polka-index is designed to scale across a growing number of storage providers and users:
+
+- mDNS Discovery: Enables seamless detection of storage providers as they join or leave the network.
+- Dialing: Ensures efficient, direct communication between the Polka-Indexer and storage providers.
+- Caching: Polka-Indexer caches CID mappings for faster responses to frequent queries.
```
jmg-duarte: Why would we do this right now? Committing to adding a cache will increase complexity at little to no benefit.

jmg-duarte (commented):
This RFC is still lacking on several points.

The heavy reliance on mDNS is risky and I don't see how it works across the internet since mDNS was built for local networks ([1](https://docs.libp2p.io/concepts/discovery-routing/mdns/), [2](https://www.rfc-editor.org/rfc/rfc6762)). Am I missing something?

My suggestion for this is twofold: 
* Place SPs and Indexers in the same [overlay network](https://docs.libp2p.io/concepts/appendix/glossary/#overlay) and have the indexers be [well known nodes that are able to bootstrap peers](https://docs.libp2p.io/concepts/appendix/glossary/#boot-node), or we could just have a separate bootstrap node so that Delia can also resolve PeerIDs.

There's also a constant mention of "dialing" but I had to look up what that actually meant, it's important to define the terms that were handed to us by libraries/frameworks — in this case, libp2p.

While I'm not expecting to have the RFC cover small details, the API format (JSON, JSON-RPC, etc) isn't to be overlooked, and neither is the approach to the database.

Other notes:
* Polka-Index or Polka-Indexer, make it consistent
* p2p -> P2P
* This format is a bit verbose, there's almost duplication but not really and it makes it harder to read than I think it should

polka-index/RFC.md
aidan46: You are right, I tested this locally, but it does not support it over different networks. We should use a libp2p [pubsub](https://github.com/libp2p/specs/blob/master/pubsub/README.md) protocol instead.

polka-index/RFC.md
aidan46: The original proposal says the polka-index maps CIDs to storage providers, including other metadata such as supported protocols. The supported protocols are not relevant in our case as they all use the same protocol. We could, of course, extend the functionality to also index deals and offer search on other fields.

polka-index/RFC.md
```diff
@@ -0,0 +1,131 @@
...
+
+In the polka-storage system it can be hard to find which CID is owned by which storage provider. Polka-index solves this problem by connecting CIDs (unique identifiers for files) to the Peer IDs of storage providers who store them.
+
+Storage providers store files outside the scope of the P2P network. Instead, they use the network to broadcast their Peer IDs and respond to CID queries from the Polka-Index. Users can easily request CID-to-Peer ID mappings from the Polka-Index via HTTP.
+
+This document explains how the system works, detailing how users, the Polka-Index, and storage providers interact in the network.
+
+## Terminology
+
+- **CID (Content Identifier)**: A unique identifier for a file stored by a storage provider.
```
cernicc: Is this the root cid of the DAG?

polka-index/RFC.md
```diff
@@ -0,0 +1,131 @@
...
+- **Dialing**: The process in [libp2p](https://docs.libp2p.io/) where one peer establishes a direct connection to another peer using its Peer ID.
+
+## 3. System Overview
+
+### 3.1 Components
+
+#### Storage Providers Component
+
+- Connect to the P2P network using unique Peer IDs.
+- Broadcast their Peer IDs using pubsub so they can be discovered by the Polka-Index.
```
cernicc: How will the peer discovery be done?

polka-index/RFC.md
```diff
@@ -0,0 +1,131 @@
...
+- A communication layer powered by [libp2p](https://docs.libp2p.io/) that supports Peer ID discovery (via pubsub) and direct connections (via dialing).
+Users
+- Connect to the Polka-Index via HTTP to find out which Peer ID is associated with a specific CID.
+
+### 3.2 How It Works
+
+1. Storage providers connect to the P2P network and broadcast their Peer IDs using pubsub.
+2. Polka-Index listens for these broadcasts and keeps track of active storage providers.
+3. Users send an HTTP request to the Polka-Index, asking for the Peer ID associated with a specific CID.
+4. If the CID isn’t already in its database, the Polka-Index dials storage providers directly to ask for the mapping.
```
cernicc: How should this direct call be done?

polka-index/RFC.md
```diff
@@ -0,0 +1,131 @@
...
+- **Peer ID**: Each storage provider generates a Peer ID when it connects to the P2P network. This ID is broadcast via pubsub for discovery.
+- **Deal Database**: A lightweight database where storage providers keep records of CIDs and file metadata. File storage itself is external to the P2P network.
+- **Dialing**: Responds to direct queries from the Polka-Index to provide CID-to-Peer ID mappings.
+
+#### Polka-Index Architecture
+
+- **Peer ID**: The Polka-Index has its own Peer ID for identifying itself in the P2P network.
+- **pubsub Discovery**: It listens for Peer ID broadcasts to discover active storage providers.
+- **CID Mapping Database**: Stores CID-to-Peer ID mappings retrieved from storage providers. This database powers the HTTP API for user queries.
+- **HTTP API**: Provides a simple interface for users:
```
cernicc: What happens when users store the same file on multiple different providers? Should we support that at this time?

polka-index/RFC.md
jmg-duarte: I'd assume it's the CID in the deal

polka-index/RFC.md
jmg-duarte: We should, there's nothing stopping the user from doing that

cernicc: In this case, it might be better to stream the "fetch candidates". The indexer might also want to find out if there are any new providers storing the data. You can't be sure that you have the best ones cached.

polka-index/RFC.md
```diff
@@ -0,0 +1,122 @@
+# RFC: Polka-index
```
th7nder: Who's running those nodes? I don't see any incentive to run them.
What about running indexing as part of the storage provider node - each storage provider is also an indexer?

polka-index/RFC.md
```diff
@@ -0,0 +1,122 @@
...
+- A communication layer powered by [libp2p](https://docs.libp2p.io/) that supports Peer ID discovery (via [libp2p's identify protocol](https://github.com/libp2p/specs/blob/master/identify/README.md)) and subscribe to topics (via [libp2p pubsub protocol](https://github.com/libp2p/specs/blob/master/pubsub/README.md)).
+Users
+- Connect to the Polka-Index via HTTP to find out which Peer ID is associated with a specific CID.
+
+### 3.2 How It Works
+
+1. Storage providers connect to the P2P network and broadcast their Peer IDs and CIDs using pubsub.
+2. Polka-Index listens for these broadcasts and keeps track of active storage providers.
+3. Users send an HTTP request to the Polka-Index, asking for the Peer ID associated with a specific CID.
+4. If the CID isn’t already in its database, the Polka-Index broadcasts a [pubsub](https://github.com/libp2p/specs/blob/master/pubsub/README.md) message asking which storage provider owns said CID.
```
th7nder: What about CIDs expiration? Deals data is stored as long as the sector.

RFCs/polka-index.md
jmg-duarte, line 1: nit: make `RFCs` -> `rfcs` and add `001` to the current document

th7nder, line 1: `rfcs/001-indexer.md`?

RFCs/polka-index.md
```diff
@@ -0,0 +1,79 @@
+# RFC-001: Polka-index
```
th7nder: Shall we just call it Indexer/Storage Indexer/Polka Storage Indexer? And let's use the name consistently (on the diagrams, with capitalization and all). 

Why? `Polka-index` makes it feel bigger than it is.

RFCs/polka-index.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Abstract
+
+This document describes Polka-index, a indexing application that maps Content Identifiers (CIDs) to the Peer IDs of storage providers in a peer-to-peer (P2P) network. Users interact with Polka-index through a simple JSON API, while Polka-Index communicates with storage providers over a P2P network using [libp2p](https://docs.libp2p.io/). This document outlines the system's architecture, key components, and communication protocols.
```
th7nder: Should we add the design of the endpoint that we're going to expose?
And what endpoint we need to implement on Storage Provider?

RFCs/polka-index.md
```diff
@@ -0,0 +1,79 @@
...
+
+- **CID (Content Identifier)**: A unique identifier for a file stored by a storage provider.
+- **Peer ID**: A unique identifier for a node in the P2P network.
+- **Storage Provider**: A node in the network that stores files and manages a database of CIDs but uses the P2P network only to share Peer IDs and provide CID mappings.
+- **Deal Database**: A local database in each storage provider where CIDs and metadata about stored files are recorded.
+- **Polka-Index**: The node responsible for querying storage providers and maintaining a database of CID-to-Peer ID mappings. It also provides an JSON API for users.
+
+## System Overview
+
+### Bootstrap Nodes
```
th7nder: Who's gonna run those bootstrap nodes as well as the previous question I asked, who's gonna run the Indexer nodes?
We need 'explanation in text' of our reasoning.

RFCs/polka-index.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Storage Providers
+
+Storage providers are nodes in the Polka-Index ecosystem, responsible for storing files and managing metadata that links Content Identifiers (CIDs) to their respective storage details. While the actual file storage occurs outside the scope of the P2P network, storage providers leverage the network to facilitate content discovery. They broadcast their Peer IDs to the network, making their presence known and enabling Polka-Index to query them for CID mappings.
+
+Each storage provider maintains a local Deal Database, implemented using RocksDB, a high-performance key-value store. Storage providers must notify the network of any new deals and any deals that have expired to ensure that Polka-Index has up to date information. This information exchange is done by sending [gossipsub](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md) messages to topics that the Polka-Index is subscribed to.
```
th7nder: If a new indexer joins late to the network, how will they catch-up with the existing deals?

RFCs/polka-index.md
```diff
@@ -0,0 +1,79 @@
...
+
+To enhance the robustness and connectivity of the network, storage providers can optionally act as bootstrap nodes. As bootstrap nodes, they provide initial points of contact for new peers joining the network, helping in peer discovery.
```
th7nder: How do we make them 'optionally act' as bootstrap node?

th7nder (changes requested):

RFCs/polka-index.md
aidan46: I am okay with that, polka-index doesn't feel like a good name to me either. I prefer `Indexer` or `Storage Indexer`. How does @eigerco/polka-storage feel about this?

RFCs/polka-index.md
aidan46: I think storage providers have an incentive to run the bootstrap nodes because if they run a well-known bootstrap node, it is good advertising to them for storage clients, making them seem more trustworthy as they run a service that helps clients discover where content is located. When running a bootstrap node, there is an incentive to run an indexer node, because they want content to be discovered inside the P2P network.

RFCs/polka-index.md
aidan46: You make a good point. There should be a message for indexer nodes joining the network and requesting all the deal information after identifying themselves in the network.

RFCs/polka-index.md
aidan46: I should maybe rephrase this as to why storage providers have an incentive to run a bootstrap node, as I mentioned in your previous comment about who runs bootstrap and indexer nodes.

RFCs/polka-index.md
aidan46: I'll add a design for the HTTP endpoint that we need in the indexer node. We do not need to add a HTTP endpoint to the storage provider server, the storage provider server functionality needs to be extended to share deal information with the P2P network.

RFCs/polka-index.md
jmg-duarte: `Storage Indexer` is half a misnomer, since you're not indexing the storage itself, but rather the Storage Providers.

`Indexer` is too general.

`Storage Network Indexer`, `Provider Network Indexer`, or similar would be better, bonus points if you can come up with an acronym that is also a name 😅 

Either way, `Polka Index` was the name picked for us, we MUST make it clear that it changed

RFCs/polka-index.md
jmg-duarte: I don't think we need to expose an API in the RFC, if we do, it should be an example and not something "set in stone"

RFCs/polka-index.md
jmg-duarte: Seems reasonable. But open an "incentives" section that covers the indexer and bootstrap

RFCs/001-indexer.md
```diff
@@ -0,0 +1,84 @@
+# RFC-001: Polka-index
+
+## Abstract
+
+This document describes Polka-index, a indexing application that maps Content Identifiers (CIDs) to the Peer IDs of storage providers in a peer-to-peer (P2P) network. Users interact with Polka-index through a simple JSON API, while Polka-Index communicates with storage providers over a P2P network using [libp2p](https://docs.libp2p.io/). This document outlines the system's architecture, key components, and communication protocols.
```
jmg-duarte: For the link, use the markdown reference format

RFCs/001-indexer.md
```diff
@@ -0,0 +1,84 @@
...
+
+This document explains how the system works, detailing how users, the Polka-Index, and storage providers interact in the network.
+
+## Terminology
+
+- **CID (Content Identifier)**: A unique identifier for a file stored by a storage provider.
+- **Peer ID**: A unique identifier for a node in the P2P network.
+- **Storage Provider**: A node in the network that stores files and manages a database of CIDs but uses the P2P network only to share Peer IDs and provide CID mappings.
+- **Deal Database**: A local database in each storage provider where CIDs and metadata about stored files are recorded.
+- **Polka-Index**: The node responsible for querying storage providers and maintaining a database of CID-to-Peer ID mappings. It also provides an JSON API for users.
```
jmg-duarte: Bold is not needed

RFCs/001-indexer.md
jmg-duarte, line 1: Please, split the longer sentences. It makes it easier to provide suggestions too

RFCs/001-indexer.md
```diff
@@ -0,0 +1,84 @@
...
+
+Storage providers and bootstrap node operators have an incentive to run an indexer node. By running an indexer, these participants contribute directly to the discoverability of content, ensuring that their stored data can be easily located and accessed by clients. This active involvement in indexing enhances their reputation and trustworthiness within the ecosystem, as it demonstrates a commitment to supporting the network's functionality and reliability. For storage providers, running an indexer reinforces their role as dependable custodians of data, while bootstrap node operators strengthen their reputation as key supporters of the network.
```
jmg-duarte: This should be under an "Incentives" section and cover both bootstrap and index nodes

RFCs/001-indexer.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Peer Discovery
+
+The [identify protocol](https://github.com/libp2p/specs/blob/master/identify/README.md#identify-v100) plays a vital role in peer discovery and establishing communication. This protocol allows peers to exchange metadata about themselves upon connection. This metadata includes:
+
+- **Peer ID**: A unique identifier for the peer.
+- **Multiaddrs**: Addresses where the peer can be reached.
+- **Supported Protocols**: A list of protocols the peer supports.
```
jmg-duarte: No need for bold

RFCs/001-indexer.md
```diff
@@ -0,0 +1,84 @@
...
...tes peer discovery by enabling nodes to exchange information upon establishing a connection. When a new peer joins the network, it connects to a well-known bootstrap nodes that assist in the initial discovery process. Upon connecting to a bootstrap node, the [identify protocol](https://github.com/libp2p/specs/blob/master/identify/README.md#identify-v100) is employed to share details such as the node's unique Peer ID, supported protocols, and reachable [multiaddrs (multi-protocol addresses)](https://github.com/libp2p/specs/blob/master/addressing/README.md#multiaddr-in-libp2p). This exchange allows the new peer to learn about the network's topology and available services. The bootstrap node provides information about other peers it knows, facilitating the new node's integration into the network. The bootstrap node keeps up-to-date information about peers through the exchange of the identify messages. This mechanism ensures that nodes can dynamically discover and connect with each other. After the peer discovery is done the node can subscribe to the available [gossipsub](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md) topics to exchange information about deals.
```
jmg-duarte: > and reachable [multiaddrs (multi-protocol addresses)](https://github.com/libp2p/specs/blob/master/addressing/README.md#multiaddr-in-libp2p).

This is not covered in the identify protocol spec, so my question is: is there a source for this information?

Seems to me that you're rather looking for something like [rendezvous](https://github.com/libp2p/specs/blob/master/rendezvous/README.md)

RFCs/001-indexer.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Storage Providers
+
+Storage providers are nodes in the Polka-Index ecosystem, responsible for storing files and managing metadata that links Content Identifiers (CIDs) to their respective storage details. While the actual file storage occurs outside the scope of the P2P network, storage providers leverage the network to facilitate content discovery. They broadcast their Peer IDs to the network, making their presence known and enabling Polka-Index to query them for CID mappings.
+
+Each storage provider maintains a local Deal Database, implemented using RocksDB, a high-performance key-value store. Storage providers must notify the network of any new deals and any deals that have expired to ensure that Polka-Index has up to date information. This information exchange is done by sending [gossipsub](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md) messages to topics that the Polka-Index is subscribed to.
+
+By integrating with the P2P network and facilitating metadata exchange, storage providers play a crucial role in the Polka-index architecture, bridging the gap between decentralized file storage and efficient content lookup.
```
jmg-duarte, line 69: Isn't this repeated from the bootstrap nodes incentives part?

RFCs/001-indexer.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Storage Providers
+
+Storage providers are nodes in the Polka-Index ecosystem, responsible for storing files and managing metadata that links Content Identifiers (CIDs) to their respective storage details. While the actual file storage occurs outside the scope of the P2P network, storage providers leverage the network to facilitate content discovery. They broadcast their Peer IDs to the network, making their presence known and enabling Polka-Index to query them for CID mappings.
+
+Each storage provider maintains a local Deal Database, implemented using RocksDB, a high-performance key-value store. Storage providers must notify the network of any new deals and any deals that have expired to ensure that Polka-Index has up to date information. This information exchange is done by sending [gossipsub](https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md) messages to topics that the Polka-Index is subscribed to.
```
jmg-duarte: RocksDB is an implementation detail, irrelevant for this RFC

RFCs/001-indexer.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Users
+
+Users in the Polka-index network act as requesters of information, leveraging the system to locate and retrieve metadata about content stored by storage providers. They query Polka-Index for mappings between Content Identifiers (CIDs) and the Peer IDs of storage providers. By doing so, users can identify the nodes responsible for storing specific data without directly interacting with the P2P network or storage providers.
+
+Through Polka-Index’s JSON API, users send simple queries containing the CID of interest and receive the associated Peer ID in response. This abstraction streamlines the content discovery process, enabling users to focus on their applications without needing to understand the complexities of the underlying network protocols.
+
+The user’s role is entirely client-facing, relying on Polka-Index as the intermediary to interact with the network, gather deal-related data, and provide accurate CID-to-Peer ID mappings. This design ensures that users experience a straightforward and efficient method for accessing decentralized storage metadata, enhancing accessibility and usability in the network.
```
jmg-duarte: This section is a bit confusing because it obfuscates the user/client relationship when they're actually the same. For reference, the term "user" is a bit too general for this system, client/provider is much clearer

jmg-duarte (commented):
Great improvements since the first draft, but still needs some tuning

RFCs/001-indexer.md
```diff
@@ -0,0 +1,126 @@
+# RFC-001: Polka-index
```
th7nder, line 1: (from the meeting) -> how can we be sure that a separate p2p network doesn't advertise thrash data?

th7nder (changes requested):

RFCs/001-indexer.md
jmg-duarte: Furthermore — can we simply scrape the chain and make the indexer just a fast way of accessing chain data?

PR #624, opened 2024-12-05
feat: submit windowed posts repeatedly

### LINKED ISSUE #621: Schedule Windowed PoSts repeatedly

th7nder:
### Description

It's a *simple*, at least for now, until it works. If it stops working, we can extend it.
Algorithm:
1. When the pipeline starts, schedule a Windowed PoSt for each deadline (send a message to the queue).
2. When the message is polled, it get the **next not elapsed deadline info for an index** from the chain.
3. Waits till a PoSt can be sutmittted for the deadline.
4. When the proof has been submitted, schedule another PoSt message to the pipeline. Start from 2. all over again.

`just testnet && examples/rpc_publish.sh examples/test-data-big.txt`

It keeps submitting a proof and works.

Closes #621. 

### Checklist

- Follow-ups:
    - #625

### DIFF
--- a/pallets/storage-provider/src/deadline.rs
+++ b/pallets/storage-provider/src/deadline.rs
@@ -720,8 +720,7 @@ where
         self.block_number >= self.fault_cutoff
     }
 
-    /// Calculates the deadline information in the next proving period in which the deadline will be open.
-    /// If the current deadline is still open, it'll return the deadline in the next proving period.
+    /// Calculates the deadline information in the next proving period.
     ///
     /// ### Example 1
     /// Current Block = 110
@@ -736,11 +735,7 @@ where
     /// next() -> [240, 260)
     ///
     pub fn next(self) -> Result<Self, GeneralPalletError> {
-        let gap = if self.is_open() {
-            BlockNumber::zero()
-        } else {
-            self.block_number - self.period_start
-        };
+        let gap = self.block_number - self.period_start;
         let delta_periods = BlockNumber::one() + gap / self.w_post_proving_period;
 
         Self::new(
@@ -763,7 +758,6 @@ where
             return Ok(self);
         }
 
-        // has elapsed, advance by some multiples of w_post_proving_period
         self.next()
     }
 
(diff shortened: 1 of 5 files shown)


pallets/storage-provider/src/deadline.rs
```diff
@@ -720,18 +720,27 @@ where
         self.block_number >= self.fault_cutoff
     }
 
-    /// Returns the next deadline that has not yet elapsed.
+    /// Calculates the deadline information in the next proving period in which the deadline will be open.
```
jmg-duarte: I read this multiple times and I still can't figure out the true intent behind this. Can you explain so I can help re-phrase?

pallets/storage-provider/src/deadline.rs
```diff
@@ -745,6 +754,29 @@ where
...
+
+    /// Returns the next deadline that has not yet elapsed.
+    ///
+    /// If the current deadline has not elapsed yet then it returns the current deadline.
+    pub fn next_not_elapsed(self) -> Result<Self, GeneralPalletError> {
+        if !self.has_elapsed() {
+            return Ok(self);
+        }
+
+        // has elapsed, advance by some multiples of w_post_proving_period
```
jmg-duarte: Some? I read `next` and think it's just one.

Maybe there's a better way of writing this

jmg-duarte (dismissed):
Docs are kinda hard to understand in some places but the code looks sound.

I'll leave it up to you / someone else if they agree with my comments

pallets/storage-provider/src/deadline.rs
th7nder: I simplified it.

pallets/storage-provider/src/deadline.rs
th7nder: This comment is unnecessary, I removed it.

It's not just one, cause if for some reason there is:
`[self.open_at, self.close_at] = [100, 120]` and current block = 162, then the next proving period is not `160, 220`, but `240, 260`.

aidan46 (dismissed):
Nice work!

jmg-duarte (approved):

cernicc (approved):

PR #631, opened 2024-12-10
docs: deal scraper RFC

aidan46:
### Description

This PR adds an RFC for the on-chain deal information scraper, which maps Peer IDs to piece_cid's. The application will need to maintain a database that holds these mapping, the database is not mentioned in this RFC but should be considered when implementing. This is an initial draft, I will do some testing to check if this will work.

### DIFF
--- a/rfcs/004-deal-scraper.md
+++ b/rfcs/004-deal-scraper.md
@@ -1,0 +1,42 @@
+# RFC: On-chain deal scraper
+
+## Abstract
+
+This RFC proposes the design and functionality of an application to map `piece_cids` to their respective owners, identified by [Peer ID][1]s, in the Polka-Storage ecosystem.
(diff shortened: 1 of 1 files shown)


rfcs/004-deal-scraper.md
```diff
@@ -0,0 +1,42 @@
+# RFC: On-chain deal scraper
+
+## Abstract
+
+This RFC proposes the design and functionality of an application to map `piece_cids` to their respective owners, identified by [Peer ID][1]s, in the Polka-Storage ecosystem.
```
th7nder: I'm wondering, who needs that from the business perspective at this point.
When a Storage Client is making a deal, they know with whom, so they *should know* where their file is stored.
The only thing they need is to resolve PeerID to a Multiaddress and then connect to the SP, download the file and that's it.

Not sure who the user of it would be.

th7nder (changes requested):

PR #633, opened 2024-12-12
feat(randomness): use BABE's AuthorVRF as a randomness source

### LINKED ISSUE #458: Substrate: Use secure randomness
We are currently using the [insecure randomness ](https://paritytech.github.io/polkadot-sdk/master/pallet_insecure_randomness_collective_flip/index.html) on chain. The reason was faster implementation because if we wanted to use the more secure random algorithm we would also have to switch to BABE consensus algorithm at the time.

This ticket requires you to use the more secure randomness.

Blocked by: 
* #457 

Some more resources:
* https://docs.substrate.io/build/randomness/

jmg-duarte:
### Description

Rewrites `pallet-randomness` to use BABE's AuthorVRF as a partial randomness source.

When running `random(subject)` it will pick up the latest AuthorVRF randomness value and hash it along with the subject.

Closes #458 

### Checklist


pallets/randomness/src/lib.rs
```diff
@@ -6,107 +6,127 @@ pub use pallet::*;
...
-            // Remove old seeds
-            blocks_to_remove.iter().for_each(|number| {
-                SeedsMap::<T>::remove(number);
-            });
+        fn is_inherent(call: &Self::Call) -> bool {
+            matches!(call, Call::set_author_vrf { .. })
         }
     }
 
-    impl<T: Config> Randomness<BlockNumberFor<T>> for Pallet<T> {
```
cernicc, line 100: We should be able to fetch the random value for some specific block. That was solved by caching the values for some period.

jmg-duarte: You can — https://github.com/eigerco/polka-storage/pull/633/files#diff-eface0c86240147709dce013ce11764683db9958eb9a3671ba82034bc2ad0dd4R60-R62 & https://github.com/eigerco/polka-storage/pull/633/files#diff-46b4f3deb1d93fe7de8bd7fe5cfbe34abdcf4a8020f4a203a6d5f6a9cd177c5f

pallets/randomness/src/lib.rs
```diff
@@ -6,107 +6,127 @@ pub use pallet::*;
...
-            }
+    /// The latest author VRF randomness from BABE.
+    #[pallet::storage]
+    #[pallet::getter(fn author_vrf)]
+    pub type AuthorVrf<T: Config> = StorageValue<_, T::Hash, ValueQuery>;
 
-            // Save the seed
-            SeedsMap::<T>::insert(block_number, seed);
-            log::info!(target: LOG_TARGET, "on_initialize: height: {block_number:?}, seed: {seed:?}");
+    /// The last 256 author VRF randomness values from BABE, organized in a ring buffer fashion.
```
aidan46, line 61: Why 256 values? Is this a random selection or a reason that it is exactly 256?

runtime/src/lib.rs
```diff
@@ -661,3 +656,44 @@ impl_runtime_apis! {
         }
     }
 }
+
+// The following code cannot be placed out of this crate because of WASM constraints
+
+/// Storage Key for AuthorVrfRandomness
+const AUTHOR_VRF_STORAGE_KEY: [u8; 32] =
+    hex2array!("1cb6f36e027abb2091cfb5110ab5087fd077dfdb8adb10f78f10a5df8742c545");
```
aidan46, line 670: Where does this key come from?

pallets/randomness/src/lib.rs
jmg-duarte: It's arbitrary but "inspired" by the 256 blocks of history, so this effectively enables a ~512 lookback window with "some tricks"

runtime/src/lib.rs
jmg-duarte: From the Polkadot UI, this is a storage key for the AuthorVRF, it's not part of the [`well_known_keys`](https://paritytech.github.io/polkadot-sdk/master/cumulus_primitives_core/relay_chain/v8/well_known_keys/index.html) module but follows the same idea

pallets/randomness/src/lib.rs
cernicc: Aha. Nice.

cernicc (approved):
Good work 👍

th7nder (approved):

PR #640, opened 2024-12-17
feat: Add peer-resolver lib

### LINKED ISSUE #639: feat: libp2p rendezvous discovery
Implement the libp2p rendezvous discovery protocol on the server (bootstrap) and client-side to share identifying information about peers. The bootstrap node acts as a rendezvous point and the client connects to it to request information about peers. This issue is only for implementing it into a library and supplying examples on how to use it.

aidan46:
### Description

Add the peer-resolver library with bootstrap and client nodes using the rendezvous and identify protocols for peer discovery. Examples for bootstrap, identification and discovery have been added.


lib/peer-resolver/README.md
jmg-duarte, line 1: Line breaks, please

lib/peer-resolver/src/client.rs
```diff
@@ -0,0 +1,154 @@
...
+use crate::error::ResolverError;
+
+/// This struct holds the rendezvous client behaviour for registration and discovery.
+#[derive(NetworkBehaviour)]
+pub struct ClientBehaviour {
+    rendezvous: client::Behaviour,
+}
+
+/// Rendezvous client swarm
+pub struct ClientSwarm {
```
jmg-duarte: `ClientSwarm` is super generic
```suggestion
pub struct RendezvousClientSwarm {
```

lib/peer-resolver/examples/discovery.rs
jmg-duarte, line 1: For an example, this is lacking on the "how and why" department. Can you add more comments explaining the code, almost step by step?

lib/peer-resolver/examples/discovery.rs
```diff
@@ -0,0 +1,80 @@
...
+    loop {
+        tokio::select! {
+            event = swarm.next() => if let Some(event) = event {
+                match event {
+                    // Connection with rendezvous point established
+                    SwarmEvent::ConnectionEstablished { peer_id, .. }
+                        if peer_id == rendezvous_point =>
+                    {
+                        tracing::info!("Connection established with rendezvous point {}", peer_id);
+                        tracing::info!("Connected to rendezvous point, discovering nodes");
```
jmg-duarte:
```suggestion
                        tracing::info!("Connection established with rendezvous point {peer_id}, discovering nodes...");
```

lib/peer-resolver/src/bootstrap.rs
```diff
@@ -0,0 +1,88 @@
...
+                )) => {
+                    if registrations.len() > 0 {
+                        tracing::info!(
+                            "Served peer {} with {} new registrations",
+                            enquirer,
+                            registrations.len()
+                        );
+                    }
+                }
+                _other => {}
```
jmg-duarte: Shouldn't we do at least a debug log?

jmg-duarte (commented):
Preliminary remarks, good job on figuring this out! 🚀

lib/peer-resolver/src/discovery.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Swarm with [`DiscoveryBehaviour`]
+    swarm: Swarm<DiscoveryBehaviour>,
+    /// Rendezvous cookie for continuous peer discovery
+    cookie: Option<Cookie>,
+}
+
+impl DiscoverySwarm {
+    /// Create a new [`DiscoverySwarm`] with the given keypair.
+    /// The given timeout is set for the idle connection timeout
+    pub fn new<K>(keypair_bytes: K, timeout: u64) -> Result<DiscoverySwarm, ResolverError>
```
jmg-duarte: Timeout should be a `Duration`

lib/peer-resolver/src/discovery.rs
```diff
@@ -0,0 +1,91 @@
...
+            .build();
+
+        Ok(DiscoverySwarm {
+            swarm,
+            cookie: None,
+        })
+    }
+
+    /// This functions requests the rendezvous_point for new or updated peers since the last cookie.
+    pub fn discover(&mut self, rendezvous_point: PeerId) {
```
jmg-duarte, line 58: I'm not sure this function name is good, it isn't really descriptive of what it does. 
I know it wraps `discover` but I'm not convinced by libp2p's naming.

lib/peer-resolver/src/register.rs
```diff
@@ -0,0 +1,146 @@
...
+            .map_err(|_| ResolverError::InvalidBehaviourConfig)?
+            .with_swarm_config(|cfg| cfg.with_idle_connection_timeout(Duration::from_secs(timeout)))
+            .build();
+        Ok(RegisterSwarm {
+            swarm,
+            namespace: Namespace::new(namespace)?,
+        })
+    }
+
+    pub async fn register(
```
jmg-duarte, line 68: Docs

lib/peer-resolver/src/register.rs
```diff
@@ -0,0 +1,146 @@
...
+                other => {
+                    tracing::debug!("Unhandled event: {other:?}");
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    pub async fn unregister(&mut self, rendezvous_point: PeerId) {
```
jmg-duarte, line 147: Docs

lib/peer-resolver/src/register.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use crate::error::ResolverError;
+
+#[derive(NetworkBehaviour)]
+pub struct RegisterBehaviour {
+    identify: identify::Behaviour,
+    rendezvous: rendezvous::client::Behaviour,
+}
+
+pub struct RegisterSwarm {
```
jmg-duarte, line 22: Docs

lib/peer-resolver/examples/identify.rs
```diff
@@ -0,0 +1,25 @@
...
+    let rendezvous_point =
+        "12D3KooWDpJ7As7BWAwRMfu1VU2WCqNjvq387JEYKDBj4kx6nXTN".parse::<PeerId>()?;
+    // Results in peer id 12D3KooWK99VoVxNE7XzyBwXEzW7xhK7Gpv85r9F3V3fyKSUKPH5
+    let keypair_bytes = [1; 32];
+
+    let mut swarm = RegisterSwarm::new(keypair_bytes, "rendezvous".to_string(), 10)?;
+
+    swarm
+        .register(rendezvous_point, rendezvous_point_address, Some(MAX_TTL))
+        .await?;
```
jmg-duarte, line 29: Does this simply "stay on"? In other words, when does this function call exit?

jmg-duarte (commented):
One question I have is — what is a cookie in this context?

aidan46:
> One question I have is — what is a cookie in this context?

A cookie is passed into the discover function to only receive peers that have changed or joined since the last discover message was served.

lib/peer-resolver/examples/identify.rs
aidan46: This function call exits as soon as the peer is registered. The peer knows this when it receives the `Registered` event from the rendezvous point.

cernicc (commented):
Try to create a single example of using all these behaviors and you'll see how do you like the usage of the current abstraction. I have a feeling it would be easier to have a single behavior that specifies how the node operates.

lib/peer-resolver/README.md
```diff
@@ -0,0 +1,75 @@
+# Peer Resolver
+
+This library is intended for peer discovery in a libp2p network.
```
th7nder, line 3: Could you remind me what is the expected usage of this library in our storage solution? 
It's kinda hard to imagine for me whether those abstractions are correct, if we write the library first and the usage second.
This looks like a nice example, but not sure whether it will fit/how it will fit in our whole system.

lib/peer-resolver/src/bootstrap.rs
```diff
@@ -0,0 +1,105 @@
...
+                noise::Config::new,
+                yamux::Config::default,
+            )
+            .map_err(|_| ResolverError::InvalidTCPConfig)?
+            .with_behaviour(|key| BootstrapBehaviour {
+                rendezvous: rendezvous::server::Behaviour::new(
+                    rendezvous::server::Config::default(),
+                ),
+                identify: identify::Behaviour::new(identify::Config::new(
+                    "rendezvous-example/1.0.0".to_string(),
```
th7nder: Why `rendevous-example/1.0.0`, what is it supposed to mean?

th7nder (changes requested):

lib/peer-resolver/src/bootstrap.rs
aidan46: This is the protocol information it shares with peers. It should not be rendezvous-example, even though we do not use this, this is an oversight on my end.

aidan46: This is the protocol information it shares with peers. It should not be rendezvous-example, even though we do not use this, this is an oversight on my end.

PR #645, opened 2024-12-18
fix: configure async backing properly

### LINKED ISSUE #644: Fix AuRa panicking on us
This is an issue: https://github.com/OpenZeppelin/polkadot-runtime-templates/issues/134
And we need to solve it.
Here is the solution (most likely):
https://substrate.stackexchange.com/questions/10613/started-getting-aura-pallet-slot-must-increase-error-on-running-collator/10904#10904

th7nder:
### Description

We were having parachain panic, because reasons (`slots must increase`).
The reasons were async backing was half-configured.
I followed:
- https://substrate.stackexchange.com/questions/10613/started-getting-aura-pallet-slot-must-increase-error-on-running-collator/10904#10904
- https://wiki.polkadot.network/docs/maintain-guides-async-backing#phase-1---update-parachain-runtime
-plus reference template: https://github.com/paritytech/polkadot-sdk/blob/master/templates/parachain/runtime/src/configs/mod.rs.

Launched testnet 3 times on the branch, no panics.
Launched testnet 3 times on the develop: panics.

Closes #644.

### Checklist

### DIFF
--- a/runtime/src/configs/mod.rs
+++ b/runtime/src/configs/mod.rs
@@ -26,7 +26,7 @@
 mod xcm_config;
 
 // Substrate and Polkadot dependencies
-use cumulus_pallet_parachain_system::RelayNumberStrictlyIncreases;
+use cumulus_pallet_parachain_system::RelayNumberMonotonicallyIncreases;
 use cumulus_primitives_core::{AggregateMessageOrigin, ParaId};
 use frame_support::{
     derive_impl,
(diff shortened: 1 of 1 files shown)


th7nder:
Cannot wait for finalized block with these changes, investigating.

th7nder:
It actually works, there is another issue with polkadot losing extrinsics when you sent a transaction on a reorg block.

jmg-duarte (approved):

cernicc (approved):

PR #646, opened 2024-12-18
fix(pallet-randomness): bring back the getter API

jmg-duarte:
### Description

I removed the coupling in the previous PR, that was a mistake.
Re-added to fix randomness, I'm not super API with the API but `code that works > pretty code that doesn't`

### Checklist


storage-provider/server/Cargo.toml
```diff
@@ -41,6 +41,7 @@ tokio-util = { workspace = true, features = ["rt"] }
 tower = { workspace = true }
 tower-http = { workspace = true, features = ["trace"] }
 tracing = { workspace = true }
+tracing-appender = "0.2.3"
```
th7nder: probably should be set at the workspace level

th7nder (dismissed):

storage-provider/server/Cargo.toml
jmg-duarte: You're right

th7nder (approved):

cernicc (approved):

Issue #634, opened 2024-12-13
[Tracking] Storage Retrieval


cernicc:
We want to provide a storage retrieving functionality to the client. This issue is used for tracking.

PR #654, opened 2024-12-20
feat: retrieval server & client setup

cernicc:
### Description

Setup the retrieval server and client. They encapsulate the logic needed to successfully transfer files between them. The current blockstore used is in memory and doesn't have a concept of car format. The idea is to swap the server's and client's blockstore to the implementation specific for them.

The example initializes server and client and transfers a single block between them.

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ members = [
   "storage-provider/client",
   "storage-provider/common",
   "storage-provider/server",
+  "storage-retrieval",
   "storage/polka-index",
   "storagext/cli",
   "storagext/lib",
@@ -51,8 +52,10 @@ async-stream = "0.3.6"
 async-trait = "0.1.80"
 axum = "0.7.5"
 base64 = "0.22.1"
+beetswap = "0.4.0"
 bitflags = "2.5.0"
 blake2b_simd = { version = "1.0.2", default-features = false }
+blockstore = "0.7.1"
 bls12_381 = "0.8"
 bs58 = "0.5.1"
 byteorder = "1.5.0"
@@ -78,6 +81,9 @@ ipld-core = "0.4.1"
 ipld-dagpb = "0.2.1"
 itertools = "0.13.0"
 jsonrpsee = { version = "0.24.7" }
+libp2p = "0.54.1"
+libp2p-core = "0.42.0"
+libp2p-swarm = "0.45.1"
 log = { version = "0.4.21", default-features = false }
 multihash-codetable = { version = "0.1.1", default-features = false }
 num-bigint = { version = "0.4.5", default-features = false }
@@ -136,6 +142,7 @@ pallet-market = { path = "pallets/market", default-features = false }
 pallet-proofs = { path = "pallets/proofs", default-features = false }
 pallet-randomness = { path = "pallets/randomness", default-features = false }
 pallet-storage-provider = { path = "pallets/storage-provider", default-features = false }
+polka-index = { path = "storage/polka-index" }
 polka-storage-proofs = { path = "lib/polka-storage-proofs", default-features = false }
 polka-storage-provider-common = { path = "storage-provider/common" }
 polka-storage-runtime = { path = "runtime" }
--- a/storage-retrieval/Cargo.toml
+++ b/storage-retrieval/Cargo.toml
@@ -1,0 +1,32 @@
+[package]
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license-file.workspace = true
+name = "storage-retrieval"
+repository.workspace = true
+version = "0.1.0"
+
+[lints]
+workspace = true
+
+[dependencies]
+anyhow = { workspace = true }
+beetswap = { workspace = true }
+blockstore = { workspace = true }
+cid = { workspace = true }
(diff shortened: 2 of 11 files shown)


storage-retrieval/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+libp2p-core = { workspace = true }
+libp2p-swarm = { workspace = true }
+mater = { workspace = true }
+polka-index = { workspace = true }
+thiserror = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt", "rt-multi-thread", "sync", "time"] }
+tracing = { workspace = true }
+
+[dev-dependencies]
+# multihash = "0.19.3"
```
th7nder, line 29:
```suggestion

```

Cargo.toml
```diff
@@ -136,6 +142,7 @@ pallet-market = { path = "pallets/market", default-features = false }
 pallet-proofs = { path = "pallets/proofs", default-features = false }
 pallet-randomness = { path = "pallets/randomness", default-features = false }
 pallet-storage-provider = { path = "pallets/storage-provider", default-features = false }
+polka-index = { path = "storage/polka-index" }
```
th7nder, line 145: What is it?

storage-retrieval/README.md
th7nder, line 1: empty?

storage-retrieval/examples/simple_retrieval.rs
```diff
@@ -0,0 +1,89 @@
+use std::{str::FromStr, sync::Arc, time::Duration};
+
+use anyhow::Result;
+use blockstore::{
```
th7nder, line 4: nice, didn't know eiger had this library

storage-retrieval/examples/simple_retrieval.rs
```diff
@@ -0,0 +1,89 @@
...
+    // })?);
+
+    // TODO: Blocks should not be hold in memory. Implement blockstore that can
+    // source blocks directly from sectors on disk with the help of an index.
+    let blockstore = Arc::new(InMemoryBlockstore::<64>::new());
+    blockstore.put(StringBlock("12345".to_string())).await?;
+
+    // Setup server
+    let server = Server::new(blockstore)?;
+    let listener: Multiaddr = format!("/ip4/127.0.0.1/tcp/8989").parse()?;
```
th7nder:
```suggestion
    let server_address: Multiaddr = format!("/ip4/127.0.0.1/tcp/8989").parse()?;
```

storage-retrieval/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        .with_tokio()
+        .with_tcp(
+            tcp::Config::default(),
+            noise::Config::new,
+            yamux::Config::default,
+        )?
+        .with_behaviour(|_| Behaviour {
+            bitswap: beetswap::Behaviour::new(blockstore),
+        })
+        .expect("infallible")
```
th7nder, line 44: because?

storage-retrieval/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        .with_swarm_config(|c| c.with_idle_connection_timeout(Duration::from_secs(60)))
```
th7nder, line 45: magic number

storage-retrieval/src/client.rs
```diff
@@ -0,0 +1,189 @@
...
+        for provider in self.providers.clone() {
+            self.swarm.dial(provider)?;
+        }
+
+        // Request the root node of the car file
+        let query_id = self.swarm.behaviour_mut().bitswap.get(&payload_cid);
+        self.queries.insert(query_id, payload_cid);
+
+        // Pin cancellation future
+        pin_mut!(cancellation);
```
th7nder, line 73: Why? What about [CancellationTokens](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html)?

th7nder (changes requested):
I lack completeness in this PR, there is a lot of todos, the functionality is unclear, lots of scenarios are unhandled.
This feels like a prototype for trying out how to transfer a 1 packet via beetswap and libp2p, not something that we can merge into 'production code.'

I'd need to know the strategy on how are we going to implement this whole thing before I agree to merge it. 

E.g. the `server.rs` looks like it'll be gone anyways, or not present in the production code -> because that's something that should be implemented on the Storage Provider Server. 

`client.rs` looks like a candidate to be a separate CLI, cause I suppose we're going to run it like:
`polka-fetch <CID> <STORAGE_PROVIDER_NODE_ADDR>`

I assume if we'd like to start writing production code and merging it, we should:
1. setup CLI `polka-fetch package`  and define it's funcionalities.
2. add libP2P server to storage provider node, so we can `dial it`
3. dial from `polka-fetch` to `storage provider node`
4. Add beetswap protocol support to Server and client, prove that they can connect and download a single file (hardcoded data always returned)
5. Add ability to storage provider server to return a CAR archive identified by CID (I guess use local indexer).
6. Call polka-fetch with those multiple CIDs and verify those archives are downloaded properly.

(I just thought on the fly about it, nothing binding)
So very similar to the list of tasks that you wrote in #634.

PR #656, opened 2024-12-23
feat(pallet-randomness): add benchmarks

jmg-duarte:
### Description

As stated in the title. 

### Checklist


pallets/randomness/src/benchmarking.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[frame_benchmarking::v2::benchmarks(
+    where T: crate::Config
+)]
+mod benchmarks {
+    use super::*;
+
+    /// The worst case scenario is when the history is full.
+    /// Instead of just adding the randomness value, it requires removing the oldest value too.
+    #[benchmark]
```
th7nder, line 18: Could you explain how does this work? I mean, I can't see how we are filling up history here.

runtime/src/configs/mod.rs
```diff
@@ -424,4 +424,5 @@ impl pallet_faucet::Config for Runtime {
 /// Config for our randomness pallet
 impl pallet_randomness::Config for Runtime {
     type AuthorVrfGetter = BabeDataGetter<Runtime>;
+    type WeightInfo = ();
```
th7nder: What does it do? Why are we setting weight info to nothing?

th7nder (changes requested):

runtime/src/configs/mod.rs
jmg-duarte: We're not setting it to "nothing", WeightInfo is implemented for unit

pallets/randomness/src/benchmarking.rs
jmg-duarte: I've fixed it, should be clear now

pallets/proofs/src/benchmarking.rs
```diff
@@ -0,0 +1,63 @@
+#![cfg(feature = "runtime-benchmarks")]
+
+use frame_benchmarking::v2::*;
+use frame_system::RawOrigin;
+
+use super::*;
+#[allow(unused)]
+use crate::Pallet as ProofsPallet;
+
+// NOTE: these benchmarks are cannot be accurate until the extrinsic takes in a bounded key size
```
th7nder: Let's add a todo/link to an existing issue so we don't forget to do this.

pallets/weights.rs
```diff
@@ -0,0 +1,57 @@
+
+//! Autogenerated weights for `pallet_randomness`
```
th7nder: Why this file exists and is in `pallets/weights.rs`, but it is for pallet randomness?

th7nder (changes requested):

pallets/weights.rs
jmg-duarte: fixed

pallets/proofs/src/benchmarking.rs
jmg-duarte: this needs a bound, otherwise "prod" compilation fails

th7nder (dismissed):

cernicc (approved):

cernicc:
Good job figuring this out 💪

th7nder (approved):

PR #657, opened 2024-12-23
feat(pallet-proofs): add benchmarks

jmg-duarte:
### Description

Adds benchmarks for the proofs pallet


runtime/src/configs/mod.rs
```diff
@@ -411,6 +411,13 @@ impl pallet_market::Config for Runtime {
 
 impl pallet_proofs::Config for Runtime {
     type RuntimeEvent = RuntimeEvent;
+    type WeightInfo = pallet_proofs::weights::Weights<Runtime>;
+}
+
+/// Config for our randomness pallet
+impl pallet_randomness::Config for Runtime {
+    type AuthorVrfGetter = BabeDataGetter<Runtime>;
+    type WeightInfo = pallet_randomness::weights::Weights<Runtime>;
```
th7nder, line 420: That's funny, weren't the weights set before for pallet randomness?
And the PR title says pallet randomness, I think it's for pallet proofs.

pallets/weights.rs
th7nder, line 1: looks like an unnecessary file

th7nder (changes requested):

runtime/src/configs/mod.rs
jmg-duarte: extra file too

th7nder (approved):

cernicc (approved):

PR #658, opened 2024-12-24
feat: verify multiple partitions in pallets

### LINKED ISSUE #595: Add support for multiple partitions when verifying PoSt
1. Add support in pallet_storage_provider::submit_windowed_post() to accept multiple proofs, i.e. 1 proof per partition
2. https://github.com/filecoin-project/builtin-actors/blob/b4ed8cf96bcba90b1d6dc2272e23d1021802d949/actors/miner/src/sectors.rs#L86 -> any faulty sector is just replaced with first non faulty in the partition
3. Add support in pallet_proofs::verify_post() to verify multiple partitions
    - https://github.com/filecoin-project/ref-fvm/blob/551e24a9b7b4c8b1e42731b497d12202732beed1/fvm/src/kernel/filecoin.rs#L346C19-L346C37
    - https://github.com/filecoin-project/rust-filecoin-proofs-api/blob/bdc9a595eba17f7fe89142b7b1fa78961bd94c50/src/post.rs#L519
    - https://github.com/filecoin-project/rust-fil-proofs/blob/90dd8b08d71c1140177fd7f81f8c7ec6d7f1fdc5/filecoin-proofs/src/api/window_post.rs#L181
    - https://github.com/filecoin-project/rust-fil-proofs/blob/90dd8b08d71c1140177fd7f81f8c7ec6d7f1fdc5/storage-proofs-post/src/fallback/compound.rs#L54
4. Add support in pipeline generate proofs to generate multiple partitions.
     - Run full PoSt cycle: https://github.com/filecoin-project/lotus/blob/e7db584232113ce70dbb011dc3d598014e02fff0/storage/wdpost/wdpost_run.go#L270 
     - https://github.com/filecoin-project/lotus/blob/e7db584232113ce70dbb011dc3d598014e02fff0/storage/wdpost/wdpost_run.go#L596
     - they need to be ordered by sector number, because the partitions will end up being different!

Key thing is there, that if a sector is faulty, we include a non-faulty one as a replacement.

~~1. Issues with cancellation token~~
~~2. Cannot publish multiple deals at once?~~

th7nder:
### Description

This PR is part of many.
It introduces the batch verification of partitions in a single extrinsic call in `pallet-proofs`/`pallet-storage-provider`.
Multiple proofs generation will come in another one and so on.
Changed the hashing algorithm on pallet randomness to `Blake128_Concat` as #656 broke something and we were getting panic god knows why.

Part of #595.

### Checklist

- [X] If there are follow-ups, have you created issues for them?
  - [X] If yes, which ones? / List them here


pallets/storage-provider/src/lib.rs
```diff
@@ -799,11 +802,18 @@ pub mod pallet {
                 &entropy,
             )?;
 
+            let mut proofs = BoundedVec::new();
+            for proof in windowed_post.proofs {
```
cernicc, line 806: Would it be possible to unify the bounds of these Vecs? That way I think we could pass the `windowed_post.proofs` directly to the `verify_post` function. Without constructing a new BoundedVec.

th7nder: It requires changes across many files, as
- extrinsic param: `proofs: BoundedVec<PoStProof, ConstU32<10>>`
- trait param: `proofs: BoundedVec<BoundedVec<u8, ConstU32<196>>, ConstU32<10>>`

examples/rpc_publish.sh
```diff
@@ -36,9 +36,11 @@ target/release/storagext-cli --sr25519-key "$PROVIDER" market add-balance 250000
 # register one of them as the storage provider
 wait
 
-target/release/storagext-cli --sr25519-key "//Charlie" storage-provider register "peer_id"
-target/release/storagext-cli --sr25519-key "//Charlie" proofs set-porep-verifying-key @2KiB.porep.vk.scale
-target/release/storagext-cli --sr25519-key "//Charlie" proofs set-post-verifying-key @2KiB.post.vk.scale
+target/release/storagext-cli --sr25519-key "//Charlie" storage-provider register "peer_id" &
```
jmg-duarte, line 42: Document why they are all different

examples/rpc_publish_multiple_sectors.sh
jmg-duarte, line 1: Given we're adding new scripts, maybe it's time to remove the old and broken ones too

examples/rpc_publish_multiple_sectors.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+set -e
```
jmg-duarte, line 2:
```suggestion
#!/usr/bin/env bash -e
```

pallets/proofs/src/lib.rs
```diff
@@ -165,28 +165,39 @@ pub mod pallet {
...
+            let mut parsed_proofs = BoundedVec::new();
+            for (index, proof) in proofs.into_iter().enumerate() {
+                let proof = Proof::<Bls12>::decode(&mut proof.as_slice()).map_err(|e| {
+                    log::error!(target: LOG_TARGET, "failed to parse PoSt proof (idx: {}){:?}", index, e);
+                    Error::<T>::Conversion
+                })?;
+
+                parsed_proofs.try_push(proof).expect(
+                    "ProofScheme to have BoundedVec with the same bound as ProofVerification trait",
+                );
```
jmg-duarte: This expect description is kind of confusing

pallets/proofs/src/post/mod.rs
```diff
@@ -40,9 +42,13 @@ impl ProofScheme {
     pub fn verify(
         &self,
         randomness: Ticket,
-        replicas: BoundedBTreeMap<SectorNumber, PublicReplicaInfo, ConstU32<MAX_SECTORS_PER_PROOF>>,
+        replicas: BoundedBTreeMap<
+            SectorNumber,
+            PublicReplicaInfo,
+            ConstU32<{ MAX_SECTORS_PER_PROOF * MAX_PROOFS_PER_BLOCK }>,
```
jmg-duarte: Maybe a const just for this would be better

primitives/src/lib.rs
```diff
@@ -17,6 +17,14 @@ pub const NODE_SIZE: usize = 32;
 /// ref: <https://github.com/filecoin-project/builtin-actors/blob/82d02e58f9ef456aeaf2a6c737562ac97b22b244/runtime/src/runtime/policy.rs#L283>
 pub const MAX_PARTITIONS_PER_DEADLINE: u32 = 3000;
 
+/// Establishes how many partitions can we verify in a single extrinsic.
+/// It's determined by the timing limitations, storage provider have an upper limit of 3000 partitions per deadline.
+/// With our current verification solution, it'll take around ~30 extrinsic calls to verify all of them.
+/// Verification of a single proof takes around ~100ms, block time is ~6000ms.
+/// This means 10 partitions will be verified in a ~1 sec.
+// TODO(@th7nder,#659,27/12/2024): possibly speed it up
```
jmg-duarte, line 25: I can't source the reference now, but we actually have at most ~2 seconds since there's more to block building than just processing extrinsics

pallets/storage-provider/src/tests/mod.rs
```diff
@@ -107,11 +108,14 @@ impl ProofVerification for DummyProofsVerification {
...
+        >,
+        proofs: BoundedVec<
+            BoundedVec<u8, ConstU32<MAX_POST_PROOF_BYTES>>,
+            ConstU32<MAX_PROOFS_PER_BLOCK>,
         >,
-        proof: BoundedVec<u8, ConstU32<MAX_POST_PROOF_BYTES>>,
     ) -> sp_runtime::DispatchResult {
-        if *proof == INVALID_PROOF {
+        if *proofs[0] == INVALID_PROOF {
             return Err(sp_runtime::DispatchError::Other("invalid proof"));
```
jmg-duarte, line 121: Does this still match the docs for `DummyProofsVerification`?

jmg-duarte (commented):
Nothing major to report

examples/rpc_publish_multiple_sectors.sh
th7nder: Shellcheck complains. 
https://www.shellcheck.net/wiki/SC2096

primitives/src/lib.rs
th7nder: I think allocating 1sec for 1 extrinsic call in a block is limit good enough then?
There is 1 sec more for rest of the extrinsics.

pallets/storage-provider/src/tests/mod.rs
th7nder: good catch, I'll update the docs.

primitives/src/lib.rs
jmg-duarte: That's for the benchmarking results to say, that's literally what they're for

jmg-duarte (approved):

primitives/src/lib.rs
th7nder: Proper bound will be verified in the benchmarks, as talked offline.

cernicc (approved):

PR #660, opened 2024-12-30
feat: implement file blockstore

### LINKED ISSUE #653: Implement mater file blockstore
In mater. Implement blockstore that saves and reads blocks directly from the car file

cernicc:
### Description

This is an implementation of FileBlockstorage used to persist car blocks directly to a file on disk.

### Checklist


mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,296 @@
...
+}
+
+impl FileBlockstore {
+    /// Create a new blockstore. If file at the path already exists it is truncated.
+    pub async fn new<P>(path: P, roots: Vec<Cid>) -> Result<Self, Error>
+    where
+        P: AsRef<Path>,
+    {
+        let mut file = File::options()
+            .truncate(true)
```
th7nder: Why would we override file existing at a given path? Shouldn't we avoid writing to the same path?

I mean, is the CARv2 archive reopenable, or not? If we can't reopen it and append data to it, we should throw an error here and say there is a file.

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,296 @@
...
+/// blocks are written immediately, while the index is stored in memory and
+/// updated incrementally.
+///
+/// The blockstore should be closed once the putting blocks is finished. Upon
+/// closing the blockstore, the index is written out to underlying file.
+pub struct FileBlockstore {
+    // Inner store
+    inner: Mutex<FileBlockstoreInner>,
+    // Index of blocks that will be appended to the file at the finalization
+    index: RwLock<IndexMap<Cid, u64>>,
```
th7nder: I think it's worth documenting that stored number is a byte offset in the CarV2 file relative to the end of the CarV2 Header

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,296 @@
...
+        // Lock writer
+        let mut inner = self.inner.lock().await;
+
+        // This is a current position of the writer. We save this to the indexer
+        // so that we know where we wrote the current block.
+        let current_position = inner.store.stream_position().await?;
+        let index_location = current_position - CarV2Header::SIZE;
+
+        // Write block
+        // let buffered_writer = BufWriter::new(inner.store);
```
th7nder: Why  `buffered_writer` isn't used here? Should it? If not, I'd remove the comment.

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,296 @@
...
+        // Get the index if exists
+        let Some(index) = self.index.read().await.get(&cid).copied() else {
+            return Ok(None);
+        };
+
+        // The lock is hold through out the method execution. That way we are
+        // certain that the file is not used and we are moving the cursor back
+        // to the correct place after the read.
+        let mut inner = self.inner.lock().await;
+        let current_cursor_location = inner.store.stream_position().await?;
```
th7nder: I think `data_size` == `the place we can start writing new block`, so we can use `data_size` as cursor start location for writing instead of saving `stream_position` and restoring it later.

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,296 @@
...
+        let mut inner = self.inner.lock().await;
+
+        // The blockstore was already finalized
+        if inner.is_finalized {
+            return Ok(());
+        }
+
+        // Correct CARv2 header
+        let header = CarV2Header {
+            characteristics: Characteristics::EMPTY,
```
th7nder, line 223: I don't understand what are characteristics, but I can see a value: `FULLY_INDEXED`, why not this?

mater/lib/src/v2/mod.rs
```diff
@@ -82,7 +87,7 @@ impl Header {
     /// The [`Header`] size in bytes (includes the pragma).
     ///
     /// As defined in the [specification](https://ipld.io/specs/transport/car/carv2/#header).
-    pub const SIZE: usize = PRAGMA.len() + 40;
+    pub const SIZE: u64 = PRAGMA_SIZE + 40;
```
th7nder, line 90: Why conversion to u64? I don't think size can be < 0.

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,296 @@
...
+/// closing the blockstore, the index is written out to underlying file.
+pub struct FileBlockstore {
+    // Inner store
+    inner: Mutex<FileBlockstoreInner>,
+    // Index of blocks that will be appended to the file at the finalization
+    index: RwLock<IndexMap<Cid, u64>>,
+}
+
+/// Inner file store. Encapsulating state that is locked and used together.
+struct FileBlockstoreInner {
```
th7nder, line 37: Could you remind me how it's gonna be used in the Storage Provider Server?

th7nder (changes requested):

mater/lib/src/stores/file.rs
cernicc: FileBlockstore will be used by the client. Basically the client requests blocks from the providers and then stores the received blocks to the FileBlockstore.

th7nder: Why does the client uses FileBlockstore to store the files?
I mean, when you download the file, what's the point of storing it in the blockstore?

mater/lib/src/stores/file.rs
cernicc: I agree. I was also not entirely sure about truncation.

mater/lib/src/v2/mod.rs
cernicc: u64 is unsigned. I've added casting here so that it's not needed at the call site.

mater/lib/src/stores/file.rs
cernicc: Fully indexed would mean that we are also indexing the identity cids. We are not doing that.

mater/lib/src/stores/file.rs
cernicc: Good one. I've updated the code

mater/lib/src/stores/file.rs
cernicc: I've committed the buffered writer.

mater/lib/src/stores/file.rs
cernicc: The content is transferred in cid-addressed blocks. [This](https://docs.ipfs.tech/assets/img/diagram-of-the-want-have-want-block-process.6ef862a2.png) image shows nicely how the protocol works. The client requests and receives blocks from the car archive. Then when it gets those blocks back it stores them in the blockstore.

You could imagine it something like, the blockstore is an intermediary state that is used to store the car file received as a stream of blocks. The blockstore is finalized (the index and correct header is written to disk) when the client sees that there are no blocks needed to be requested anymore. When the blockstore is finalized, client is left with the correct car file on disk.

th7nder: 1. Ok, so when client wants to retrieve a file, it needs to ask for each of the blocks in the CARv2 archive and then reconstruct it on their own. How does the client know which blocks to ask for?
2. If we're putting the blocks one by one in to the block store, without any ordering, how can we know how to reconstruct the source data from the CARv2 archive? Are we requesting them sequentially, 1 by 1, so the order is always the same? 
3. (rephrasing 2) So, if a clients requests all of the blocks (assuming they know their CIDs), how do they reconstruct the final file they've been asking for?

I guess what I'm asking is, what's the end to end flow, of putting for example 1GB video (of me kitesurfing) on-chain and retrieving it using this component, so it's playable again?

cernicc: 1. The client gets the payload (root) cid of the car file from the user. That means the client knows which is the first block that needs to be fetched. If the received block represents a node in the DAG (is not a leaf). Then it decodes the content. By doing that it gets the next cids it has to request. The client is doing this until there are some blocks left to request.

2. The easiest and fastest client to implement is to request the blocks in sequence. That way the client knows that the blocks received are in sequence and they can be appended to the file. The improvement would be to implement some kindof linking system that would hold the state of the links.

3. The reconstruction is done on the fly by the blockstore in this PR. So to quickly sum up. The client requests the blocks sequentially. The received blocks are added to the FileBlockstore and as a side effect the car is being build on the disk. After all blocks are requested and received, the blockstore is finalized by the client.

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+    store: File,
+    // The byte length of the CARv1 payload. This is used by the indexing, so we
+    // know the locations of each blocks in the file.
+    data_size: u64,
+    // Is true if the blockstore was finalized.
+    is_finalized: bool,
+}
+
+impl FileBlockstore {
+    /// Create a new blockstore. If file at the path already exists it is truncated.
```
jmg-duarte: This isn't the case, `create_new` will fail if the file already exists.

> This option indicates whether a new file will be created. No file is allowed to exist at the target location, also no (dangling) symlink.

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+    pub async fn new<P>(path: P, roots: Vec<Cid>) -> Result<Self, Error>
+    where
+        P: AsRef<Path>,
+    {
```
jmg-duarte, line 56: Ideally, you should check that the roots are not empty
```suggestion
    pub async fn new<P>(path: P, roots: Vec<Cid>) -> Result<Self, Error>
    where
        P: AsRef<Path>,
    {
        if roots.is_empty() {
            return Err(crate::Error::EmptyRoots);
        }
```

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+        let mut file = File::options()
+            .create_new(true)
+            .write(true)
+            .read(true)
+            .open(path)
+            .await?;
+
+        // Write headers
+        v2::write_header(&mut file, &CarV2Header::default()).await?;
+        let written = v1::write_header(&mut file, &CarV1Header::new(roots.clone())).await?;
```
jmg-duarte, line 70: Unnecessary clone, I'm kinda surprised clippy didn't catch this one
```suggestion
        let written = v1::write_header(&mut file, &CarV1Header::new(roots)).await?;
```

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+/// The blockstore should be closed once the putting blocks is finished. Upon
+/// closing the blockstore, the index is written out to underlying file.
+pub struct FileBlockstore {
+    // Inner store
+    inner: Mutex<FileBlockstoreInner>,
+    // Index of blocks that will be appended to the file at the finalization.
+    // Stored number is an offset that locates the first byte of the block
+    // within the CARv1 payload. The offset is relative to the start of the
+    // CARv1 payload.
+    index: RwLock<IndexMap<Cid, u64>>,
```
jmg-duarte: Why isn't the index part of the `inner`? Having a lock for each sounds like a great way to shoot ourselves in the foot

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+    /// Finalize this blockstore by writing the CARv2 header, along with
+    /// flattened index for more efficient subsequent read.
+    async fn finalize(self) -> Result<(), Error> {
+        // Locked underlying file handler
+        let mut inner = self.inner.lock().await;
+
+        // The blockstore was already finalized
+        if inner.is_finalized {
+            return Ok(());
+        }
```
jmg-duarte: This doesn't make a lot of sense. 

The `finalize` method consumes `self`, as such, there's no way the user can get the object back, consequently, that means there's no way for the user to call `finalize` twice. The whole flag seems pointless

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+            data_offset: CarV2Header::SIZE,
+            data_size: inner.data_size,
+            index_offset: CarV2Header::SIZE + inner.data_size,
+        };
+
+        // Write correct CARv2 header
+        inner.store.rewind().await?;
+        v2::write_header(&mut inner.store, &header).await?;
+
+        // Flatten and write the index
```
jmg-duarte: In this context, what does flatten mean?

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+    ) -> Result<(), blockstore::Error> {
+        let cid = Cid::try_from(cid.to_bytes()).map_err(|_err| blockstore::Error::CidTooLarge)?;
+
+        self.put(&cid, data)
+            .await
+            .map_err(|err| blockstore::Error::FatalDatabaseError(err.to_string()))
+    }
+
+    async fn remove<const S: usize>(&self, _cid: &CidGeneric<S>) -> Result<(), blockstore::Error> {
+        unimplemented!("Operation not supported")
```
jmg-duarte: This panics, you should return an error instead

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    #[tokio::test]
+    async fn test_blockstore() {
+        // Car file
+        let mut file =
+            File::open(PathBuf::from_str("tests/fixtures/car_v2/spaceglenda.car").unwrap())
+                .await
+                .unwrap();
+        let mut original_archive = Vec::new();
+        file.read_to_end(&mut original_archive).await.unwrap();
```
jmg-duarte: You can replace all this with `tokio::fs::read`

mater/lib/src/stores/file.rs
th7nder: 1. Ok, this decoding and requesting next blocks makes sense. A quick search told me that the block size is 256kB. So for now, if we have 2KB sector size then it's always 1 block, cause file will always be contained in a 1 block.
2. I guess when we have the tree structure in the 1st block and in the following non-leaf nodes block, then we know the order.
3. Nice, I understand now.

So from the user perspective:
```bash
# Upload && Preparation
$ mater-cli convert -q --overwrite "konrad-kiting.mp4" "konrad-kiting.mp4.car" 
FILE_CID
$ polka-storage-provider-client propose-deal @deal.json
DEAL_CID
$ curl -X PUT -F "upload=@konrad-kiting.mp4" "http://localhost:8001/upload/$DEAL_CID"
$ polka-storage-provider-client publish-deal @signed-deal,json
# ... Downloading
$ polka-fetch $FILE_CID -o konrad-kiting-downloaded.mp4.car
$ mater-cli extract konrad-kiting-downloaded.mp4.car
```

The only thing user needs to know is the `FILE_CID` which is generated using mater, rest of it will work.
That reassured me.

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    /// Finalize this blockstore by writing the CARv2 header, along with
+    /// flattened index for more efficient subsequent read.
+    async fn finalize(self) -> Result<(), Error> {
+        // Locked underlying file handler
+        let mut inner = self.inner.lock().await;
+
+        // The blockstore was already finalized
+        if inner.is_finalized {
+            return Ok(());
```
th7nder: This should be an error, I don't see why anyone in normal usage would try to finalize twice.

th7nder (dismissed):

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+                    let position = reader.get_inner_mut().stream_position().await.unwrap();
+                    let data_end = header.data_offset + header.data_size;
+                    if position >= data_end {
+                        break;
+                    }
+                }
+                else_ => {
+                    // With the length check above this branch should actually be unreachable
+                    assert!(matches!(else_, Err(Error::IoError(_))));
+                    break;
```
jmg-duarte: It either is or isn't. Test it out with `unreachable`

mater/lib/src/stores/file.rs
```diff
@@ -0,0 +1,306 @@
...
+        let v1_header = reader.read_v1_header().await.unwrap();
+
+        let tmp_dir = TempDir::new().unwrap();
+        let blockstore_file_path = tmp_dir.path().join("blockstore.car");
+        let blockstore = FileBlockstore::new(&blockstore_file_path, v1_header.roots)
+            .await
+            .unwrap();
+
+        loop {
+            // NOTE(@jmg-duarte,22/05/2024): review this
```
jmg-duarte: This looks ok, what were you looking from me?

jmg-duarte (commented):
Still some stuff to be cleared

mater/lib/src/stores/file.rs
cernicc: Good catch. The comment is a leftover from before times

mater/lib/src/stores/file.rs
cernicc: True. Good catch

mater/lib/src/stores/file.rs
cernicc: True. Good catch

mater/lib/src/stores/file.rs
cernicc: I've removed the flatten from the comment. But in this context, it meant like "prepare the index in the correct shape for writing".

mater/lib/src/stores/file.rs
cernicc: Thanks 😅

mater/lib/src/stores/file.rs
cernicc: I've copied the loop from one of your tests. This was a runaway comment.

mater/lib/src/stores/file.rs
cernicc: I've removed the flag. This was solved by taking self by ownership. The borrow checker prevents calling finalize more than once.

mater/lib/src/stores/file.rs
cernicc: I was thinking of adding it to the inner. But `has` method is only checking the index and we don't need to lock the inner when doing that.

jmg-duarte: You can use a RwLock over the inner containing everything instead then. This level of fine grained locking should be avoided until we're sure we need it. Otherwise it brings extra complexity that may or may not be needed

mater/lib/src/stores/file.rs
```diff
@@ -129,17 +131,12 @@ impl FileBlockstore {
         Ok(())
     }
 
-    /// Finalize this blockstore by writing the CARv2 header, along with
-    /// flattened index for more efficient subsequent read.
+    /// Finalize this blockstore by writing the CARv2 header, along with index
```
jmg-duarte: Does the index cover all blocks?

mater/lib/src/stores/file.rs
```diff
@@ -284,10 +278,8 @@ mod tests {
                         break;
                     }
                 }
-                else_ => {
-                    // With the length check above this branch should actually be unreachable
-                    assert!(matches!(else_, Err(Error::IoError(_))));
-                    break;
+                _ => {
+                    unreachable!();
```
jmg-duarte:
```suggestion
                    unreachable!("the length check should avoid this from being reached");
```

jmg-duarte (commented):
One more question: the CAR file should be able to support multiple files inside it, did you test that?

mater/lib/src/stores/file.rs
cernicc: Good point

mater/lib/src/stores/file.rs
cernicc: It indexes all blocks that are added to the store.

th7nder (dismissed):

cernicc:
> One more question: the CAR file should be able to support multiple files inside it, did you test that?

It should work because the store is not doing any additional processing. It stores the received blocks to the underlying car file.

mater/lib/src/stores/file.rs
jmg-duarte: Then you need to review if the full indexed characteristic should be set or not
https://ipld.io/specs/transport/car/carv2/#characteristics

cernicc: Hmm. Let me rephrase. It indexes all blocks that are stored to the underlying file. But it doesn't store blocks with identity CIDs. The `FULLY_INDEXED` means that the identity cids are also part of the written index. Thanks for the heads up. Based on this question I've added the identity handling to the store.

mater/lib/src/multicodec.rs
```diff
@@ -35,3 +38,8 @@ where
     Multihash::wrap(H::CODE, &hashed_bytes)
         .expect("the digest should be valid (enforced by the type system)")
 }
+
+// Returns Some(data) if the CID is an identity. If not, None is returned.
+pub fn is_identity<const S: usize>(cid: &CidGeneric<S>) -> Option<&[u8]> {
+    (cid.hash().code() == IDENTITY_CODE).then_some(cid.hash().digest())
+}
```
jmg-duarte: `is_` implies a predicate and thus a bool as return value, a better name would be `get_identity`

mater/lib/src/stores/file.rs
```diff
@@ -22,25 +22,30 @@ use crate::{
...
+    /// Original roots.
+    roots: Vec<Cid>,
+    /// Inner store
+    inner: RwLock<FileBlockstoreInner>,
 }
 
 /// Inner file store. Encapsulating state that is locked and used together.
 struct FileBlockstoreInner {
-    // Car file data store
+    /// Underlying data store in a car format
```
jmg-duarte:
```suggestion
    /// Underlying data store in CAR format
```

mater/lib/src/stores/file.rs
```diff
@@ -62,43 +67,111 @@ impl FileBlockstore {
...
     }
 
-    /// Check if the store contains a block with the cid.
-    async fn has(&self, cid: Cid) -> Result<bool, Error> {
-        Ok(self.index.read().await.get(&cid).is_some())
+    /// Initialize the blockstore from the existing archive file. Error is
+    /// thrown if the archive can't be read.
+    ///
+    /// Note: The underlying store is opened in read only mode. That means the
+    /// returned blockstore can only be used to read an existing blocks.
```
jmg-duarte, line 88: As to not delay the PR further, this can stay for now but ideally, it would return a different type as to avoid misunderstandings in the API

mater/lib/src/stores/file.rs
```diff
@@ -109,15 +182,21 @@ impl FileBlockstore {
...
-        let mut inner = self.inner.lock().await;
+    /// Put the new block in the store. The data integrity is not checked. We
+    /// expect that the CID correctly represents the data being passed. In case
+    /// of the identity CID, nothing is written to the store.
+    pub async fn put_keyed(&self, cid: &Cid, data: &[u8]) -> Result<(), Error> {
+        if is_identity(&cid).is_some() {
+            return Ok(());
+        }
+
+        // The lock is hold through out the method execution
```
jmg-duarte, line 193:
```suggestion
        // The lock is held throughout the method execution
```

mater/lib/src/multicodec.rs
jmg-duarte: Another idea:
```suggestion
trait IdentityExt {
    fn is_identity(&self) -> bool;
    fn get_identity(&self) -> Option<&[u8]>;
}

impl<const S: usize> IdentityExt for CidGeneric<S> {
    fn is_identity(&self) -> bool {
        cid.hash().code() == IDENTITY_CODE
    }
    
    fn get_identity(&self) -> Option<&[u8]> {
        self.is_identity().then(|| cid.hash().digest())
    }
}
```

mater/lib/src/multicodec.rs
```diff
@@ -35,3 +38,8 @@ where
     Multihash::wrap(H::CODE, &hashed_bytes)
         .expect("the digest should be valid (enforced by the type system)")
 }
+
+// Returns Some(data) if the CID is an identity. If not, None is returned.
+pub fn is_identity<const S: usize>(cid: &CidGeneric<S>) -> Option<&[u8]> {
+    (cid.hash().code() == IDENTITY_CODE).then_some(cid.hash().digest())
```
jmg-duarte: The contents of `.then_some` are always evaluated even if the bool is false, this alternative only gets the digest iff the bool is true
```suggestion
    (cid.hash().code() == IDENTITY_CODE).then(|| cid.hash().digest())
```

mater/lib/src/stores/file.rs
```diff
@@ -174,95 +265,106 @@ impl FileBlockstore {
...
+                .await
+                .map_err(|err| Error::FatalDatabaseError(err.to_string()))
+        }
 
-        self.put(&cid, data)
-            .await
-            .map_err(|err| blockstore::Error::FatalDatabaseError(err.to_string()))
-    }
+        async fn has<const S: usize>(&self, cid: &CidGeneric<S>) -> blockstore::Result<bool> {
+            let cid = Cid::try_from(cid.to_bytes()).map_err(|_err| Error::CidTooLarge)?;
```
jmg-duarte: I'm confused, this is not the only error that can happen (?)

mater/lib/src/stores/file.rs
```diff
@@ -279,20 +381,138 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_blockstore_from_existing() {
+        // Loaded blockstore
+        let blockstore = FileBlockstore::from_existing("tests/fixtures/car_v2/spaceglenda.car")
+            .await
+            .unwrap();
+
+        // Writing a new block should fail because the Blockstore is in ready
+        // only mode.
```
jmg-duarte:
```suggestion
        // Writing a new block should fail because the Blockstore is in read
        // only mode.
```

jmg-duarte (dismissed):

mater/lib/src/stores/file.rs
cernicc: Yeah. I agree

mater/lib/src/multicodec.rs
cernicc: I was also thinking of implementing the extension trait. but I didn't see a big enough of a benefit

mater/lib/src/stores/file.rs
cernicc: Are you asking because of me returning CidTooLarge for all errors? In this case, I think this is the only reason why the conversion would fail. Because we already receive the instantiated CidGeneric so we already know that the format is correct. The problem was that I needed to get a CidGeneric<64> from CidGeneric<S: usize>. This was the best solution I managed to find.

mater/lib/src/stores/file.rs
cernicc: https://github.com/eigerco/polka-storage/pull/660#discussion_r1900245114 The second point in a comment is wrong. Because we are using CARv2 format. The ordering of blocks doesn't matter. It doesn't matter because of the index and stored root cid in a header. We can store blocks in any order. The index holds an offset to each of the blocks.

mater/lib/src/stores/file.rs
jmg-duarte: Ahhh ok, but then why not just map the error from the Cid instead? + the size is not too large, it could also be too small, the library has a SizeMismatch error or similar

jmg-duarte (dismissed):

th7nder (dismissed):

mater/lib/src/stores/file.rs
cernicc: We have to return a [blockstore::Error](https://github.com/eigerco/blockstore/blob/7a0248abc923728238e8a6e9a0dd3f0ae17ba5bc/src/lib.rs#L46). The actual digest size can be smaller from the generic S specified for Multihash struct. The S is the allocated size for the digest. See [read_multihash](https://github.com/multiformats/rust-multihash/blob/90a6c19ec71ced09469eec164a3586aafeddfbbd/src/multihash.rs#L271).

cernicc: We could also probably do something like this 
```rust 
/// Convert generic CID to the CID with the specific multihash size.
    fn to_blockstore_cid(cid: &CidGeneric<S>) -> Result<Cid, Error>{
        let digests_size = cid.hash().size();
        if digest_size > 64 {
            return Err(Error::CidError(CidError::InvalidMultihashLength(digests_size)))
        }

        Ok(Cid::try_from(cid.to_bytes()).expect("cid is correct here"))
    }
```

cernicc: Uuuu. Nice. I just found [resize](https://github.com/multiformats/rust-multihash/blob/90a6c19ec71ced09469eec164a3586aafeddfbbd/src/multihash.rs#L160). I've comitted the change with using resize.

cernicc: I've comitted the change

jmg-duarte (approved):

th7nder (approved):

PR #661, opened 2024-12-30
feat: generate multiple proofs

th7nder:
### Description

Part of #595.
Generates proofs for all of the partitions without considering faults and all.

### Checklist

### DIFF
--- a/storage-provider/server/src/pipeline/mod.rs
+++ b/storage-provider/server/src/pipeline/mod.rs
@@ -1,6 +1,6 @@
 pub mod types;
 
-use std::{path::PathBuf, sync::Arc};
+use std::{collections::BTreeSet, path::PathBuf, sync::Arc};
 
 use polka_storage_proofs::{
     porep::{
@@ -667,9 +667,6 @@ async fn submit_windowed_post(
         return Err(PipelineError::DeadlineStateNotFound);
     };
 
-    if deadline_state.partitions.len() > 1 {
-        todo!("I don't know what to do: polka-storage#595");
-    }
     if deadline_state.partitions.len() == 0 {
         tracing::info!("There are not partitions in this deadline yet. Nothing to prove here.");
         schedule_post(state, deadline_index)?;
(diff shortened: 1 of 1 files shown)


storage-provider/server/src/pipeline/mod.rs
```diff
@@ -667,36 +667,33 @@ async fn submit_windowed_post(
...
 
     let partitions = deadline_state.partitions.keys().cloned().collect();
-    let (_partition_number, PartitionState { sectors }) = deadline_state
-        .partitions
-        .first_key_value()
-        .expect("1 partition to be there");
+    let mut all_sectors = BTreeSet::new();
+    for (_, PartitionState { sectors }) in deadline_state.partitions {
+        all_sectors.extend(sectors);
+    }
```
jmg-duarte: How about something like
```suggestion
    let all_sectors = BTreeSet::from_iter(deadline_state.partitions.iter().map(|(_, PartitionState { sectors })| sectors));
```

storage-provider/server/src/pipeline/mod.rs
```diff
@@ -714,16 +711,19 @@ async fn submit_windowed_post(
...
-        .try_into()
-        .expect("converstion between rust-fil-proofs and polka-storage-proofs to work");
-    let proof = codec::Encode::encode(&proof);
+    let proofs = proofs
+        .into_iter()
+        .map(|p| PoStProof {
+            post_proof: state.server_info.post_proof,
+            proof_bytes: codec::Encode::encode(
+                &TryInto::<SubstrateProof>::try_into(p.clone())
+                    .expect("converstion between rust-fil-proofs and polka-storage-proofs to work"),
```
jmg-duarte, line 724: Why don't we have a `From` or `Into` implementation then?

storage-provider/server/src/pipeline/mod.rs
th7nder: Close enough! `flat_map` and it works.

storage-provider/server/src/pipeline/mod.rs
th7nder: We can have `From` implementation, but it'd `.expect()` anyway.
The reason those conversions work is because `blstrs` library was built on top of `bls12_381` library.
If at some point folks from `rust-fil-proofs` on which we depend, change the implementation, it'd stop working.
This gives us a nice message, instead of a panic in random place.

jmg-duarte: If you have a `From`/`Into` it shouldn't fail at all. If then something changes in the API we have a breakage, no?

th7nder: Exactly, that's why I'd like to leave this expect.

jmg-duarte: solved on slack, try stays

cernicc (approved):

jmg-duarte (approved):

PR #666, opened 2025-01-07
feat(storage-provider-server): add p2p bootstrap and registration

### LINKED ISSUE #652: feat: Implement resolver in sp-server
Implement a resolver registration and bootstrap node in the storage provider server.

### LINKED ISSUE #667: feat: Implement public key to Peer ID conversion
In the polka-storage-provider-client add a command to convert a public key (ED25519 PEM format) to a libp2p Peer ID

aidan46:
### Description

This PR adds bootstrap and registration p2p node functionality to the polka-storage-provider-server. The node type is selected by passing the node-type argument into the CLI (bootstrap or registration). Configuration is read from a TOML file, see the example in the p2p-example-config folder. The P2P network supports an ED25519 private key in PEM file format.

The `generate-peer-id` command has been added to the polka-storage-provider-client. This command generates a peer ID from an ED25519 public key PEM file and writes it to a file or stdout, depending on the `--file` argument.

Note:
I have not added any P2P discovery, this will be implemented in the fetching application. I can add a discovery example if needed.


Cargo.toml
```diff
@@ -66,6 +66,7 @@ color-print = "0.3.4"
 criterion = "0.5.1"
 digest = "0.10.7"
 docify = { version = "0.2.8" }
+ed25519-dalek = { version = "2.1.1", default-features = false }
```
th7nder, line 70: Any reason why we're using ed25519 specifcally?

storage-provider/client/src/commands/mod.rs
```diff
@@ -98,6 +112,18 @@ pub(crate) enum Cli {
         #[command(flatten)]
         signer_key: MultiPairArgs,
     },
+
+    /// Generate a Peer ID from a ED25519 public key pem file
+    GeneratePeerID {
```
th7nder, line 113: I don't see a reason why we're using a new public key instead of reusing the `signer_key`/`MultiPairArgs` which is used for other operations as well.

storage-provider/client/src/commands/mod.rs
```diff
@@ -182,4 +209,17 @@ impl Cli {
...
     }
+
+    fn generate_peer_id(path: PathBuf, file: Option<PathBuf>) -> Result<(), CliError> {
+        let pubkey_bytes = PublicKeyBytes::read_public_key_pem_file(path)?;
+        let pubkey = EdPubKey::try_from_bytes(&pubkey_bytes.to_bytes())?;
+        let key = libp2p::identity::PublicKey::from(pubkey);
+        let peer_id = PeerId::from_public_key(&key).to_string();
+
+        match file {
+            None => io::stdout().lock().write_all(&peer_id.as_bytes())?,
```
th7nder: I don't think writing it as a raw bytes is useful, it should be something like `println!("{}", hex::encode(peer_id.as_bytes())`.

storage-provider/server/p2p-example-config/private.pem
th7nder, line 1: If we use the same idea as in `signer_key`, then those files should not be needed.

storage-provider/server/p2p-example-config/bootstrap.toml
th7nder, line 1: I'd suggest putting those example configs in `<repo-dir>/examples`, so we have a centralized place for this.

storage-provider/server/src/main.rs
```diff
@@ -244,6 +249,14 @@ pub struct ServerArguments {
     /// E.g. With 2KiB sector sizes and 16GiB of RAM, it goes OOM at 4 parallel.
     #[arg(long, default_value = "2")]
     parallel_prove_commits: NonZero<usize>,
+
+    /// P2P Node type, can be either a bootstrap node or a registration node.
+    #[arg(long)]
+    node_type: NodeType,
```
th7nder: How about setting it as default to `bootstrap` node and using some nice defaults as well, so we don't have to specify everything when we're running the node?

storage-provider/server/src/main.rs
```diff
@@ -244,6 +249,14 @@ pub struct ServerArguments {
...
+
+    /// Path to P2P config file
+    #[arg(long)]
+    p2p_config: PathBuf,
```
th7nder: I think we should have 1 global config for everything related to storage provider and then subsections for stuff like p2p.

So like `ServerArguments` being `serde::Serialize/deserialize` and then loading from a file and overriding by CLI options.

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,220 @@
...
+    tracker.wait().await;
+
+    Ok(())
+}
+
+async fn run_p2p_node(node_type: NodeType, config: PathBuf) -> Result<(), P2PError> {
+    match node_type {
+        NodeType::Bootstrap => {
+            let contents = read_to_string(config)?;
+            let config: BootstrapConfig = toml::from_str(&contents)?;
```
th7nder: Config should be parsed as a first step, before running anything.

th7nder (changes requested):

storage-provider/client/src/commands/mod.rs
aidan46: This is because if the storage provider uses a sr25519 key, we cannot re-use it. libp2p currently only supports [RSA, ED25519, Secp256k1 and ECDSA](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types) of which ED25519 must be supported and the other ones should or may be supported (see the link provided). Therefore, I chose to only support ED25519, at least for now. I did find an [open issue for sr25519 support](https://github.com/libp2p/specs/issues/409) but no linked PRs yet.

storage-provider/server/src/main.rs
aidan46: I agree with this, having TOML/YAML support for all `ServerArguments` would be the way to go. I'll open an issue for this.

storage-provider/client/src/commands/mod.rs
aidan46: It should not be hex-encoded, this gives a wrong peer ID. I will just print the peer ID without converting it.

Cargo.toml
aidan46: See my [comment on the other thread](https://github.com/eigerco/polka-storage/pull/666#discussion_r1905340342).

storage-provider/server/src/main.rs
th7nder: #670

storage-provider/server/src/p2p.rs
```diff
@@ -52,24 +55,29 @@ pub enum P2PError {
     P2PTransportError(#[from] libp2p::TransportError<std::io::Error>),
 }
 
-fn create_keypair<P: AsRef<Path> + std::fmt::Debug>(path: P) -> Result<Keypair, P2PError> {
+pub fn create_keypair<P: AsRef<Path> + std::fmt::Debug>(path: P) -> Result<Keypair, P2PError> {
```
jmg-duarte: Does it need to be a P2P error? Or do they wrap around something more specific we could use?

jmg-duarte: Could we also support the @ dialect or its not practical?

aidan46: In this function the P2P error wraps around 	`ed25519_dalek::pkcs8::Error` and `libp2p::identity::DecodingError`.

aidan46: Do you mean support to @ dialect when using a PEM file and otherwise put the private key straight into the config file?

jmg-duarte: > In this function the P2P error wraps around `ed25519_dalek::pkcs8::Error` and `libp2p::identity::DecodingError`.

We could probably use the lower one for more accurate errors? Not sure, not super important



> Do you mean support to @ dialect when using a PEM file and otherwise put the private key straight into the config file?

Yes

examples/p2p-config/register.toml
```diff
@@ -0,0 +1,3 @@
+keypair = "@storage-provider/server/p2p-example-config/private.pem"
```
th7nder: wrong path, should be to the '@examples`

pallets/storage-provider/src/tests/mod.rs
```diff
@@ -198,7 +198,7 @@ impl pallet_storage_provider::Config for Test {
     type RuntimeEvent = RuntimeEvent;
     type Randomness = DummyRandomnessGenerator<Self>;
     type AuthorVrfHistory = DummyRandomnessGenerator<Self>;
-    type PeerId = BoundedVec<u8, ConstU32<32>>; // Max length of SHA256 hash
+    type PeerId = BoundedVec<u8, ConstU32<42>>; // https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#peer-ids
```
th7nder: I guess it should be a `const` in `primitives`?

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,236 @@
...
+    };
+    Keypair::ed25519_from_bytes(key.to_bytes()).map_err(de::Error::custom)
+}
+
+pub async fn run_bootstrap_node(
+    config: BootstrapConfig,
+    token: CancellationToken,
+) -> Result<(), P2PError> {
+    info!("Starting P2P bootstrap node");
+    let tracker = TaskTracker::new();
```
th7nder: What's the reason for using tracker here? Can't we just spawn the task and await it in a select?

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,236 @@
...
+    let (swarm, rendezvous_point_address, rendezvous_point) = config.create_swarm()?;
+
+    tokio::select! {
+        res = register(
+            swarm,
+            rendezvous_point,
+            rendezvous_point_address,
+            None,
+            Namespace::from_static("polka-storage"),
```
th7nder: const

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    tracker.close();
+    tracker.wait().await;
+
+    Ok(())
+}
+
+/// Register the peer with the rendezvous point.
+/// The ttl is how long the peer will remain registered in seconds.
+async fn register(
```
th7nder: I don't understand why we need to have two different swarms :( I'll need to read up on that.

th7nder (dismissed):

storage-provider/server/src/p2p.rs
aidan46: The tracker is used to ensure a [graceful shutdown](https://tokio.rs/tokio/topics/shutdown#waiting-for-things-to-finish-shutting-down) by waiting for the p2p task to shutdown after the cancellation token has been invoked.

storage-provider/server/src/p2p.rs
aidan46: We need two swarms because each node behaves differently. Since we are not enforcing every storage provider to be a bootstrap node, we cannot require the bootstrap behaviour on a registration node. [More information on swarm behaviour](https://libp2p.github.io/rust-libp2p/libp2p/swarm/trait.NetworkBehaviour.html)

storage-provider/server/src/p2p.rs
aidan46: I added the @ dialect for the key and got rid of the errors by parsing the keypair in the deserialise function instead.

storage-provider/client/src/commands/mod.rs
```diff
@@ -98,6 +108,18 @@ pub(crate) enum Cli {
...
+    /// Generate a Peer ID from a ED25519 public key pem file
+    GeneratePeerID {
+        /// Path to the ED25519 public key pem file
+        #[arg(long)]
+        pubkey: PathBuf,
+
+        /// Path to the output file to write the peer ID to.
+        /// If `None` the peer ID will be printed to stdout.
+        #[arg(long)]
+        file: Option<PathBuf>,
```
jmg-duarte: Lets always print to the stdout and if people want they can `cmd > peer_id`

storage-provider/server/src/main.rs
```diff
@@ -244,6 +255,14 @@ pub struct ServerArguments {
     /// E.g. With 2KiB sector sizes and 16GiB of RAM, it goes OOM at 4 parallel.
     #[arg(long, default_value = "2")]
     parallel_prove_commits: NonZero<usize>,
+
+    /// P2P Node type, can be either a bootstrap node or a registration node.
+    #[arg(long, default_value = "bootstrap")]
```
jmg-duarte:
```suggestion
    #[arg(long, default_value_t = NodeType::Bootstrap)]
```

storage-provider/server/src/main.rs
```diff
@@ -286,6 +305,12 @@ pub struct ServerConfiguration {
 
     /// The number of prove commits to be run in parallel.
     parallel_prove_commits: usize,
+
+    /// P2P Network node type, can either be a bootstrap or registration node
+    node_type: NodeType,
```
jmg-duarte, line 263: Food for thought: could this be part of the P2P config?

storage-provider/server/src/main.rs
```diff
@@ -357,6 +386,20 @@ impl ServerConfiguration {
...
+                    toml::from_str(&contents).map_err(|e| P2PError::TOMLError(e))?;
+                tokio::spawn(run_bootstrap_node(config, cancellation_token.child_token()))
+            }
+            NodeType::Register => {
+                let contents = read_to_string(p2p_config)?;
+                let config: RegisterConfig =
+                    toml::from_str(&contents).map_err(|e| P2PError::TOMLError(e))?;
+                tokio::spawn(run_register_node(config, cancellation_token.child_token()))
+            }
+        };
```
jmg-duarte: Nit: move this into a separate function

storage-provider/server/src/main.rs
```diff
@@ -402,10 +445,17 @@ impl ServerConfiguration {
                 let _ = ok.as_ref().inspect_err(|err| tracing::error!(%err));
             });
 
+        let p2p_task = p2p_task
+            .inspect_err(|err| tracing::error!(%err))
+            .inspect(|ok| {
+                let _ = ok.as_ref().inspect_err(|err| tracing::error!(%err));
+            });
```
jmg-duarte: Maybe we could use something like:
```
macro_rules! inspect_nested_errors {
    ($($task:expr),+ $(,)?) => {
        (
            $(
                $task
                    .inspect_err(|err| tracing::error!(%err))
                    .inspect(|ok| {
                        let _ = ok.as_ref().inspect_err(|err| tracing::error!(%err));
                    })
            ),+
        )
    };
}
```

Usage:
```
let (upload_result, rpc_task, pipeline_task, p2p_task) = inspect_nested_errors!(storage_task, rpc_task, pipeline_task, p2p_task);
```

Playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c36d6887947a8190fd75108a326d1737
Note that in the playground I used `println!` so it actually prints something

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,238 @@
...
+#[derive(Debug, Clone, Copy, ValueEnum)]
+pub enum NodeType {
+    Bootstrap,
+    Register,
+}
+
+impl Display for NodeType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
```
jmg-duarte: I feel that this makes the code more explicit
```suggestion
        Debug::fmt(&self, f)
```

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum P2PError {
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+    #[error(transparent)]
+    DialError(#[from] libp2p::swarm::DialError),
```
jmg-duarte: It's already an error — `P2PError::Dial`
```suggestion
    Dial(#[from] libp2p::swarm::DialError),
```
Don't make the mistake I did in `mater`

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[error("Invalid TCP config for swarm")]
+    InvalidTcpConfig,
+    #[error("Invalid behaviour config for swarm")]
+    InvalidBehaviourConfig,
+    #[error(transparent)]
+    TOMLError(#[from] toml::de::Error),
+    #[error("Failed to register at rendezvous point {0}")]
+    RegistrationFailed(PeerId),
+    #[error(transparent)]
+    P2PTransportError(#[from] libp2p::TransportError<std::io::Error>),
```
jmg-duarte: Same as above

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+fn deser_keypair<'de, D: de::Deserializer<'de>>(d: D) -> Result<Keypair, D::Error> {
```
jmg-duarte: Docs pls, I don't want to read the function to know what it does

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,238 @@
...
+    tracker.wait().await;
+
+    Ok(())
+}
+
+pub async fn run_register_node(
+    config: RegisterConfig,
+    token: CancellationToken,
+) -> Result<(), P2PError> {
+    info!("Starting P2P register node");
```
jmg-duarte: Either use `tracing::<level>` or just `<level>` but make it consistent (personally, I prefer the qualified one)

storage-provider/server/src/p2p.rs
jmg-duarte: So, if I understand correctly, the two swarms are required because the behaviours are not supposed to be composable?

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,238 @@
...
+                        enquirer,
+                        registrations.len()
+                    );
+                }
+            }
+            _other => {}
+        }
+    }
+    Ok(())
+}
```
jmg-duarte: Shouldn't this be in the `boostrap.rs` file?

storage-provider/server/src/p2p/register.rs
```diff
@@ -0,0 +1,54 @@
...
+                noise::Config::new,
+                yamux::Config::default,
+            )
+            .map_err(|_| P2PError::InvalidTcpConfig)?
+            .with_behaviour(|key| RegisterBehaviour {
+                identify: identify::Behaviour::new(identify::Config::new(
+                    "identify/1.0.0".to_string(),
+                    key.public(),
+                )),
+                rendezvous: rendezvous::client::Behaviour::new(key.clone()),
```
jmg-duarte: Could you leave docs on what each behaviour is trying to achieve?

storage-provider/server/src/p2p/bootstrap.rs
```diff
@@ -0,0 +1,49 @@
...
+            )
+            .map_err(|_| P2PError::InvalidTcpConfig)?
+            .with_behaviour(|key| BootstrapBehaviour {
+                rendezvous: rendezvous::server::Behaviour::new(
+                    rendezvous::server::Config::default(),
+                ),
+                identify: identify::Behaviour::new(identify::Config::new(
+                    "identify/1.0.0".to_string(),
+                    key.public(),
+                )),
```
jmg-duarte: Same as in `register.rs`

storage-provider/server/src/p2p.rs
aidan46: Yep, makes more sense. I'll move the register function into its respective module too.

storage-provider/server/src/main.rs
aidan46: Since the config change the P2P config does not exist anymore and moved into the `ConfigurationArgs`.

storage-provider/server/src/main.rs
aidan46: This makes sense to me, makes the code more readable imo.

storage-provider/server/src/main.rs
aidan46: I tried that too but I get the following error:
```shell
error: invalid value 'Bootstrap' for '--node-type <NODE_TYPE>'
  [possible values: bootstrap, register]

  tip: a similar value exists: 'bootstrap'
```
clap is trying to fuck me over 😅

storage-provider/server/src/p2p.rs
aidan46: We might be able to use a single swarm with the [Toggle](https://libp2p.github.io/rust-libp2p/libp2p/swarm/behaviour/toggle/struct.Toggle.html) but I have to look into it a bit more. I suggest going through with this PR and I can check if I can refactor to a single swarm.

jmg-duarte: I don't see `Toggle` as being a solution. Let's move with it as is and we'll review iff needed in the future

docs/src/storage-provider-cli/server.md
```diff
@@ -81,28 +81,50 @@ The path to the PoRep proving parameters. They are shared across all of the node
...
-| `post_parameters`       | NA                             |
+| Name                       | Default                        |
+| -------------------------- | ------------------------------ |
+| `upload-listen-address`    | `127.0.0.1:8000`               |
+| `rpc-listen-address`       | `127.0.0.1:8001`               |
+| `node-url`                 | `ws://127.0.0.1:42069`         |
+| `database-directory`       | `/tmp/<random>/deals_database` |
+| `storage-directory`        | `/tmp/<random>/deals_storage`  |
+| `seal-proof`               | 2KiB                           |
+| `post-proof`               | 2KiB                           |
```
jmg-duarte: I forgot this in my PR 🙈 
```suggestion
| `seal-proof`               | `2KiB`                         |
| `post-proof`               | `2KiB`                         |
```

examples/rpc_publish.sh
```diff
@@ -22,11 +22,24 @@ PROVIDER="//Charlie"
...
 PIECE_CID="$(echo "$INPUT_COMMP" | jq -r ".cid")"
 PIECE_SIZE="$(echo "$INPUT_COMMP" | jq ".size")"
+PEER_ID="$(target/release/polka-storage-provider-client generate-peer-id --pubkey "$P2P_PUBLIC_KEY")"
+CONFIG="/tmp/bootstrap.toml"
+echo "seal_proof = '2KiB'
+post_proof = '2KiB'
+porep_parameters = '2KiB.porep.params'
+post_parameters = '2KiB.post.params'
+rendezvous_point_address = '$P2P_ADDRESS'
+p2p_key = '@$P2P_PRIVATE_KEY'" > "$CONFIG"
```
jmg-duarte: As it is, it's kinda hard to read, could you add more newlines to split groups of related variables? 

And possibly document what they're for, like you did on line 29

examples/p2p-config/bootstrap.toml
jmg-duarte, line 1: Shouldn't this file have been merged with the config?

examples/p2p-config/register.toml
jmg-duarte, line 1: Shouldn't this file have been merged with the config?

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,168 @@
...
+
+pub(crate) use bootstrap::BootstrapConfig;
+pub(crate) use register::RegisterConfig;
+
+const P2P_NAMESPACE: &str = "polka-storage";
+
+#[derive(Default, Debug, Clone, Copy, ValueEnum, Deserialize)]
+pub enum NodeType {
+    #[default]
+    Bootstrap,
```
jmg-duarte: TIL

storage-provider/server/src/main.rs
```diff
@@ -165,6 +173,25 @@ pub enum ServerError {
 
     #[error(transparent)]
     Json(#[from] serde_json::Error),
+
+    #[error(transparent)]
+    P2P(#[from] P2PError),
+}
+
+/// Takes an expression that returns a nested result and
+/// inspected the error while logging it.
```
jmg-duarte:
```suggestion
/// Takes an expression that returns a `Result<Result<T, E2>, E1>`.
/// It tries to inspect and log the first error (`E1`), otherwise,
/// it inspects the result and tries to inspect the nested error (`E2`).
///
/// This macro is *roughly* equivalent to calling:
/// ```text
/// res // : Result<Result<T, E2>, E1>
///     .inspect_err(|e| error!(%e))
///     .inspect(|r| r.inspect_err(|e| error!(%e))
/// ```
```

storage-provider/server/src/main.rs
```diff
@@ -339,31 +386,18 @@ impl Server {
...
-            .inspect_err(|err| tracing::error!(%err))
-            .inspect(|ok| {
-                let _ = ok.as_ref().inspect_err(|err| tracing::error!(%err));
-            });
+        let (upload_result, rpc_task, pipeline_task, p2p_task) =
+            tokio::join!(storage_task, rpc_task, pipeline_task, p2p_task);
+
+        // Inspect and log errors
+        let (upload_result, rpc_task, pipeline_task, p2p_task) =
+            inspect_and_log_nested_errors!(upload_result, rpc_task, pipeline_task, p2p_task);
```
jmg-duarte, line 402: Lovely <3

storage-provider/server/src/main.rs
```diff
@@ -497,3 +539,29 @@ impl Server {
...
+            Some(rendezvous_point) => {
+                let config = RegisterConfig::new(
+                    p2p_state.p2p_key,
+                    p2p_state.rendezvous_point_address,
+                    rendezvous_point,
+                );
+                Ok(tokio::spawn(run_register_node(config, cancellation_token)))
+            }
+            None => return Err(ServerError::P2P(P2PError::InvalidBehaviourConfig)),
+        },
```
jmg-duarte: Nit — I feel that a let guard could be easier to read here because it enables an early return style
```suggestion
        NodeType::Register => {
            let Some(rendezvous_point) = p2p_state.rendezvous_point else {
                return Err(ServerError::P2P(P2PError::InvalidBehaviourConfig))
            };
            let config = RegisterConfig::new(
                p2p_state.p2p_key,
                p2p_state.rendezvous_point_address,
                rendezvous_point,
            );
            Ok(tokio::spawn(run_register_node(config, cancellation_token)))
        },
```

storage-provider/server/src/p2p.rs
jmg-duarte, line 1: Let's not use this pattern. Move this to `p2p/mod.rs`

It become really hard to follow what is what when adding more files

storage-provider/server/src/p2p.rs
```diff
@@ -0,0 +1,168 @@
...
+    #[error(transparent)]
+    P2PTransport(#[from] libp2p::TransportError<std::io::Error>),
+}
+
+/// State struct for P2P node.
+/// Holds all the information needed for spawning a node.
+/// Node can be either a bootstrap or a registration node.
+pub(crate) struct P2PState {
+    /// P2P Node type, bootstrap or registration
+    pub(crate) node_type: NodeType,
```
jmg-duarte: I feel that we could turn this abstraction around and make the `NodeType` contain the necessary state for the respective node type.

This would make weird configs unrepresentable and avoid the `Option<PeerId>`

jmg-duarte (commented):
Looks good to me, I've left some comments that require addressing, others for discussion

examples/p2p-config/bootstrap.toml
aidan46: Oopsie, forgot to remove these configs

aidan46: I'll remove the keys too, since they are documented and generated in the scripts. I have already opened [an issue for p2p key generation in the polka-client](https://github.com/eigerco/polka-storage/issues/676)

storage-provider/server/src/main.rs
```diff
@@ -339,31 +394,18 @@ impl Server {
...
-
-        let pipeline_task = pipeline_task
-            .inspect_err(|err| tracing::error!(%err))
-            .inspect(|ok| {
-                let _ = ok.as_ref().inspect_err(|err| tracing::error!(%err));
-            });
+        let (upload_result, rpc_task, pipeline_task, p2p_task) =
+            tokio::join!(storage_task, rpc_task, pipeline_task, p2p_task);
+
+        // Inspect and log errors
```
cernicc, line 400: Does this mean that every node that is serving upload endpoint and doing a sealing, will also run a p2p bootstrap node?

storagext/cli/src/cmd/storage_provider.rs
```diff
@@ -37,7 +38,7 @@ pub enum StorageProviderCommand {
     #[command(name = "register")]
     RegisterStorageProvider {
         /// PeerId in Storage Provider P2P network, can be any String.
```
cernicc: Update the comment

storage-provider/server/src/p2p.rs
aidan46: Opened [an issue for this](https://github.com/eigerco/polka-storage/issues/677) as it should not block the current PR. I'll work on this refactor tomorrow.

storage-provider/server/src/main.rs
aidan46: They run either a bootstrap node or a registration node. At this time we are not enforcing them to run a bootstrap node although they would be perceived as more trustworthy if choose to they do so.

cernicc: What is the requirement for them to run a registration node? What kind of data will they exchange over the p2p network?

aidan46: A registration node will share its Peer ID and multiaddress over the p2p network. Providing discoverability of this mapping to client on the network by supplying this information to bootstrap nodes.

jmg-duarte (approved):

cernicc (approved):

PR #674, opened 2025-01-13
feat(polka-storage-provider-server): implement config file

jmg-duarte:
### Description

Implements support for a config file (JSON or TOML)

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13474,6 +13474,7 @@ dependencies = [
  "thiserror 2.0.8",
  "tokio",
  "tokio-util",
+ "toml 0.8.19",
  "tower 0.4.13",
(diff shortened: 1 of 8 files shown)


cernicc (dismissed):
👍

examples/sp.config.json
aidan46, line 1: I don't think we need this file unless I am missing something

aidan46 (dismissed):
LGTM :+1:. Should we provide an example configuration for the storage provider?

examples/sp.config.json
jmg-duarte: You're right

cernicc (dismissed):

aidan46 (approved):

cernicc (approved):

Issue #615, opened 2024-11-26
fix(maat): refactor it so it uses pipeline as a library


th7nder:
We cannot hardcode and regenrate porep/post params as we do now, it's combinatorial explosion.
We need to refactor pipeline and reuse it's parts.

PR #678, opened 2025-01-16
feat: wrap contents in unixfs dag

cargopete:
# Implement DAG metadata in CAR file conversion

### Description
This PR modifies how files are converted to CAR format in the mater-cli to properly encode DAG (Directed Acyclic Graph) metadata. Previously, when converting files, the system only stored raw file data as blocks without preserving relationships between blocks. This meant that when a client received a root node, it couldn't determine which blocks to request next.

The implementation changes `convert_file_to_car` to use UnixFS DAG structure by:
1. Exposing the UnixFS module from mater library
2. Using `stream_balanced_tree` to create proper DAG nodes with metadata
3. Storing both file chunks and their relationships in the CAR file

Key code changes:
```rust
// Before: Simple file storage
let cid = create_filestore(source_file, output_file, Config::default()).await?;

// After: DAG-aware storage
let chunk_stream = ReaderStream::new(BufReader::new(source_file));
let mut tree_stream = Box::pin(mater::unixfs::stream_balanced_tree(chunk_stream, 256));
while let Some((cid, data)) = tree_stream.try_next().await? {
    writer.write_block(&cid, &data).await?;
}
```

- UnixFS DAG implementation follows the specification at: https://ipld.io/specs/transport/car/carv1/
- Tests verify DAG structure by confirming multiple blocks are created and properly linked (129 blocks for a 512KB test file)
- Feature flags for serde functionality have been properly maintained

### Checklist
  - [x] The chunk size is set to 256 for balanced tree creation - this affects the DAG structure granularity
  - Added test_dag_structure that verifies proper DAG creation
  - Could have modified create_filestore but chose direct UnixFS implementation for better control over DAG structure
  - UnixFS module and stream_balanced_tree are now public and documented in this description


primitives/src/proofs.rs
jmg-duarte, line 1: This change is unrelated

mater/lib/src/lib.rs
jmg-duarte, line 1: These are internal details. You shouldn't expose this

mater/cli/src/convert.rs
jmg-duarte, line 1: The user should be able to decide if they want unixfs wrapping or not, like go-car does

jmg-duarte (changes requested):

mater/cli/src/convert.rs
```diff
@@ -111,4 +128,53 @@ mod tests {
...
+
+        let mut input_file = File::create(&input_path).await?;
+        input_file.write_all(&test_data).await?;
+
+        let root_cid = convert_file_to_car(&input_path, &output_path, false).await?;
+
+        let car_file = File::open(&output_path).await?;
+        let mut reader = CarV2Reader::new(car_file);
+
+        // Count blocks and verify links
```
cernicc: This test only counts the blocks. We should probably try to deserialize the block data and check if the links are actually encoded.

mater/cli/src/convert.rs
```diff
@@ -111,4 +128,53 @@ mod tests {
...
+        }
+
+        println!("Found {} blocks", block_count);
+        println!("Root block found: {}", found_root);
+
+        // We should have more than one block (indicating DAG structure)
+        assert!(block_count > 1, "Expected multiple blocks in DAG structure");
+        assert!(found_root, "Root block not found in CAR file");
+
+        temp_dir.close()?;
```
cernicc: The temp_dir is auto closed on the drop

mater/cli/src/convert.rs
```diff
@@ -12,14 +13,30 @@ pub(crate) async fn convert_file_to_car(
...
         File::create_new(output_path).await
     }?;
-    let cid = create_filestore(source_file, output_file, Config::default()).await?;
 
-    Ok(cid)
+    let reader = BufReader::new(source_file);
+    let chunk_stream = ReaderStream::new(reader);
+
+    let mut writer = CarV2Writer::new(output_file);
+    let mut tree_stream = Box::pin(mater::unixfs::stream_balanced_tree(chunk_stream, 256));
```
cernicc: If you check the removed `create_filestore` function and then `balanced_import`, it is already using `stream_balanced_tree` internally.

cernicc (changes requested):

maat/tests/real_world.rs
```diff
@@ -251,7 +251,7 @@ where
         assert_eq!(event.sectors.0, sectors_pre_commit_info);
     }
 
-    result.height
+    client.height(true).await.unwrap()
```
th7nder, line 254: It won't work in every case. If we query for it 1ms later and get the next block, instead of the current one, the whole machine will fall down. Lots of stuff depends on this pre-commit height (proofs-related). 
We can let it be, but leave a todo marking that this won't always return the correct height. I will fix it in my PR with making maat work again (MMWA).

maat/tests/real_world.rs
```diff
@@ -408,9 +408,11 @@ async fn real_world_use_case() {
 
     tracing::debug!("base dir: {:?}", network.base_dir());
 
-    let collator = network.get_node(COLLATOR_NAME).unwrap();
-    let client =
-        storagext::Client::from(collator.wait_client::<PolkaStorageConfig>().await.unwrap());
+    let client = storagext::Client::new(
```
th7nder, line 411: Yeah, I'm supposed to fixed that in #615 as well.

mater/lib/src/v2/reader.rs
```diff
@@ -1,12 +1,17 @@
...
 
 use super::index::read_index;
+use crate::multicodec::DAG_PB_CODE;
 use crate::{
     v2::{index::Index, Characteristics, Header, PRAGMA},
     Error,
 };
-
+use ipld_core::cid::Cid;
+use ipld_core::codec::Codec;
```
th7nder, line 10: Looks like the formatting will break on this one (import order): `cargo +nightly fmt` I guess?

mater/lib/src/v2/reader.rs
```diff
@@ -425,4 +461,48 @@ mod tests {
             assert_eq!(fst[0].entries.len(), 4);
         }
     }
+
+    #[tokio::test]
+    async fn test_dag_pb_links() {
```
th7nder, line 466: What are we testing here? Doesn't look like the functionality that was added.

th7nder (changes requested):

cargopete:
will open another PR since i've changed the approach

PR #681, opened 2025-01-16
feat: setup retrieval server and client

### LINKED ISSUE #636: Implement retrieval server/client
Implement server and client used to retrieve a file from the provider. They should exchange blocks using bitswap protocol.

### LINKED ISSUE #637: Implement retrieval cli app
We should create a cli app called `polka-fetch`, that will enable storage clients to retrieve some content specified by the payload CID from the provider.

cernicc:
### Description
The PR implements the retrieval server and client. The server is running a p2p node that listens for block requests. The client runs a temporary p2p node that is used to request the needed blocks by the provider. The new blocks are requested until we receive a whole unixfs DAG tree.

The next step is integrating the server with the storage provider.

Start a simple server as a proof of concept: 
`cargo run --package polka-storage-retrieval --example simple_server --all-features`

Request content from the server:
`cargo run --bin polka-fetch -- --provider /ip4/127.0.0.1/tcp/8989 --output ./download.car --payload-cid bafybeiefli7iugocosgirzpny4t6yxw5zehy6khtao3d252pbf352xzx5q`

Note: For content extraction, you should use `go-car`. The reason is that our mater crate currently doesn't know how to extract content from the encoded unixfs dag tree.

### Checklist


storage-retrieval/cli/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+    #[arg(long)]
+    provider: Multiaddr,
+    /// The CAR file to write to.
+    #[arg(long)]
+    output: PathBuf,
+    /// Cancel the download if not completed after the specified duration in
+    /// seconds. If not set the download will never timeout.
+    #[arg(long, value_parser = parse_duration)]
+    global_timeout: Option<Duration>,
+    /// payload cid
```
jmg-duarte:
```suggestion
    /// Payload CID.
```

storage-retrieval/cli/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// Provider used for data download
+    #[arg(long)]
+    provider: Multiaddr,
+    /// The CAR file to write to.
+    #[arg(long)]
+    output: PathBuf,
+    /// Cancel the download if not completed after the specified duration in
+    /// seconds. If not set the download will never timeout.
+    #[arg(long, value_parser = parse_duration)]
+    global_timeout: Option<Duration>,
```
jmg-duarte: What is global about this?

storage-retrieval/lib/examples/simple_server.rs
jmg-duarte, line 1: Needs a module doc to explain what this is supposed to do / tries to achieve

storage-retrieval/lib/src/server.rs
```diff
@@ -0,0 +1,100 @@
...
+            }
+            SwarmEvent::ConnectionClosed {
+                peer_id,
+                connection_id,
+                ..
+            } => {
+                self.on_peer_disconnected(peer_id, connection_id);
+            }
+            _ => {
+                // Nothing to do here
```
jmg-duarte, line 79: Maybe log the unknown event?

storage-retrieval/lib/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+        .with_tokio()
+        .with_tcp(
+            tcp::Config::default(),
+            noise::Config::new,
+            yamux::Config::default,
+        )?
+        .with_behaviour(|_| Behaviour {
+            bitswap: beetswap::Behaviour::new(blockstore),
+        })
+        .expect("infallible")
```
jmg-duarte: Could you elaborate why, or is the error returned actually [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html)?

storage-retrieval/lib/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+
+const MAX_MULTIHASH_LENGHT: usize = 64;
+
+/// Custom Behaviour used by the server and client.
+#[derive(NetworkBehaviour)]
+struct Behaviour<B>
+where
+    B: Blockstore + 'static,
+{
+    bitswap: beetswap::Behaviour<MAX_MULTIHASH_LENGHT, B>,
```
jmg-duarte:
```suggestion
const MAX_MULTIHASH_LENGTH: usize = 64;

/// Custom Behaviour used by the server and client.
#[derive(NetworkBehaviour)]
struct Behaviour<B>
where
    B: Blockstore + 'static,
{
    bitswap: beetswap::Behaviour<MAX_MULTIHASH_LENGTH, B>,
```

storage-retrieval/lib/src/client.rs
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// The blockstore used by the client. It simulates a blockstore that never
+/// holds any blocks.
+struct EmptyBlockstore;
```
jmg-duarte: Empty works but I feel that `Passthrough` is more adequate since it also better explains its goal
```suggestion
struct PassthroughBlockstore;
```

storage-retrieval/lib/src/client.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    #[instrument(skip_all, fields(peer_id = %peer_id))]
+    fn on_peer_disconnected(&mut self, peer_id: PeerId, _connection_id: ConnectionId) {
+        debug!("Peer disconnected");
+    }
+
+    #[instrument(level = "trace", skip(self))]
+    async fn on_bitswap_event(&mut self, event: beetswap::Event) -> Result<(), ClientError> {
+        match event {
```
jmg-duarte, line 164: This would shorten the following code
```suggestion
        use beetswap::Event;
        match event {
```

storage-retrieval/lib/src/server.rs
```diff
@@ -0,0 +1,100 @@
...
+        _endpoint: ConnectedPoint,
+    ) {
+        debug!("Peer connected");
+    }
+
+    #[instrument(skip_all, fields(peer_id = %peer_id))]
+    fn on_peer_disconnected(&mut self, peer_id: PeerId, _connection_id: ConnectionId) {
+        debug!("Peer disconnected");
+    }
+}
```
jmg-duarte: I think there might be some value in keeping the connection id in the `instrument`

storage-retrieval/lib/src/client.rs
```diff
@@ -0,0 +1,252 @@
...
+                connection_id,
+                ..
+            } => {
+                self.on_peer_disconnected(peer_id, connection_id);
+            }
+            SwarmEvent::Behaviour(BehaviourEvent::Bitswap(event)) => {
+                self.on_bitswap_event(event).await?;
+            }
+            _ => {
+                // Nothing to do here
```
jmg-duarte, line 140: Log other events?

storage-retrieval/lib/src/client.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /// Start download of some content with a payload cid.
+    pub async fn download(mut self) -> Result<(), ClientError> {
+        // Dial all providers
+        for provider in self.providers.clone() {
+            self.swarm.dial(provider)?;
+        }
+
+        // sleep(Duration::from_secs(1)).await;
```
jmg-duarte: Leftover?

storage-retrieval/lib/src/client.rs
```diff
@@ -0,0 +1,252 @@
...
+        self.roots
+            .clone()
+            .into_iter()
+            .for_each(|root| self.request_block(root));
+
+        // Timeout future
+        let mut timeout: BoxFuture<()> = self
+            .timeout
+            .map(|t| sleep(t).boxed())
+            .unwrap_or_else(|| std::future::pending().boxed());
```
jmg-duarte: I'm confused, why not use tokio's timeout?

If it is about doing a global timeout, you can encapsulate the loop or something?

storage-retrieval/cli/src/main.rs
cernicc: I've removed the `global_`. The global was meant like the timeout for the whole download. We can implement in the future the timeout for a single provider.

storage-retrieval/lib/src/server.rs
cernicc: It's already logged by the libp2p

storage-retrieval/lib/src/lib.rs
cernicc: It's actually Infallible

storage-retrieval/lib/src/client.rs
cernicc: YES. I couldn't find a better name. Empty didn't sit quite right.

storage-retrieval/lib/src/client.rs
cernicc: I've refactored this part. Brainfreeze :D

storage-retrieval/lib/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+    #[error("Failed to initialize noise: {0}")]
+    Noise(#[from] noise::Error),
+}
+
+/// Initialize a new swarm with our custom Behaviour.
+fn new_swarm<B>(blockstore: Arc<B>) -> Result<Swarm<Behaviour<B>>, InitSwarmError>
+where
+    B: Blockstore + 'static,
+{
+    let swarm = SwarmBuilder::with_new_identity()
```
aidan46, line 36: Would we not want to initialize with an existing identity here to control the PeerID? `with_new_identity` generates a new random key for PeerID

storage-retrieval/lib/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+
+const MAX_MULTIHASH_LENGTH: usize = 64;
+
+/// Custom Behaviour used by the server and client.
+#[derive(NetworkBehaviour)]
+struct Behaviour<B>
+where
+    B: Blockstore + 'static,
+{
+    bitswap: beetswap::Behaviour<MAX_MULTIHASH_LENGTH, B>,
```
aidan46, line 20: Why wrap the `beetswap::Behaviour` here? `beetswap::Behaviour` already implements `NetworkBehaviour`. Couldn't we just use `beetswap::Behaviour` directly?

storage-retrieval/lib/src/client.rs
cernicc: They are already logged by the libp2p

storage-retrieval/lib/src/lib.rs
cernicc: I am not sure if the peer id persistence is needed here. What do you think?

storage-retrieval/lib/src/lib.rs
cernicc: The usage is nicer. If we didn't wrap it in the custom behaviour, we would have to pass the `beetswap::Behaviour` around. The wrapper is also nice for the future when you add an additional protocol. There are fewer things to change.

jmg-duarte (approved):

storage-retrieval/lib/src/lib.rs
aidan46: That makes sense

storage-retrieval/lib/src/lib.rs
aidan46: For now, maybe not, I don't want to block this PR. What I was thinking is that the client might want to make sure that they are connected with the correct PeerID/Multiaddr combination (due to a typo or something), preventing them from attempting to download something and using resources for something that will fail.

aidan46 (approved):

storage-retrieval/lib/src/lib.rs
cernicc: Aha. This is an improbable edge case. The multiaddress will usually be provided by the indexer

Issue #381, opened 2024-09-17
[Testing] remove timing dependencies on the maat tests


jmg-duarte:
Code in question — https://github.com/eigerco/polka-storage/pull/368/files#diff-ab4384b058376cec0b132538d1a3cf44d302a8db4ec9d45f6b8f6ef6bbec5c6dR39-R50

PR #684, opened 2025-01-20
test: make prove commit work with maat

### LINKED ISSUE #615: fix(maat): refactor it so it uses pipeline as a library

th7nder:
### Description

Partially fixes #615, #381.
Now maat does not depend on hard-coded values but uses the same logic as the pipeline does.
The PoSt logic will come in a follow-up PR, I don't want to make this a bigger one.

```
2025-01-22T12:53:41.132156Z DEBUG subxt_core::events: Decoding Event 'System::ExtrinsicSuccess'
2025-01-22T12:53:41.132207Z  INFO polka_storage_provider_common::sector: Successfully proven sectors on-chain: [SectorsProven { owner: AccountId32([144, 181, 171, 32, 92, 105, 116, 201, 234, 132, 27, 230, 136, 134, 70, 51, 220, 156, 168, 163, 87, 132, 62, 234, 207, 35, 20, 100, 153, 101, 254, 34]), sectors: BoundedVec([ProveCommitResult { sector_number: SectorNumber(1), partition_number: 0, deadline_idx: 0 }]) }]
test real_world_use_case ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 648.44s

   Doc-tests maat

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

It basically moves the important logic to the `UnsealedSector`, `PreCommittedSector`, `ProvenSector` types.
Those structs become empowered so they can be used in the maat.
Pipeline is just calling them and saving state.

### Checklist

  - [x] If yes, which ones?
  - [x] If yes, which ones? / List them here


storage-provider/common/src/sector.rs
```diff
@@ -0,0 +1,510 @@
...
+    ///
+    /// It is used as a randomness seed to create a PoRep.
+    /// Available at [`SectorState::Precommitted`] and later.
+    pub precommit_block: u64,
+}
+
+// TODO(@th7nder,#622,02/12/2024): query it from the chain.
+const SECTOR_EXPIRATION_MARGIN: u64 = 20;
+
+impl UnsealedSector {
```
jmg-duarte, line 79: Nit: any change the impl can be close to the struct?

storage-provider/common/src/sector.rs
```diff
@@ -0,0 +1,510 @@
...
+
+    /// Block at which the sector was precommitted (extrinsic submitted on-chain).
+    ///
+    /// It is used as a randomness seed to create a PoRep.
+    /// Available at [`SectorState::Precommitted`] and later.
+    pub precommit_block: u64,
+}
+
+// TODO(@th7nder,#622,02/12/2024): query it from the chain.
+const SECTOR_EXPIRATION_MARGIN: u64 = 20;
```
jmg-duarte: Nit: consts go at the top of their scope

jmg-duarte (dismissed):
LGTM

maat/Cargo.toml
```diff
@@ -19,23 +19,33 @@ target-debug = []
...
 
 [dependencies]
+bls12_381 = { workspace = true }
 bs58.workspace = true
 cid.workspace = true
+codec = { workspace = true }
 futures.workspace = true
 hex = { workspace = true }
+libp2p = { workspace = true, features = ["ecdsa", "identify"] }
+libp2p-identity = { workspace = true, features = ["ecdsa", "rand"] }
```
cernicc: Identity is not used(?)

storage-provider/common/src/sector.rs
cernicc: I started to like this approach too. Struct definition at the top and then impl blocks. The reader sees immediately which structs are part of the current file. But probably we shouldn't change the approach

Cargo.toml
```diff
@@ -85,6 +85,7 @@ itertools = "0.13.0"
 jsonrpsee = { version = "0.24.7" }
 libp2p = { version = "0.54", default-features = false }
 libp2p-core = "0.42.0"
+libp2p-identity = { version = "0.2.10", default-features = false }
```
cernicc: This could also be removed probably.

maat/Cargo.toml
th7nder: Yeah... I thought I needed to add it to make `PeerId::random()` work, but didn't check without adding it. Good catch.
Removing.

cernicc (approved):

jmg-duarte (approved):

PR #685, opened 2025-01-21
feat: Add re-registration on expiry for p2p node

### LINKED ISSUE #682: feat(polka-storage-provider): Add re-registration for p2p node
Currently, if running a registration node in the p2p network, the node registers for [DEFAULT_TTL](https://docs.rs/libp2p/latest/libp2p/rendezvous/constant.DEFAULT_TTL.html) seconds. Since the polka-storage-provider-server is a long-running process, we would want to add automatic re-registration after expiry.

aidan46:
### Description

Add a TTL for the registration node and auto re-register when the registration expires. The re-registration is achieved using an [Interval](https://docs.rs/tokio/1.42.0/tokio/time/struct.Interval.html) that ticks at expiry. 

Something that does not seem right to me is when re-registering, the port of the node registering changes, see the logs below. I have tried unregistering first but this messes up the connection. I have also tried adding a delay but this does not solve it either. I have looked into the code of the rendezvous [server register event](https://github.com/libp2p/rust-libp2p/blob/cc8607ce3ec494e86073c4abd6d2491280bb55d0/protocols/rendezvous/src/server.rs#L294-L303) and the [Registrations add method](https://github.com/libp2p/rust-libp2p/blob/cc8607ce3ec494e86073c4abd6d2491280bb55d0/protocols/rendezvous/src/server.rs#L386-L434), these seem to indicate that the old registration does get booted.

The above issue has been fixed by adding the external address only once. Adding the external address every time we register makes it change.

<details>
<summary>Logs from bootstrap node</summary>

```
INFO full: Peer 12D3KooWEs4rKaEAmcYpnjiPf834DTMfefeKe3SfFydepwANGWrx registered for namespace 'polka-storage' for 5 seconds
Peer Addresses: [/ip4/127.0.0.1/tcp/54401]
INFO full: Registration for peer 12D3KooWEs4rKaEAmcYpnjiPf834DTMfefeKe3SfFydepwANGWrx expired in namespace polka-storage
INFO full: Connected to 12D3KooWEs4rKaEAmcYpnjiPf834DTMfefeKe3SfFydepwANGWrx
INFO full: Peer 12D3KooWEs4rKaEAmcYpnjiPf834DTMfefeKe3SfFydepwANGWrx registered for namespace 'polka-storage' for 5 seconds
Peer Addresses: [/ip4/127.0.0.1/tcp/54402, /ip4/127.0.0.1/tcp/54401]
```

</details>

<details>
<summary>Logs from discovery node</summary>

```
INFO discover: New peer entered with addresses [
    /ip4/127.0.0.1/tcp/54838,
] peer_id=12D3KooWEs4rKaEAmcYpnjiPf834DTMfefeKe3SfFydepwANGWrx
INFO discover: Peer updated peer_id=12D3KooWEs4rKaEAmcYpnjiPf834DTMfefeKe3SfFydepwANGWrx address=/ip4/127.0.0.1/tcp/54841
INFO discover: Peer updated peer_id=12D3KooWEs4rKaEAmcYpnjiPf834DTMfefeKe3SfFydepwANGWrx address=/ip4/127.0.0.1/tcp/54842
```
</details>

### DIFF
--- a/docs/src/storage-provider-cli/server.md
+++ b/docs/src/storage-provider-cli/server.md
@@ -121,7 +121,7 @@ The supported configuration parameters are:
 | `p2p_key`                  | NA                             |
 | `rendezvous_point_address` | NA                             |
 | `rendezvous_point`         | `None`                         |
-| `registration_ttl`         | `None`                         |
+| `registration_ttl`         | `24 hours`                     |
 
 #### Bare bones configuration
 
--- a/docs/src/storage-provider-cli/server.md
+++ b/docs/src/storage-provider-cli/server.md
@@ -97,6 +97,10 @@ Rendezvous point address that the registration node connects to or the bootstrap
 
 Peer ID of the rendezvous point that the registration node connects to. Only needed if running a registration P2P node.
 
+### `--registration-ttl`
+
+The TTL of the p2p registration in seconds. After the node registration expires, the server automatically re-registers itself.
+
 ### `--config`
 
 Takes in a path to a configuration file, it supports both JSON and TOML (files _must_ have the right extension).
@@ -117,6 +121,7 @@ The supported configuration parameters are:
 | `p2p_key`                  | NA                             |
 | `rendezvous_point_address` | NA                             |
 | `rendezvous_point`         | `None`                         |
+| `registration_ttl`         | `None`                         |
 
 #### Bare bones configuration
 
--- a/storage-provider/server/src/config.rs
+++ b/storage-provider/server/src/config.rs
@@ -15,6 +15,8 @@ use crate::{
     DEFAULT_NODE_ADDRESS,
 };
 
+pub const DEFAULT_REGISTRATION_TTL: u64 = 86400;
+
 /// Default address to bind the RPC server to.
 const fn default_rpc_listen_address() -> SocketAddr {
     SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 8000)
@@ -29,6 +31,11 @@ const fn default_upload_listen_address() -> SocketAddr {
 const fn default_parallel_prove_commits() -> NonZero<usize> {
     // SAFETY: 2 != 0
     unsafe { NonZero::new_unchecked(2) }
+}
+
+/// Default registration TTL, how long the node is registered.
+const fn default_registration_ttl() -> u64 {
+    DEFAULT_REGISTRATION_TTL
 }
 
 fn default_node_address() -> Url {
@@ -125,5 +132,7 @@ pub struct ConfigurationArgs {
     pub(crate) rendezvous_point: Option<PeerId>,
 
     /// TTL of the p2p registration in seconds
-    pub(crate) registration_ttl: Option<u64>,
+    #[serde(default = "default_registration_ttl")]
+    #[arg(long, default_value_t = DEFAULT_REGISTRATION_TTL)]
+    pub(crate) registration_ttl: u64,
 }
--- a/storage-provider/server/src/config.rs
+++ b/storage-provider/server/src/config.rs
@@ -123,4 +123,7 @@ pub struct ConfigurationArgs {
     #[serde(default, deserialize_with = "string_to_peer_id_option")]
     #[arg(long)]
     pub(crate) rendezvous_point: Option<PeerId>,
+
+    /// TTL of the p2p registration in seconds
+    pub(crate) registration_ttl: Option<u64>,
 }
--- a/storage-provider/server/src/p2p/bootstrap.rs
+++ b/storage-provider/server/src/p2p/bootstrap.rs
@@ -9,7 +9,8 @@ use libp2p::{
     tcp, yamux, Multiaddr, Swarm, SwarmBuilder,
 };
 
-use super::{P2PError, TTL_24_HOURS};
+use super::P2PError;
+use crate::config::DEFAULT_REGISTRATION_TTL;
 
 #[derive(NetworkBehaviour)]
 pub struct BootstrapBehaviour {
@@ -39,7 +40,7 @@ impl BootstrapConfig {
             .with_behaviour(|key| BootstrapBehaviour {
                 // Rendezvous server behaviour for serving new peers to connecting nodes.
                 rendezvous: rendezvous::server::Behaviour::new(
-                    rendezvous::server::Config::default().with_max_ttl(TTL_24_HOURS), // Max TTL of 24 hours
+                    rendezvous::server::Config::default().with_max_ttl(DEFAULT_REGISTRATION_TTL), // Max TTL of 24 hours
                 ),
                 // The identify behaviour is used to share the external address and the public key with connecting clients.
                 identify: identify::Behaviour::new(identify::Config::new(
--- a/storage-provider/server/src/p2p/bootstrap.rs
+++ b/storage-provider/server/src/p2p/bootstrap.rs
@@ -66,11 +66,8 @@ pub(crate) async fn bootstrap(
     swarm.listen_on(addr)?;
     while let Some(event) = swarm.next().await {
         match event {
-            SwarmEvent::ConnectionEstablished { peer_id, .. } => {
-                tracing::info!("Connected to {}", peer_id);
-            }
-            SwarmEvent::ConnectionClosed { peer_id, .. } => {
-                tracing::info!("Disconnected from {}", peer_id);
+            SwarmEvent::NewListenAddr { address, .. } => {
+                tracing::info!("Listening on {}", address);
             }
             SwarmEvent::Behaviour(BootstrapBehaviourEvent::Rendezvous(
                 rendezvous::server::Event::PeerRegistered { peer, registration },
@@ -105,7 +102,7 @@ pub(crate) async fn bootstrap(
                     registration.namespace
                 );
             }
-            _other => {}
+            other => tracing::debug!("Encountered event: {other:?}"),
         }
     }
     Ok(())
--- a/storage-provider/server/src/p2p/bootstrap.rs
+++ b/storage-provider/server/src/p2p/bootstrap.rs
@@ -9,7 +9,7 @@ use libp2p::{
     tcp, yamux, Multiaddr, Swarm, SwarmBuilder,
 };
 
-use super::P2PError;
+use super::{P2PError, TTL_24_HOURS};
 
 #[derive(NetworkBehaviour)]
 pub struct BootstrapBehaviour {
@@ -39,7 +39,7 @@ impl BootstrapConfig {
             .with_behaviour(|key| BootstrapBehaviour {
                 // Rendezvous server behaviour for serving new peers to connecting nodes.
                 rendezvous: rendezvous::server::Behaviour::new(
-                    rendezvous::server::Config::default(),
+                    rendezvous::server::Config::default().with_max_ttl(TTL_24_HOURS), // Max TTL of 24 hours
                 ),
                 // The identify behaviour is used to share the external address and the public key with connecting clients.
                 identify: identify::Behaviour::new(identify::Config::new(
@@ -94,6 +94,15 @@ pub(crate) async fn bootstrap(
                         registrations.len()
                     );
                 }
+            }
+            SwarmEvent::Behaviour(BootstrapBehaviourEvent::Rendezvous(
+                rendezvous::server::Event::RegistrationExpired(registration),
+            )) => {
+                tracing::info!(
+                    "Registration for peer {} expired in namespace {}",
+                    registration.record.peer_id(),
+                    registration.namespace
+                );
             }
             _other => {}
         }
--- a/storage-provider/server/src/p2p/mod.rs
+++ b/storage-provider/server/src/p2p/mod.rs
@@ -15,8 +15,10 @@ pub(crate) use bootstrap::BootstrapConfig;
 pub(crate) use register::RegisterConfig;
 
 const P2P_NAMESPACE: &str = "polka-storage";
+const TTL_24_HOURS: u64 = 86400;
 
 #[derive(Default, Debug, Clone, Copy, ValueEnum, Deserialize)]
+#[serde(rename_all = "lowercase")]
 pub enum NodeType {
     #[default]
     Bootstrap,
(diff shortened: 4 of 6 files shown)


docs/src/storage-provider-cli/server.md
```diff
@@ -117,6 +121,7 @@ The supported configuration parameters are:
 | `p2p_key`                  | NA                             |
 | `rendezvous_point_address` | NA                             |
 | `rendezvous_point`         | `None`                         |
+| `registration_ttl`         | `None`                         |
```
jmg-duarte: What does `None` mean? Is it "never re-register"?

storage-provider/server/src/p2p/mod.rs
```diff
@@ -141,14 +146,20 @@ pub async fn run_register_node(
 ) -> Result<(), P2PError> {
     tracing::info!("Starting P2P register node");
     let tracker = TaskTracker::new();
-    let (swarm, rendezvous_point_address, rendezvous_point) = config.create_swarm()?;
+    // If TTL is not set, set it to 24 hours
+    let ttl = if config.registration_ttl.is_some() {
+        config.registration_ttl
+    } else {
+        Some(TTL_24_HOURS)
+    };
```
jmg-duarte: This goes against `None` as a default and makes the code harder to follow since the default is set late. Just use `TTL_24_HOURS` as a default at the CLI level

storage-provider/server/src/p2p/register.rs
```diff
@@ -66,26 +69,39 @@ impl RegisterConfig {
...
-    mut swarm: Swarm<RegisterBehaviour>,
+    swarm: &mut Swarm<RegisterBehaviour>,
     rendezvous_point: PeerId,
     rendezvous_point_address: Multiaddr,
     ttl: Option<u64>,
     namespace: Namespace,
 ) -> Result<(), P2PError> {
     tracing::info!("Attempting to register with rendezvous point {rendezvous_point} at {rendezvous_point_address}");
-    swarm.dial(rendezvous_point_address.clone())?;
+    let mut register_tick = tokio::time::interval(Duration::from_secs(ttl.unwrap_or(TTL_24_HOURS)));
```
jmg-duarte: Just highlighting this to further drive my point

storage-provider/server/src/config.rs
```diff
@@ -123,4 +123,7 @@ pub struct ConfigurationArgs {
     #[serde(default, deserialize_with = "string_to_peer_id_option")]
     #[arg(long)]
     pub(crate) rendezvous_point: Option<PeerId>,
+
+    /// TTL of the p2p registration in seconds
+    pub(crate) registration_ttl: Option<u64>,
```
jmg-duarte:
```suggestion
    #[serde(default = "<const fn that_returns_TTL_24_HOURS>"]
    #[arg(long, default_value_t = TTL_24_HOURS)]
    pub(crate) registration_ttl: Option<u64>,
```

jmg-duarte (changes requested):

storage-provider/server/src/config.rs
cernicc: Note. If you don't write `#[arg(long)]` then the ttl is accepted as an argument and not an option. 

```
Usage: polka-storage-provider-server [OPTIONS] --p2p-key <P2P_KEY> --rendezvous-point-address <RENDEZVOUS_POINT_ADDRESS> [REGISTRATION_TTL]

Arguments:
  [REGISTRATION_TTL]
          TTL of the p2p registration in seconds

Options:
      --sr25519-key <SR25519_KEY>
          Sr25519 keypair, encoded as hex, BIP-39 or a dev phrase like `//Alice`.
```

storage-provider/server/src/p2p/register.rs
```diff
@@ -66,26 +69,39 @@ impl RegisterConfig {
...
 
+    loop {
```
cernicc, line 95: This will never loop. It could loop if the swarm was exhausted but that will never happen.

storage-provider/server/src/p2p/register.rs
```diff
@@ -59,33 +61,46 @@ impl RegisterConfig {
...
-    ttl: Option<u64>,
+    ttl: u64,
     namespace: Namespace,
 ) -> Result<(), P2PError> {
     tracing::info!("Attempting to register with rendezvous point {rendezvous_point} at {rendezvous_point_address}");
-    swarm.dial(rendezvous_point_address.clone())?;
+    let mut register_tick = tokio::time::interval(Duration::from_secs(ttl));
+
+    loop {
+        register_tick.tick().await;
```
cernicc: Imagine that the `register_tick` was 24 hours. The first iteration would hang for the 24 hours on this call. That also means that the first registration would happen after 24 hours?

storage-provider/server/src/p2p/register.rs
aidan46: The `register_and_check_events` returns `Ok(())` once the registration is successful and an error if anything goes wrong, propagating the error to this function with `?`. If the function exits successfully the loop continues ([ControlFlow::Continue](https://doc.rust-lang.org/stable/std/ops/enum.ControlFlow.html#variant.Continue)) and waits for the tick.

storage-provider/server/src/p2p/register.rs
aidan46: The first tick for `interval` [completes immediately](https://docs.rs/tokio/latest/tokio/time/fn.interval.html). There is an `interval_at` [function which takes in a start before the first tick](https://docs.rs/tokio/latest/tokio/time/fn.interval_at.html).

storage-provider/server/src/p2p/register.rs
cernicc: You are correct. I missed the `return Ok(());` line

storage-provider/server/src/p2p/register.rs
cernicc: I still think this setup won't work correctly. The thing is that the swarm is only progressing when it's polled. And in this case it wont be pulled. It won't be doing anything for the 24hours. Basically a node that would be dead for the tick period.

https://docs.rs/libp2p/latest/libp2p/swarm/struct.Swarm.html
`Note: Needs to be polled via <Swarm as Stream> in order to make progress.`

storage-provider/server/src/p2p/bootstrap.rs
```diff
@@ -95,6 +96,15 @@ pub(crate) async fn bootstrap(
                     );
                 }
             }
+            SwarmEvent::Behaviour(BootstrapBehaviourEvent::Rendezvous(
+                rendezvous::server::Event::RegistrationExpired(registration),
+            )) => {
+                tracing::info!(
+                    "Registration for peer {} expired in namespace {}",
```
th7nder, line 100: What do we do when the registration expires? What does it mean for the node to be registered?

th7nder (dismissed):

storage-provider/server/src/p2p/register.rs
aidan46: I have added a `tokio::select` so the swarm keeps polling and not just waiting for the tick to complete.

storage-provider/server/src/p2p/bootstrap.rs
aidan46: When the registration expires the registration node re-registers automatically.
A node being registered means that they have identified themselves with the bootstrap node and shared their [Registration](https://docs.rs/libp2p/latest/libp2p/rendezvous/struct.Registration.html) with them. This Registration holds information like their Peer ID and multiaddr.

storage-provider/server/src/p2p/register.rs
```diff
@@ -59,83 +61,98 @@ impl RegisterConfig {
...
+                if let Err(error) =
+                    swarm
+                        .behaviour_mut()
+                        .rendezvous
+                        .register(namespace.clone(), rendezvous_point, Some(ttl))
+                {
                     tracing::error!("Failed to register: {error}");
                     return Err(P2PError::RegistrationFailed(rendezvous_point));
+                } else {
+                    tracing::info!("Registration with {rendezvous_point} successful");
```
cernicc: I think that the registration is not yet confirmed on this line.
```suggestion
                    tracing::info!("Registration requested with {rendezvous_point}");
```

storage-provider/server/src/p2p/register.rs
```diff
@@ -59,83 +61,98 @@ impl RegisterConfig {
...
-                    namespace,
-                    error,
-                },
+            // Check incoming event.
+            event = swarm.select_next_some() => check_swarm_event(event)
+        }
+    }
+}
+
+fn check_swarm_event(event: SwarmEvent<RegisterBehaviourEvent>) {
```
cernicc:
```suggestion
fn on_swarm_event(event: SwarmEvent<RegisterBehaviourEvent>) {
```

storage-provider/server/src/p2p/register.rs
```diff
@@ -59,83 +61,98 @@ impl RegisterConfig {
...
+            );
+        }
+        SwarmEvent::Behaviour(RegisterBehaviourEvent::Rendezvous(
+            rendezvous::client::Event::RegisterFailed {
+                rendezvous_node,
+                namespace,
+                error,
+            },
+        )) => {
+            tracing::error!(%rendezvous_node, %namespace,
```
cernicc, line 147: Should we return an error here?

storage-provider/server/src/p2p/mod.rs
```diff
@@ -141,14 +145,17 @@ pub async fn run_register_node(
 ) -> Result<(), P2PError> {
     tracing::info!("Starting P2P register node");
     let tracker = TaskTracker::new();
```
cernicc: This tracker and the one in `run_bootstrap_node` are not used

storage-provider/server/src/config.rs
```diff
@@ -123,4 +130,9 @@ pub struct ConfigurationArgs {
     #[serde(default, deserialize_with = "string_to_peer_id_option")]
     #[arg(long)]
     pub(crate) rendezvous_point: Option<PeerId>,
+
+    /// TTL of the p2p registration in seconds
+    #[serde(default = "default_registration_ttl")]
+    #[arg(long, default_value_t = DEFAULT_REGISTRATION_TTL)]
+    pub(crate) registration_ttl: u64,
```
cernicc, line 137: I think it would be better if we use Duration here. That way the reader can immediately see how the duration is parsed

storage-provider/server/src/p2p/register.rs
aidan46: Yep, you are right, it has only been requested at this point.

storage-provider/server/src/config.rs
aidan46: For the tick it makes sense but we re-use this value for the [register call in the swarm which expects a](https://docs.rs/libp2p/latest/libp2p/rendezvous/client/struct.Behaviour.html#method.register) `Option<u64>`

storage-provider/server/src/p2p/bootstrap.rs
```diff
@@ -95,7 +93,16 @@ pub(crate) async fn bootstrap(
...
+                    registration.record.peer_id(),
+                    registration.namespace
+                );
+            }
+            other => tracing::debug!("Encountered event: {other:?}"),
```
jmg-duarte, line 105: Won't this catch the previous `ConnectionEstablished` and `ConnectionClosed` events?

storage-provider/server/src/p2p/register.rs
```diff
@@ -59,83 +61,100 @@ impl RegisterConfig {
...
-                    namespace.clone(),
-                    rendezvous_point,
-                    ttl,
-                ) {
+    loop {
+        tokio::select! {
+            // Poll tick every TTL to re-register.
+            // First tick completes immediately.
+            _ = register_tick.tick() => {
+                tracing::info!("Registering with p2p node");
```
jmg-duarte, line 100: I think @cernicc knows a bit more than me on this, but I think these logs should be contained inside an `#[instrument]` or a span to keep context

storage-provider/server/src/p2p/register.rs
```diff
@@ -59,83 +61,100 @@ impl RegisterConfig {
...
+        )) => {
+            tracing::error!(%rendezvous_node, %namespace,
+                "Failed to register error = {error:?}"
+            );
+            return Err(P2PError::RegistrationFailed(rendezvous_node));
+        }
+        other => tracing::debug!("Encountered event: {other:?}"),
+    }
     Ok(())
 }
```
jmg-duarte, line 155: A docstring would be appreciated

storage-provider/server/src/config.rs
jmg-duarte: It's ok as is, but for reference, you can always just [`.as_secs`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs)

jmg-duarte (commented):
It's ok as is, but there's just a bit of room for improvement

jmg-duarte (dismissed):
(I meant to approve)

storage-provider/server/src/p2p/mod.rs
```diff
@@ -158,11 +164,10 @@ pub async fn run_register_node(
         },
         _ = token.cancelled() => {
             tracing::info!("P2P node has been stopped by the cancellation token...");
+            tracker.close();
+            tracker.wait().await;
```
cernicc: What is the reason for using TaskTracker here?

jmg-duarte: Good catch, I don't even see what task this is tracking. I think this might be a leftover from an experiment?

storage-provider/server/src/p2p/bootstrap.rs
aidan46: It will catch those events. I decided to reduce the log level on this as it is not very relevant to the protocol and reduces the log spam. Any event we are interested in is logged at `info!` level.

storage-provider/server/src/p2p/mod.rs
cernicc: This task TaskTracker is still unused.

aidan46: Oops, I removed it in the bootstrap node but forgot it here

cernicc (approved):

jmg-duarte (approved):
Looks great 🚀 good job!

Issue #675, opened 2025-01-14
feat: Mater `convert`, wrap contents in unixfs dag


cernicc:
The client retrieval system expects a car file with dag information so that it knows which blocks to request from the server. The current issue is that when the client receives a root node. It doesn't know which blocks are next in the tree to be requested.

The current `convert` command converts contents to a car, but it doesn't encode any dag nodes metadata. That should be changed so that the nodes that are written to car also at least contain links to other nodes.

jmg-duarte:
Even though this is no longer relevant for retrieval right now, I'm keeping this one open since this is a difference between mater and go-car that we can address

PR #688, opened 2025-01-21
feat: sector indexing and data retrieval

### LINKED ISSUE #635: Index sectors and pieces on the storage provider
For storage retrieval purposes. The provider needs to index sectors and data inside.

### LINKED ISSUE #680: feat: retrieval server blockstore
Implement a blockstore which the retrieval server will use. The blockstore `get(cid: Cid)` will be called when some client requests a block with the cid.

The blockstore should be able to read the blocks directly from the unsealed sectors.

Blocker: #635

cernicc:
### Description

Add sector indexing on the successful prove commitment and integrates retrieval server that uses the indexing when retrieving blocks.

The #675 should be solved before this is merged. The reason is that the client doesn't know currently which blocks it should request after the root node.

### Checklist


mater/lib/src/cid.rs
```diff
@@ -3,8 +3,10 @@ use tokio::io::{AsyncRead, AsyncReadExt};
 
 use crate::{async_varint::read_varint, IDENTITY_CODE};
 
+/// Extension trait for Cid
 pub trait CidExt {
-    async fn read_bytes_async<R>(r: R) -> Result<(Self, usize), Error>
+    /// Reads the bytes from a byte stream.
+    fn read_bytes_async<R>(r: R) -> impl std::future::Future<Output = Result<(Self, usize), Error>>
```
th7nder, line 9: Why can't we use `async fn` here? async in traits is stabilized for quite some time.

mater/lib/src/v1/mod.rs
```diff
@@ -66,12 +66,12 @@ impl Default for Header {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub struct BlockMetadata {
     /// Cid of the block
-    cid: Cid,
+    pub cid: Cid,
```
th7nder: I don't think making those fields public is a good idea. I might be too Javaesque, however I'd prefer to expose those fields as getters.

storage-provider/common/src/sector.rs
```diff
@@ -311,12 +327,11 @@ impl PreCommittedSector {
         seal_randomness_height: u64,
         precommit_block: u64,
     ) -> Result<Self, std::io::Error> {
-        tokio::fs::remove_file(unsealed.unsealed_path).await?;
```
th7nder: We don't need the `unsealed sector` anymore I think? We're storing them raw anyways.
I think we can remove this like we did.

storage-provider/server/src/indexer/local_index_directory/rdb.rs
```diff
@@ -107,6 +107,7 @@ impl RocksDBLid {
     /// * If the column families ([`PIECE_CID_TO_CURSOR_CF`],
     ///   [`MULTIHASH_TO_PIECE_CID_CF`], [`PIECE_CID_TO_FLAGGED_CF`],
     ///   [`CURSOR_TO_OFFSET_SIZE_CF`]) do not exist, they will be created.
+    /// * If the cursor is not initialized. It will be initialized with the 100 value.
```
th7nder: How come? What does 100 mean?

storage-provider/server/src/indexer/local_index_directory/rdb.rs
```diff
@@ -274,7 +293,8 @@ impl RocksDBLid {
     /// Source:
     /// * <https://github.com/filecoin-project/boost/blob/16a4de2af416575f60f88c723d84794f785d2825/extern/boostd-data/ldb/db.go#L109-L118>
     fn get_next_cursor(&self) -> Result<(u64, String), LidError> {
-        let pinned_slice = self.database.get_pinned(NEXT_CURSOR_KEY)?;
+        let pinned_slice: Option<rocksdb::DBPinnableSlice> =
```
th7nder: nit: do we need the type annotation here?

storage-provider/server/src/indexer/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+/// Prepares indexes of a raw piece.
+async fn piece_indexes<P>(location: P) -> Result<Vec<IndexRecord>, ServerError>
+where
+    P: AsRef<Path>,
+{
+    let blocks = stream_blocks_metadata(&location).await?;
+    pin_mut!(blocks);
+
+    let mut records = vec![];
+    if let Some(metadata) = blocks.next().await {
```
th7nder: If I understood this code correctly, it indexes only one block inside of the location.
`while let Some(..)` seems more likely here.

storage-provider/server/src/indexer/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        .await
+        {
+            Ok(Ok(_)) => {
+                info!(%piece_cid, "indexing completed");
+            }
+            Ok(Err(err)) => {
+                error!(?err, "piece indexing failed with an error");
+            }
+            Err(err) => {
+                error!(?err, "piece indexing panicked");
```
th7nder: adding piece_cid here would probably also be useful.

storage-provider/server/src/pipeline/mod.rs
```diff
@@ -169,14 +175,21 @@ impl PipelineOperations for TaskTracker {
...
                     tracing::info!(
                         "ProveCommit for sector {} finished successfully.",
                         sector_number
-                    )
+                    );
+
+                    // Start indexing the sector
+                    if let Err(err) = indexer_tx.send(IndexMessage::IndexSector(sector)) {
+                        error!(?err, "error occurred while messaging the indexer");
+                    }
```
th7nder: I don't think we should start indexing this late, *after prove commit is finished*.
From my latest tests, pre-committing 512MB sectors on my machine is like ~30minutes, so around  ~300 blocks.
Didn't test prove-commiting yet, however it'll probably take another 30 minutes.

This latency is kind-of intolerable (?), I think (?). 
Imagine, you upload a file and you can fetch it only 60 minutes after, depends how long you want to store it, but still.
If we could make this indexing faster, or incremental (after it's assigned to the sector, so straight after add_piece), I think it'd be better.

What do you think?

storage-retrieval/lib/examples/simple_server.rs
```diff
@@ -23,7 +23,7 @@ async fn main() -> Result<()> {
     // Setup & run the server
     let server = Server::new(blockstore)?;
     let listener: Multiaddr = format!("/ip4/127.0.0.1/tcp/8989").parse()?;
-    server.run(vec![listener]).await?;
+    server.run(vec![listener], std::future::pending()).await?;
```
th7nder, line 26: It makes the process unkillable (?)

storage-retrieval/lib/src/server.rs
```diff
@@ -43,17 +43,35 @@ where
...
+    where
+        F: Future<Output = ()> + Send + 'static,
+    {
         // Listen on
         for listener in listeners {
             self.swarm.listen_on(listener)?;
         }
 
-        // Keep server running
+        pin_mut!(shutdown_signal);
```
th7nder, line 59: We do we need to pin it?

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,123 @@
...
+            raw_pieces_dir: raw_pieces_dir.into(),
+        }
+    }
+}
+
+impl<I> blockstore::Blockstore for ProviderBlockstore<I>
+where
+    I: Service + Send + Sync + 'static,
+{
+    async fn get<const S: usize>(
```
th7nder, line 37: The `cid` is the Cid of a block inside the CAR file?

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    async fn put_keyed<const S: usize>(
+        &self,
+        _cid: &cid::CidGeneric<S>,
+        _data: &[u8],
+    ) -> blockstore::Result<()> {
+        Err(Error::FatalDatabaseError(
+            "put operation not supported".to_string(),
+        ))
```
th7nder:
```suggestion
        unimplemented!("blockstore for retrieval is read-only")
```

It's a programming error that someone tries to use it like that.

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,123 @@
...
+        };
+
+        // Get the offset of the block within the piece (CAR file)
+        let block = self
+            .indexer
+            .get_offset_size(*piece_cid, *cid.hash())
+            .map_err(|err| blockstore::Error::StoredDataError(err.to_string()))?;
+
+        // Open the raw piece and read the block
+        let raw_pieces_dir = self
```
th7nder:
```suggestion
        let raw_piece_path = self
```

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,123 @@
...
+            .raw_pieces_dir
+            .join(piece_cid.to_string())
+            .with_extension("car");
+        let mut raw_piece = File::open(&raw_pieces_dir).await.map_err(|_| {
+            blockstore::Error::StoredDataError(format!("piece {:?} not found", raw_pieces_dir))
+        })?;
+        raw_piece
+            .seek(SeekFrom::Start(block.offset))
+            .await
+            .map_err(|_| {
```
th7nder, line 79: Probably not so performant right now. Imagine you have a file with 1000 blocks, then you need to do 1k seeks and 1k file opens in short period of time.

I wouldn't worry about it now, but should we add a note?

th7nder (changes requested):
Huuge piece of work!
I left some major questions in the comments and have another one.
How is this thing tested?

storage-provider/server/src/indexer/local_index_directory/rdb.rs
```diff
@@ -264,6 +271,18 @@ impl RocksDBLid {
         Ok(multihash)
     }
 
+    /// Initialize cursor with the default value if not set already.
+    fn init_cursor(&self) -> Result<(), LidError> {
+        if let Err(err) = self.get_next_cursor() {
+            if matches!(err, LidError::CursorNotFound) {
+                self.set_next_cursor(100)?;
```
aidan46: Why 100? Is that the default value?

storage-provider/server/src/config.rs
```diff
@@ -42,6 +42,13 @@ fn default_node_address() -> Url {
     Url::parse(DEFAULT_NODE_ADDRESS).expect("DEFAULT_NODE_ADDRESS must be a valid Url")
 }
 
+fn default_retrieval_address() -> Multiaddr {
+    "/ip4/127.0.0.1/tcp/8002"
+        .to_string()
+        .parse()
+        .expect("multiaddres is correct")
```
aidan46:
```suggestion
    "/ip4/127.0.0.1/tcp/8002"
        .parse()
        .expect("multiaddres is correct")
```
[No need to convert to string first.](https://docs.rs/libp2p/latest/libp2p/struct.Multiaddr.html#impl-TryFrom%3C%26str%3E-for-Multiaddr)

storage-provider/common/src/sector.rs
aidan46: Why were we removing the file before and not anymore now?

cernicc: The first idea was to index the unsealed sectors. This is a leftover from before.

jmg-duarte:
Converted to draft since it's blocked

storage-provider/server/src/retrieval/blockstore.rs
cernicc: I kind of agree.  This comment was the reason of returning an error. https://github.com/eigerco/polka-storage/pull/660#discussion_r1900715460

storage-provider/server/src/retrieval/blockstore.rs
cernicc: Yes. The cid is from a specific block in the car file

storage-provider/server/src/indexer/local_index_directory/rdb.rs
cernicc: Good catch. Yeah. This is the default value with which the database is initialized.

mater/lib/src/cid.rs
cernicc: We can. The difference is that the `async fn` adds an automatic Send bound on the returned future. That might not be desired in some cases. This way we still have an option to conditionally add a bound if needed.

https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html

mater/lib/src/v1/mod.rs
cernicc: I don't think getters are needed here. The [CarV1Header](https://github.com/eigerco/polka-storage/blob/26d6c611b31d39b9740fa628a9e0b88eec9899bf/mater/lib/src/v1/mod.rs#L22) and [CarV2Header](https://github.com/eigerco/polka-storage/blob/26d6c611b31d39b9740fa628a9e0b88eec9899bf/mater/lib/src/v2/mod.rs#L64) are already exposing their internal fields the same way. 

It's not an issue because the `BlockMetadata` is valid nevertheless of the combination of its internal values. e.g. `BlockMetadata {cid: <some valid cid>, data_offset_source: 87987, data_size: 10000}` is valid like `BlockMetadata {cid: <some other valid cid>, data_offset_source: 738276, data_size: 8798}` is also valid the same way.

storage-provider/common/src/sector.rs
cernicc: Good catch. This was a leffover from the sector indexing impl.

storage-provider/server/src/indexer/local_index_directory/rdb.rs
cernicc: I've changed the initial cursor to 0. The 100 was used in the reference impl.

storage-provider/server/src/indexer/local_index_directory/rdb.rs
cernicc: It's not needed. I probably double-clicked on the inspected type 😄

storage-provider/server/src/indexer/mod.rs
cernicc: Uffff. Great catch

storage-retrieval/lib/examples/simple_server.rs
cernicc: Yes. `shutdown_signal` is a future when resolved the server is gracefully shutdown. In the case above that future is never resolved.

storage-retrieval/lib/src/server.rs
cernicc: We need to pin it because we are using the same future instance in each iteration of the loop. So we must be sure that the future will not move while we are waiting for it.

The other explanation is that we are using the `&mut shutdown_signal` in the `select!` and the `&mut F` has to implement the Future trait. The [Future](https://doc.rust-lang.org/std/future/trait.Future.html) trait requires that the `Self` is pinned before it can be pulled.

storage-provider/server/src/retrieval/blockstore.rs
cernicc: I agree. We can't get rid of the seeking part. But we could hold a pool of open file handlers. I've added the comment.

cargopete:
awesome work man, a few questions more for my own understanding than for the  exact implementation but still

1. How are we ensuring that the local index remains consistent with on‑chain state? For example, in case of a prove commit failure or a chain reorganization, do we have a recovery or sanity‑check mechanism in place? Could we add documentation or even a periodic reconciliation process to handle such cases? 
2. I know it's been sorta discussed by now but currently, indexing happens after prove commit, which can take a long time for large sectors. Would it be possible to start indexing incrementally (e.g., immediately after a piece is added) so that retrieval can be available sooner?
3. error handling is spread out across the RPC, indexer, and retrieval modules. Would it be beneficial to define a common error type or conversion strategy so that higher‑level components can manage errors more uniformly?
4. Given that this PR touches multiple areas (indexing, retrieval, the pipeline, etc.), it might help if we add a high‑level diagram or brief design document outlining the data flow from piece addition through indexing to retrieval
5. The retrieval blockstore currently performs a file open and seek for each block access. With large files containing many blocks, this might become a performance bottleneck. Should we consider a strategy like pooling file handles or caching recent file accesses, or at least add a TODO note to revisit this once we have performance metrics?

storage-provider/server/src/pipeline/mod.rs
th7nder: Ok, I'm actually pulling out of this comment.
This is the distributed storage, we're not aiming for instant access, this is what D3L of JAM will be for.
It's fine if the file will be accessible a bit later.
File should be accessible when `start_block` is there, so after prove commit is totally fine.

cernicc:
@pete-eiger Thanks for the review 👍 My answers are bellow.

> awesome work man, a few questions more for my own understanding than for the exact implementation but still
> 
> 1. How are we ensuring that the local index remains consistent with on‑chain state? For example, in case of a prove commit failure or a chain reorganization, do we have a recovery or sanity‑check mechanism in place? Could we add documentation or even a periodic reconciliation process to handle such cases?
> 2. I know it's been sorta discussed by now but currently, indexing happens after prove commit, which can take a long time for large sectors. Would it be possible to start indexing incrementally (e.g., immediately after a piece is added) so that retrieval can be available sooner?
> 3. error handling is spread out across the RPC, indexer, and retrieval modules. Would it be beneficial to define a common error type or conversion strategy so that higher‑level components can manage errors more uniformly?
> 4. Given that this PR touches multiple areas (indexing, retrieval, the pipeline, etc.), it might help if we add a high‑level diagram or brief design document outlining the data flow from piece addition through indexing to retrieval
> 5. The retrieval blockstore currently performs a file open and seek for each block access. With large files containing many blocks, this might become a performance bottleneck. Should we consider a strategy like pooling file handles or caching recent file accesses, or at least add a TODO note to revisit this once we have performance metrics?

1. This was the thing that we discussed on the meeting. There will have to be some part of the system that will remove indexes when something happens eg. the deal is filled, etc. The "garbage collection" system is not part of this pr.
2. Currently the indexing is started when the piece is uploaded to the storage provider. [Here](https://github.com/eigerco/polka-storage/pull/688#discussion_r1947304379) Konrad suggested I change it back to the prove commit. Do you think the files should be available before the prove commit is finished?
3. What do you mean with this comment? The indexer already returns a common `ServerError`, where an error should be returned to the caller.
4. The process was briefly explained in `rfcs/002-data-retrieval.md`. I agree that some additional diagram would improve the rfc.
5. This is also the thing that Konrad exposed [here](https://github.com/eigerco/polka-storage/pull/688#discussion_r1939398798). [This](https://github.com/eigerco/polka-storage/blob/11a95a14fbc88cfbb563b777505a05894d37a9d8/storage-provider/server/src/retrieval/blockstore.rs#L68) was the `TODO` I've added.

storage-provider/server/src/pipeline/mod.rs
cernicc: Good point. I've changed it back

storage-provider/server/src/retrieval/blockstore.rs
th7nder: okk

th7nder (dismissed):

mater/lib/src/v1/mod.rs
jmg-duarte: I agree with Rok

mater/lib/src/cid.rs
```diff
@@ -3,8 +3,10 @@ use tokio::io::{AsyncRead, AsyncReadExt};
 
 use crate::{async_varint::read_varint, IDENTITY_CODE};
 
+/// Extension trait for Cid
```
jmg-duarte: Make it clickable in the docs I guess
```suggestion
/// Extension trait for [`Cid`]
```

mater/lib/src/cid.rs
```diff
@@ -13,8 +15,10 @@ pub trait CidExt {
     fn get_identity_data(&self) -> Option<&[u8]>;
 }
 
+/// Extension trait for Multihash
```
jmg-duarte:
```suggestion
/// Extension trait for [`Multihash`]
```

storage-provider/server/src/indexer/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+/// Prepares indexes of a raw piece.
+async fn piece_indexes<P>(location: P) -> Result<Vec<IndexRecord>, ServerError>
+where
+    P: AsRef<Path>,
+{
+    let file = File::open(location).await?;
+    let reader = BufReader::new(file);
+    let blocks = stream_blocks_metadata(reader).await?;
```
jmg-duarte: I don't find the extra code for `stream_blocks_metadata` strictly necessary. 
It's abstracting over reading the pragma, header and friends and then doing a loop over the blocks. 

I feel that just performing the "upfront" overhead using a CAR reader would remove the whole `utils.rs` module and make code less opaque

storage-provider/common/src/sector.rs
```diff
@@ -70,6 +70,11 @@ pub struct UnsealedSector {
     /// Indexes match with corresponding deals in [`Sector::deals`].
     pub piece_infos: Vec<PieceInfo>,
 
+    /// Tracks locations of the actual pieces added to the unsealed sector. This
+    /// vector does not contain padding pieces. It contains the actual pieces
+    /// corresponding with the deals from the users.
+    pub pieces_locations: Vec<(Commitment<CommP>, PathBuf)>,
```
jmg-duarte, line 76: For now this is ok, but for really big sectors, this should be in a separate column family.

It would simplify managing this structure & make serialization faster at the cost of an extra lookup

storage-provider/server/src/indexer/local_index_directory/rdb.rs
```diff
@@ -264,6 +271,18 @@ impl RocksDBLid {
...
+    /// Initialize cursor with the default value if not set already.
+    fn init_cursor(&self) -> Result<(), LidError> {
+        if let Err(err) = self.get_next_cursor() {
+            if matches!(err, LidError::CursorNotFound) {
+                self.set_next_cursor(0)?;
+            } else {
+                return Err(err);
+            }
+        }
+        Ok(())
```
jmg-duarte:
```suggestion
        match self.get_next_cursor() {
            Ok(_) => Ok(()),
            Err(LidError::CursorNotFound) => self.set_next_cursor(0),
            Err(err) => Err(err),
        }
```

storage-provider/server/src/indexer/local_index_directory/rdb.rs
```diff
@@ -328,6 +347,12 @@ impl RocksDBLid {
         // as long as it doesnt fail
         for it in iterator {
             let (key, _) = it?;
+
+            // Filter out the keys not prefixed with the cursor
+            if !key.as_ref().starts_with(cursor_prefix.as_bytes()) {
+                continue;
+            }
```
jmg-duarte: If the prefix iterator stops yielding keys matching the prefix, you shouldn't ever receive another one with a prefix — i.e. if the iterator for prefix "/" returns a key like "@", no future keys should start with "/" making it so you should call `break` here instead. No?

storage-provider/server/src/indexer/local_index_directory/rdb.rs
```diff
@@ -600,6 +625,12 @@ impl Service for RocksDBLid {
         let mut records = vec![];
         for it in iterator {
             let (key, value) = it?;
+
+            // Filter out the keys not prefixed with the cursor
+            if !key.as_ref().starts_with(cursor_prefix.as_bytes()) {
+                continue;
+            }
```
jmg-duarte: Same as above

storage-provider/server/src/indexer/local_index_directory/rdb_ext.rs
jmg-duarte, line 1: IMO we can leave these with the `rdb.rs` file and move them out if they become too much

storage-provider/server/src/indexer/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+pub mod local_index_directory;
+
+#[derive(Debug, Clone)]
+pub enum IndexerMessage {
+    /// Start indexing the sector
+    IndexSector(ProvenSector),
+}
+
+#[derive(Clone)]
+pub struct IndexerState<D> {
```
jmg-duarte: This `D` seems kinda random, what does the `D` stand for?

storage-provider/server/src/indexer/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+#[instrument(skip_all, fields(piece_cid = %commitment.cid()))]
+async fn index_piece<D, P>(db: Arc<D>, commitment: Commitment<CommP>, piece_path: P)
+where
+    D: Service + Send + Sync + 'static,
+    P: AsRef<Path>,
+{
+    let records = match piece_indexes(piece_path).await {
+        Ok(records) => records,
+        Err(err) => {
+            error!(?err, "piece indexing failed with an error");
```
jmg-duarte: Log the respective piece path too, can be useful for debugging

storage-provider/server/src/indexer/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    .await
+    {
+        Ok(Ok(_)) => {
+            info!("indexing completed");
+        }
+        Ok(Err(err)) => {
+            error!(?err, "piece indexing failed with an error");
+        }
+        Err(err) => {
+            error!(?err, "piece indexing panicked");
```
jmg-duarte: Is this 100% accurate? Shouldn't a panic bring down the program since we're not doing any panic handlers?

storage-provider/server/tests/fixtures/spaceglenda_wrapped_v2.car
jmg-duarte, line 1: This file is a duplicate of the mater tests, do we really need to check it in? Could we, at the very least, add a symbolic link instead

storage-retrieval/lib/examples/simple_server.rs
jmg-duarte: Could you doc that just so it's made obvious it was intentional and none of us takes it out?

storage-retrieval/lib/examples/simple_server.rs
jmg-duarte, line 1: Given that we now have the full server, I'm not sure how beneficial it is to keep this file around

storage-provider/server/src/rpc.rs
```diff
@@ -314,7 +314,7 @@ impl StorageProviderRpcServer for RpcServerState {
             .send(PipelineMessage::AddPiece(AddPieceMessage {
                 deal: deal_proposal,
                 published_deal_id: deal_id,
-                piece_path,
+                piece_path: piece_path.clone(),
```
jmg-duarte: This change looks like a forgotten test

storage-provider/server/src/config.rs
```diff
@@ -63,6 +69,11 @@ pub struct ConfigurationArgs {
     #[arg(long, default_value_t = default_node_address())]
     pub(crate) node_url: Url,
 
+    /// Storage provider listen address.
```
jmg-duarte:
```suggestion
    /// Storage provider retrieval service listen address.
```

jmg-duarte: Or
```suggestion
    /// Storage provider retrieval listen address.
```

storage-provider/server/src/retrieval/mod.rs
jmg-duarte, line 1: Question for discussion, could we make retrieval a separate crate to increase compilation parallelization?

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,206 @@
...
+        // If CID is an identity
+        if let Some(data) = cid.get_identity_data() {
+            return Ok(Some(data.to_owned()));
+        }
+
+        let cid = to_blockstore_cid(cid)?;
+
+        // Pieces containing the cid.
+        let Ok(pieces) = self.indexer.pieces_containing_multihash(*cid.hash()) else {
+            return Ok(None);
```
jmg-duarte, line 51: Add logs please

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,206 @@
...
+        };
+
+        // We take the first piece that contains the multihash
+        let Some(piece_cid) = pieces.first() else {
+            return Ok(None);
```
jmg-duarte, line 57: Same as above

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,206 @@
...
+        Ok(Some(buffer))
+    }
+
+    async fn put_keyed<const S: usize>(
+        &self,
+        _cid: &cid::CidGeneric<S>,
+        _data: &[u8],
+    ) -> blockstore::Result<()> {
+        Err(Error::FatalDatabaseError(
+            "put operation not supported".to_string(),
```
jmg-duarte:
```suggestion
            "Blockstore is read-only".to_string(),
```

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,206 @@
...
+        ))
+    }
+
+    async fn remove<const S: usize>(&self, _cid: &cid::CidGeneric<S>) -> blockstore::Result<()> {
+        Err(Error::FatalDatabaseError(
+            "remove operation not supported".to_string(),
```
jmg-duarte:
```suggestion
            "Blockstore is read-only".to_string(),
```

storage-provider/server/src/retrieval/blockstore.rs
```diff
@@ -0,0 +1,206 @@
...
+/// Convert CID with the generic Multihash size to the CID with the specific
+/// Multihash size that the underlying blockstore expects.
+fn to_blockstore_cid<const S: usize>(cid: &CidGeneric<S>) -> Result<Cid, Error> {
+    let digest_size = cid.hash().size() as usize;
+    let hash = cid
+        .hash()
+        .resize::<64>()
+        .map_err(|_| Error::CidError(CidError::InvalidMultihashLength(digest_size)))?;
+
+    Ok(Cid::new(cid.version(), cid.codec(), hash).expect("we know cid is correct here"))
```
jmg-duarte:
```suggestion
    Ok(Cid::new(cid.version(), cid.codec(), hash).expect("new CID is derived from a valid existing CID, this should never fail"))
```

jmg-duarte: Though we could also just return the error since we know it won't happen (?)

jmg-duarte (commented):
Looks good! 

I think there are (IMO) some small refactors to be made that make the whole PR smaller and more understandable

storage-provider/server/src/indexer/local_index_directory/rdb.rs
cernicc: Correct. Good catch

storage-provider/server/src/indexer/mod.rs
cernicc: Hmm. I don't remember 😄 I think it was arbitrary.

storage-provider/server/src/indexer/mod.rs
cernicc: This is a good point for long living tasks. In this case the index_piece is spawned for each indexing separately.

storage-retrieval/lib/examples/simple_server.rs
cernicc: I've used it when testing and developing the retrieval. I expect to need it in the future again someday

storage-provider/server/src/rpc.rs
cernicc: Good catch. A leftover from the other impl when the indexing was started immediately

storage-provider/server/src/retrieval/mod.rs
cernicc: That was the idea at first. This could probably live in the `storage-retrieval`. But the ProviderBlockStore heavely depends on the lid. The lid is used by the provider-server and retrieval. We could also move the lid out again to the separate crate or to the `storage-retrieval`. I am not sure yet if the lid wil only be used for the data retrieval.  just didn't want to complicate things until we get a clearer picture.

storage-provider/server/src/retrieval/blockstore.rs
cernicc: I've left an explicit indication of not failing. The error returned from the `Cid::new` is different. So we would still have to transform an error to the one returned by the function.

storage-provider/server/tests/fixtures/spaceglenda_wrapped_v2.car
cernicc: Converted to symlink

storage-provider/server/src/retrieval/mod.rs
jmg-duarte: Yeah, my point isn't really about changing now either, but I'll take compilation wins where possible

storage-provider/server/src/indexer/mod.rs
cernicc: I changed it to `I`

storage-provider/server/src/indexer/mod.rs
```diff
@@ -72,15 +71,15 @@ where
 }
 
 #[instrument(skip_all, fields(piece_cid = %commitment.cid()))]
-async fn index_piece<D, P>(db: Arc<D>, commitment: Commitment<CommP>, piece_path: P)
+async fn index_piece<I, P>(db: Arc<I>, commitment: Commitment<CommP>, piece_path: P)
 where
-    D: Service + Send + Sync + 'static,
-    P: AsRef<Path>,
+    I: Service + Send + Sync + 'static,
+    P: AsRef<Path> + Debug,
```
jmg-duarte: Instead of this you can do `piece_path = ?piece_path.as_ref()` on the log line

storage-provider/server/src/indexer/mod.rs
```diff
@@ -157,7 +166,7 @@ pub mod tests {
         piece_path: P,
     ) where
         D: Service + Send + Sync + 'static,
-        P: AsRef<Path>,
+        P: AsRef<Path> + Debug,
```
jmg-duarte: Same as above

jmg-duarte (changes requested):
Final touchups

storage-provider/server/src/indexer/mod.rs
```diff
@@ -0,0 +1,239 @@
...
+    .await
+    {
+        Ok(Ok(_)) => {
+            info!("indexing completed");
+        }
+        Ok(Err(err)) => {
+            error!(?err, "piece indexing failed with an error");
+        }
+        Err(err) => {
+            error!(?err, "failed to join task");
```
aidan46, line 101: Why is this not a hard error? I get that indexing would fail if you attempt to get from the DB but shouldn't the error on insertion be handled too?

aidan46 (dismissed):

storage-provider/server/src/indexer/mod.rs
cernicc: If you are asking why the `index_piece` is not returning a result. If the `index_piece` were to return a result. Then that result would not be handled because we are not joining a task in `on_command`. That way we at least get a log that the error happened.

jmg-duarte (dismissed):

jmg-duarte (approved):

aidan46 (approved):

PR #689, opened 2025-01-21
feat: replace Currency logic with Market in the SP pallet

jmg-duarte:
### Description

Stops using the `Currency` trait for handling balances. Centralizing everything in the Market pallet.

If you find uses of `Balances` inside the SP pallet flag them down!

### DIFF
--- a/pallets/market/src/lib.rs
+++ b/pallets/market/src/lib.rs
@@ -23,7 +23,7 @@ pub mod pallet {
     use cid::Cid;
     use codec::{Decode, Encode};
     use frame_support::{
-        dispatch::{DispatchClass, DispatchResult},
+        dispatch::DispatchResult,
         ensure,
         pallet_prelude::*,
         sp_runtime::{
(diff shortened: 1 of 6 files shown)


pallets/market/src/lib.rs
```diff
@@ -305,6 +305,7 @@ pub mod pallet {
     /// Invariant must be held at all times:
     /// `account(MarketPallet).balance == all_accounts.map(|balance| balance[account]].locked + balance[account].free).sum()`
     #[pallet::storage]
+    #[pallet::getter(fn balance_table)] // NOTE(@jmg-duarte,21/1/25): maybe this could be gated?
```
aidan46: Do we need this getter? AFIAK this is just an alias for `BalanceTable::get()` and it does not seem to be used.

cernicc: It's used bellow (next change)

EDIT: My statement above is wrong :D

jmg-duarte: We don't need it anymore, I exposed the state in other, more ergonomic ways. Good catch

cernicc (approved):
Nice

th7nder (approved):

jmg-duarte:
No one noticed the PR name was wrong 💀 my bad but too late now

PR #690, opened 2025-01-21
feat(mater): add unixfs wrapping support

### LINKED ISSUE #675: feat: Mater `convert`, wrap contents in unixfs dag

cargopete:
# PR: Add UnixFS Wrapping Support

## Description

This PR adds the ability for users to decide whether to wrap file contents in raw blocks or UnixFS DAG nodes when converting to a CARv2. Specifically:

1. **`WrapMode` enum** (`Raw` or `UnixFS`) is introduced.
2. **CLI updated**: A `--wrap` option allows the user to choose how data is packaged in the CAR.
3. **Balanced‐tree builder**: A new `stream_balanced_tree_unixfs` function creates proper UnixFS leaves and parent nodes, ensuring the DAG is navigable by retrieval clients.
4. **Raw flow preserved**: If `--wrap=raw` is chosen, it uses `stream_balanced_tree` with raw leaves.

All relevant tests now pass, including a new check verifying that the UnixFS mode indeed produces the expected number of leaf blocks (matching the total chunk count) and parent nodes.

## Important points for reviewers

- **UnixFS DAG creation**: Handled by `stream_balanced_tree_unixfs`, which encodes each chunk as a DAG‐PB `File` node.
- **Raw mode**: Leaves remain raw blocks, handled by the original `stream_balanced_tree`.
- **CLI**: A new `ConvertConfig` uses `WrapMode` to drive the logic in `create_filestore`.
- **Tests**: The test `test_filestore_unixfs_dag_structure` confirms a correct UnixFS DAG with equal numbers of chunks and leaf blocks.

## Checklist

- [x] **Important points for reviewers?**  
  The code merges neatly with existing balanced‐tree logic. Tests confirm that both raw and UnixFS modes work properly.
- [x] **Description of the change?**  
  Yes, see above.
- [x] **Follow‐ups?**  
  None; this PR completes the UnixFS wrapping feature.
- [x] **Has it been tested?**  
  Yes, new tests pass locally (`test_filestore_unixfs_dag_structure`).
- [x] **Any alternative approaches considered?**  
  A manual parent node builder was considered, but leveraging the existing balanced‐tree approach is cleaner.
- [x] **Documented new APIs?**  
  The `WrapMode` enum is documented and the CLI’s `--wrap` flag is explained in the help text.


mater/lib/src/stores/mod.rs
```diff
@@ -16,32 +15,63 @@ pub(crate) const DEFAULT_TREE_WIDTH: usize = 174;
...
+    /// Wrap content in UnixFS format which includes file metadata and DAG structure.
+    /// This is compatible with IPFS and provides richer metadata about the content.
+    UnixFS,
+}
+
+impl Default for WrapMode {
+    fn default() -> Self {
+        Self::Raw
+    }
+}
```
aidan46: Defaults for unit-type enums can be derived too:

```suggestion
/// Determines how content is encoded in the CAR file.
#[derive(clap::ValueEnum, Clone, Debug, Default)]
pub enum WrapMode {
    /// Store content directly in the CAR file without additional metadata.
    /// This is the most space-efficient option but provides minimal metadata.
    #[default]
    Raw,
    /// Wrap content in UnixFS format which includes file metadata and DAG structure.
    /// This is compatible with IPFS and provides richer metadata about the content.
    UnixFS,
}
```

mater/lib/src/v2/reader.rs
```diff
@@ -62,21 +67,48 @@ where
...
             let position = self.get_inner_mut().stream_position().await?;
-            let data_end = header.data_offset + header.data_size;
-            // Add the `written != 0` clause for files that are less than a single block.
             if position >= data_end && written != 0 {
                 break;
             }
-            written += output_file.write(&contents).await?;
+
+            if let Ok((cid, contents)) = self.read_block().await {
+                if contents.len() == 0 {
```
aidan46, line 88: I remember when implementing this that this break condition cause the output file to miss some data. Did you check that the output file contains all the data?

mater/lib/src/stores/mod.rs
jmg-duarte: There's no other wrapping. Having an enum here is overengineering

mater/lib/Cargo.toml
```diff
@@ -29,6 +29,7 @@ thiserror.workspace = true
 tokio = { workspace = true, features = ["fs", "macros", "rt-multi-thread"] }
 tokio-stream.workspace = true
 tokio-util = { workspace = true, features = ["io"] }
+clap = { workspace = true, features = ["derive"] }
```
jmg-duarte: This must be removed. The mater crate has nothing to do with the CLI

mater/cli/src/main.rs
```diff
@@ -29,6 +31,11 @@ enum MaterCli {
         /// If enabled, the output will overwrite any existing files.
         #[arg(long, action)]
         overwrite: bool,
+
+        /// Determines how content should be wrapped in the CAR file.
+        /// 'raw' stores content directly, 'unixfs' wraps it in UnixFS format (default: raw)
+        #[arg(long, value_enum, default_value = "raw")]
+        wrap: WrapMode,
```
jmg-duarte: It's either wrapped or not. This is a bool

maat/tests/real_world.rs
jmg-duarte, line 1: How is this file related to mater?

pallets/proofs/src/porep/mod.rs
jmg-duarte, line 1: How is this file related to mater?

mater/lib/src/stores/mod.rs
```diff
@@ -16,32 +15,63 @@ pub(crate) const DEFAULT_TREE_WIDTH: usize = 174;
...
     }
 }
 
 impl Default for Config {
     fn default() -> Self {
         Self::Balanced {
-            chunk_size: DEFAULT_BLOCK_SIZE,
-            tree_width: DEFAULT_TREE_WIDTH,
+            chunk_size: 256 * 1024,  // 256KiB
+            tree_width: 174,         // Default from ipfs
```
jmg-duarte: 1 - const
2 - why are you changing the existing value?

mater/lib/src/stores/filestore.rs
```diff
@@ -1,32 +1,41 @@
...
+use futures::stream::Peekable;
+use crate::unixfs::stream_balanced_tree_unixfs;
+use super::{Config, WrapMode};
 use crate::{
     multicodec::SHA_256_CODE, unixfs::stream_balanced_tree, CarV1Header, CarV2Header, CarV2Writer,
     Error, Index, IndexEntry, MultihashIndexSorted, SingleWidthIndex,
 };
+use std::collections::HashMap;
+
+type SendableStream = Pin<Box<dyn Stream<Item = Result<(Cid, Vec<u8>), Error>> + Send>>;
```
jmg-duarte: I don't understand why the `dyn` is used over a type parameter

mater/lib/src/stores/filestore.rs
```diff
@@ -1,32 +1,41 @@
...
     mut output: Out,
     chunk_size: usize,
     tree_width: usize,
+    wrap_mode: WrapMode,
 ) -> Result<Cid, Error>
 where
-    Src: AsyncRead + Unpin,
-    Out: AsyncWrite + AsyncSeek + Unpin,
+    Src: AsyncRead + Unpin + Send + 'static,
+    Out: AsyncWrite + AsyncSeek + Unpin + Send + 'static,
```
jmg-duarte, line 29: I'm not the most versed in async Rust, could you elaborate why you needed these bounds?

mater/lib/src/stores/filestore.rs
```diff
@@ -45,28 +54,60 @@ where
             // this means there is no right way of knowing when the reader is fully exhausted!
             // If we need to support a case like that, we just need to track how many times
             // the reader returned 0 and break at a certain point
-            if source.read_buf(&mut buf).await? == 0 {
-                // EOF but there's still content to yield -> yield it
+            let read_bytes = source.read_buf(&mut buf).await?;
+            println!("Read {} bytes, buffer now has {} bytes", read_bytes, buf.len());
```
jmg-duarte: `println!` inside a library

mater/lib/src/stores/filestore.rs
```diff
@@ -45,28 +54,60 @@ where
...
+
+            if read_bytes == 0 {
                 if buf.len() > 0 {
+                    total_chunks += 1;
+                    println!("Final chunk {}: {} bytes", total_chunks, buf.len());
```
jmg-duarte: `println!` inside a library

mater/lib/src/stores/filestore.rs
```diff
@@ -45,28 +54,60 @@ where
...
                 }
-                break
-            } else if buf.len() >= chunk_size {
+                break;
+            }
+
+            // While we have enough for a full chunk
+            while buf.len() >= chunk_size {
+                total_chunks += 1;
+                println!("Full chunk {}: {} bytes", total_chunks, chunk_size);
```
jmg-duarte: `println!` inside a library

mater/lib/src/stores/filestore.rs
```diff
@@ -45,28 +54,60 @@ where
...
+            }
+            // If we reach EOF but still have a partial chunk, yield it
+            if read_bytes == 0 && buf.len() > 0 {
+                total_chunks += 1;
+                println!("Partial chunk {}: {} bytes", total_chunks, buf.len());
+                let chunk = buf.split();
+                yield chunk.freeze();
+            }
+        }
+        println!("Total chunks yielded: {}", total_chunks);
```
jmg-duarte: `println!` inside a library

mater/lib/src/stores/filestore.rs
```diff
@@ -45,28 +54,60 @@ where
...
                 // thats why we check if the length if bigger than the chunk_size and if so
                 // we split the buffer to the chunk_size, then freeze and return
                 let chunk = buf.split_to(chunk_size);
                 yield chunk.freeze();
-            } // otherwise, the buffer is not full, so we don't do a thing
+            }
+            // If we reach EOF but still have a partial chunk, yield it
+            if read_bytes == 0 && buf.len() > 0 {
+                total_chunks += 1;
+                println!("Partial chunk {}: {} bytes", total_chunks, buf.len());
```
jmg-duarte: `println!` inside a library

maat/tests/real_world.rs
```diff
@@ -1,4 +1,4 @@
-use std::collections::BTreeSet;
+use std::{collections::BTreeSet, time::Duration};
```
cernicc: Konrad is currently working on fixing the maat. Do you think we should remove the changes so that he won't have conflicts when merging? The thing is that there are more changes needed to fix the tests. There are not only compile time problems.

mater/cli/src/convert.rs
```diff
@@ -3,22 +3,30 @@ use std::path::PathBuf;
...
     }?;
-    let cid = create_filestore(source_file, output_file, Config::default()).await?;
 
+    let mater_config = match config.wrap_mode {
+        WrapMode::Raw => Config::default(),
+        WrapMode::UnixFS => Config::Balanced {
+            chunk_size: 64 * 1024,
+            tree_width: 2,
+            wrap_mode: WrapMode::UnixFS,
+        },
```
jmg-duarte: Just create a UnixFS default config + these are, once more, magic values

storage-provider/server/src/storage.rs
```diff
@@ -194,20 +198,27 @@ async fn upload(
...
+        let field = multipart
             .next_field()
-            .map_err(|err| (StatusCode::BAD_REQUEST, err.to_string()))
-            .await?
-        else {
-            return Err((StatusCode::BAD_REQUEST, "empty request".to_string()));
-        };
+            .await
+            .map_err(|err| (StatusCode::BAD_REQUEST, err.to_string()))?
+            .ok_or((StatusCode::BAD_REQUEST, "empty request".to_string()))?;
```
jmg-duarte: You're forcing a string allocation without need
```suggestion
            .ok_or_else(|| (StatusCode::BAD_REQUEST, "empty request".to_string()))?;
```

mater/lib/src/stores/mod.rs
cernicc: True. Also lets have a wrapped mode by default. If the user wants the raw they should provide a flag.

storage-provider/server/src/storage.rs
```diff
@@ -161,17 +165,17 @@ fn configure_router(state: Arc<StorageServerState>) -> Router {
 /// ```
 #[tracing::instrument(skip_all, fields(cid))]
 async fn upload(
-    ref s @ State(ref state): State<Arc<StorageServerState>>,
+    State(state): State<Arc<StorageServerState>>,
```
jmg-duarte, line 172: This change seems arbitrary

mater/lib/src/stores/mod.rs
```diff
@@ -16,32 +15,63 @@ pub(crate) const DEFAULT_TREE_WIDTH: usize = 174;
...
         }
     }
 }
 
 impl Default for Config {
     fn default() -> Self {
         Self::Balanced {
-            chunk_size: DEFAULT_BLOCK_SIZE,
-            tree_width: DEFAULT_TREE_WIDTH,
+            chunk_size: 256 * 1024,  // 256KiB
```
cernicc: What was the reason of removing the DEFAULT_BLOCK_SIZE and DEFAULT_TREE_WIDTH? I just want to see if there is something to learn here :D

mater/lib/src/stores/mod.rs
```diff
@@ -16,32 +15,63 @@ pub(crate) const DEFAULT_TREE_WIDTH: usize = 174;
...
+            tree_width: 174,         // Default from ipfs
+            wrap_mode: WrapMode::Raw // Default to raw like go-car
```
cernicc: go-car creates a wrapped car by default

mater/cli/src/convert.rs
```diff
@@ -3,22 +3,30 @@ use std::path::PathBuf;
...
+    let output_file = if config.overwrite {
         File::create(output_path).await
     } else {
         File::create_new(output_path).await
     }?;
-    let cid = create_filestore(source_file, output_file, Config::default()).await?;
 
+    let mater_config = match config.wrap_mode {
+        WrapMode::Raw => Config::default(),
+        WrapMode::UnixFS => Config::Balanced {
```
cernicc: How were the `chunk_size` and `tree_width` here determined?

mater/cli/src/convert.rs
```diff
@@ -49,18 +55,42 @@ mod tests {
...
+        // Configure for UnixFS mode
+        let config = ConvertConfig {
+            wrap_mode: WrapMode::UnixFS,
+            overwrite: false,
+        };
+
+        let result = convert_file_to_car(&input_path, &output_path, config).await;
+        assert!(result.is_ok());
+
+        temp_dir.close()?;
```
cernicc: No need to manually close the temp_dir. Lets remove this in other tests too.

storage-provider/server/src/storage.rs
```diff
@@ -194,20 +198,27 @@ async fn upload(
...
+            .map_err(|err| (StatusCode::BAD_REQUEST, err.to_string()))?
+            .ok_or((StatusCode::BAD_REQUEST, "empty request".to_string()))?;
 
-        let field_reader = StreamReader::new(field.map_err(std::io::Error::other));
-        stream_contents_to_car(state.car_piece_storage_dir.clone().as_ref(), field_reader)
+        // Read the entire field into memory
+        let bytes = field
+            .bytes()
+            .await
+            .map_err(|err| (StatusCode::BAD_REQUEST, err.to_string()))?;
```
jmg-duarte: Won't this consume the entire byte stream?

storage-provider/server/src/storage.rs
```diff
@@ -391,7 +401,14 @@ where
     // Stream the body from source to the temp file.
     let file = File::create(&temp_file_path).await?;
     let writer = BufWriter::new(file);
-    let cid = mater::create_filestore(source, writer, mater::Config::default()).await?;
+
+    let config = mater::Config::Balanced {
+        chunk_size: 256 * 1024,
+        tree_width: 174,
+        wrap_mode: WrapMode::UnixFS,
+    };
```
jmg-duarte: More magic

mater/lib/src/stores/filestore.rs
cernicc: No need for the type alias. There is https://docs.rs/futures/latest/futures/stream/type.BoxStream.html

mater/lib/src/unixfs/mod.rs
```diff
@@ -2,6 +2,7 @@
 //! <https://github.com/n0-computer/beetle/blob/3e137cb2bc18e1d458c3f72d5e817b03d9537d5d/iroh-unixfs/src/balanced_tree.rs>.
 
 mod unixfs_pb;
+pub use unixfs_pb::{Data, mod_Data};
```
jmg-duarte: This won't pass CI, you must run `cargo +nightly fmt`

mater/lib/src/unixfs/mod.rs
```diff
@@ -331,6 +331,154 @@ where
     }
 }
 
+impl TreeNode {
+    fn encode_unixfs_leaf(self) -> Result<((Cid, Bytes), LinkInfo), Error> {
```
jmg-duarte: Split this further into two specialized methods, one for the leaf and another for the stem

mater/lib/src/unixfs/mod.rs
```diff
@@ -331,6 +331,154 @@ where
...
+                    raw_data_length: data_buf.len() as u64,
+                    encoded_data_length: encoded.len() as u64,
+                };
+                Ok(((cid, encoded.into()), info))
+            }
+
+            TreeNode::Stem(children) => {
+                let total_file_size: u64 = children.iter().map(|c| c.1.raw_data_length).sum();
+                let total_encoded: u64 = children.iter().map(|c| c.1.encoded_data_length).sum();
+                let blocksizes: Vec<_> = children.iter().map(|c| c.1.raw_data_length).collect();
```
jmg-duarte: You're performing 3 loops when you could do just 1

mater/lib/src/unixfs/mod.rs
```diff
@@ -331,6 +331,154 @@ where
...
+                    let mut w = quick_protobuf::Writer::new(&mut data_buf);
+                    unixfs_data.write_message(&mut w)?;
+                }
+                let pb_node = PbNode {
+                    links: vec![], // leaf => no links
+                    data: Some(data_buf.clone().into()),
+                };
+                let encoded = DagPbCodec::encode_to_vec(&pb_node)?;
+                let mh = generate_multihash::<Sha256,_>(&encoded);
+                let cid = Cid::new_v1(DAG_PB_CODE, mh);
```
jmg-duarte: Why v1?

mater/lib/src/stores/filestore.rs
```diff
@@ -45,28 +54,60 @@ where
...
+            })
+        )),
+        WrapMode::UnixFS => {
+                futures::stream::StreamExt::peekable(Box::pin(
+                    stream_balanced_tree_unixfs(chunker, tree_width)
+                        .map(|res| res)
+                ))
+
         }
     };
```
cernicc:
```suggestion
    let nodes: Peekable<SendableStream> = match wrap_mode {
            stream_balanced_tree(chunker, tree_width).map(|result| {
                result.map(|(cid, bytes)| (Cid::from(cid), bytes.to_vec()))
            })
            .boxed()
            .peekable(),
        WrapMode::UnixFS => {
            stream_balanced_tree_unixfs(chunker, tree_width)
            .map(|res| res)
            .boxed()
            .peekable()
        }
    };
```

Didn't try. Should work.

mater/lib/src/stores/filestore.rs
cernicc: Do not use `println`. Use `tracing::trace!` instead.

mater/lib/src/stores/filestore.rs
```diff
@@ -45,28 +54,60 @@ where
...
                 // The buffer may have a larger capacity than chunk_size due to reserve
                 // this also means that our read may have read more bytes than we expected,
                 // thats why we check if the length if bigger than the chunk_size and if so
                 // we split the buffer to the chunk_size, then freeze and return
                 let chunk = buf.split_to(chunk_size);
                 yield chunk.freeze();
-            } // otherwise, the buffer is not full, so we don't do a thing
+            }
+            // If we reach EOF but still have a partial chunk, yield it
+            if read_bytes == 0 && buf.len() > 0 {
```
cernicc: Was this a bug?

mater/lib/src/stores/filestore.rs
```diff
@@ -78,12 +119,20 @@ where
...
     let mut entries = vec![];
+
     while let Some(node) = nodes.next().await {
         let (node_cid, node_bytes) = node?;
-        let digest = node_cid.hash().digest().to_owned();
-        let entry = IndexEntry::new(digest, (position - car_v1_start) as u64);
-        entries.push(entry);
-        position += writer.write_block(&node_cid, &node_bytes).await?;
+
+        if let Some(existing_position) = written_blocks.get(&node_cid) {
```
cernicc: This change was needed because of deduplication? Was that a bug?

mater/lib/Cargo.toml
cernicc: I think we should hide the clap behind a feature flag. This was one of the reasons why we split the cli and lib.

mater/lib/src/v2/reader.rs
```diff
@@ -21,7 +26,7 @@ impl<R> Reader<R> {
 
 impl<R> Reader<R>
 where
-    R: AsyncRead + Unpin,
+    R: AsyncRead + Unpin + AsyncSeek,
```
cernicc, line 29: Why is Seek needed?

storage-provider/server/src/storage.rs
```diff
@@ -117,7 +118,10 @@ fn configure_router(state: Arc<StorageServerState>) -> Router {
     #[cfg(not(feature = "delia"))]
     fn config_non_delia(state: Arc<StorageServerState>) -> Router {
         Router::new()
-            .route("/upload/:cid", put(upload))
+            .route(
+                "/upload/:cid",
+                put(upload as fn(State<Arc<StorageServerState>>, Path<String>, Request<Body>) -> _)
```
cernicc: Why is this change needed?

cernicc (commented):
Good job on figuring this out 💪 I've added some comments and questions. The tests are currently not compiling. 

Heads up. If you want to rerun the ci you have to retag the pr with the `ready for review`. The ci doesn't run automatically on each commit.

mater/lib/src/unixfs/mod.rs
```diff
@@ -331,6 +331,154 @@ where
...
+                    unixfs_data.write_message(&mut w)?;
+                }
+
+                let pb_links = children.into_iter().map(|(child_cid, link_info)| {
+                    PbLink {
+                        cid: child_cid,
+                        name: Some("".to_string()),
+                        size: Some(link_info.encoded_data_length),
+                    }
+                }).collect::<Vec<_>>();
```
jmg-duarte: You can also add this to the previous loop issue

mater/lib/src/stores/mod.rs
```diff
@@ -1,10 +1,9 @@
 mod blockstore;
 mod file;
-mod filestore;
+pub mod filestore;
```
jmg-duarte: Why are you exposing the whole module?

mater/lib/src/stores/mod.rs
```diff
@@ -16,32 +15,63 @@ pub(crate) const DEFAULT_TREE_WIDTH: usize = 174;
...
+
+/// Configuration options for file conversion.
+pub struct ConvertConfig {
+    /// How the content should be wrapped in the CAR file.
+    /// See [`WrapMode`] for details.
+    pub wrap_mode: WrapMode,
+    /// Whether to overwrite existing files at the output path.
+    /// If false, will error if the output file already exists.
+    pub overwrite: bool,
+}
```
jmg-duarte: This structure adds more mental overhead than it solves issues. They're two parameters, it's ok to use them explicitly

jmg-duarte (changes requested):
Good work. The PR description would be better if it explained how the new code achieves the functionality since there's a lack of comments on the code for that specific purpose

mater/lib/src/stores/mod.rs
cargopete: switched to bool flag per other feedback, I've simplified to a boolean raw_mode flag, defaulting to false (wrapped mode)

mater/lib/src/v2/reader.rs
cargopete: Fixed with proper DAG traversal. Tested with byte-by-byte comparisons in test_filestore_roundtrip.

mater/lib/src/stores/mod.rs
cargopete: restored

mater/cli/src/convert.rs
cargopete: Using IPFS defaults: 256KiB chunks, 174 tree width. Added as constants with documentation

mater/cli/src/convert.rs
cargopete: Removed

mater/lib/src/stores/filestore.rs
cargopete: Replaced custom type alias with futures::stream::BoxStream

mater/lib/src/stores/filestore.rs
cargopete: Replaced println! with tracing::trace!

mater/lib/src/stores/filestore.rs
cargopete: I think so, fixed logic to properly handle partial chunks at EOF

mater/lib/src/stores/filestore.rs
cargopete: yeah - previous implementation wasn't tracking duplicate blocks correctly (I think)

mater/lib/src/v2/reader.rs
cargopete: Required for DAG traversal in extract_content to follow block links, me thinks

storage-provider/server/src/storage.rs
cargopete: Needed to satisfy type inference for UnixFS async bounds

mater/lib/src/unixfs/mod.rs
cargopete: I think unixFS DAG nodes require CIDv1 to support arbitrary codecs (DAG-PB in this case) ;>

mater/lib/src/stores/mod.rs
cargopete: reverted to only expose what we need

mater/lib/src/unixfs/mod.rs
pete-eiger: good call, will do in next PR (the one that replaced this one)

mater/lib/src/stores/mod.rs
cargopete: I like that the struct provides a single place to manage all configuration related to file conversion behavior. As we potentially add more options, having them grouped in a config struct:

- Makes the API more stable (no need to change function signatures)
- Makes configurations reusable across different parts of the codebase
- Provides a clear boundary between config and business logic
- Makes it easier to serialize/deserialize configurations

it is an overhead now with so little params, but it follows the principle of designing for future extensibility ;>

maat/tests/real_world.rs
cargopete: reverted

pallets/proofs/src/porep/mod.rs
cargopete: reverted

mater/lib/src/stores/mod.rs
cargopete: reverted, using constants DEFAULT_CHUNK_SIZE and DEFAULT_TREE_WIDTH

mater/lib/src/stores/filestore.rs
cargopete: reverted

mater/lib/src/stores/filestore.rs
cargopete: added code comments

mater/lib/src/stores/filestore.rs
cargopete: whoops, used it for debugging

storage-provider/server/src/storage.rs
cargopete: simpler, more idiomatic pattern for handling Axum state, the old way was over-complicated imo

PR #686, opened 2025-01-21
docs: Update docs to include p2p Peer ID and keys

aidan46:
### Description

Add docs to Polka Storage Provider server on how to generate an ed25519 keypair.
Add docs to Polka Storage Provider client on how to convert an ed25519 public key to Peer ID.

PR #696, opened 2025-01-23
feat: Add p2p resolver example

### LINKED ISSUE #687: feat: Add p2p discovery example
Add an example that can discover peers in the p2p network. This example should take in a Peer ID and return the multiaddress associated with that Peer ID if it is available.

aidan46:
### Description

Add an example for resolving a Peer ID to a multiaddr. I have created a new binary for this in the root, I am aware that this is not the place to put it but I am not sure where I should place it.

This PR should be merged after #686


examples/peer-resolver/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+) -> Result<PeerInfo> {
+    // Dial in to the rendezvous point.
+    swarm.dial(rendezvous_point_address)?;
+
+    loop {
+        match swarm.select_next_some().await {
+            SwarmEvent::ConnectionEstablished { peer_id, .. } => {
+                if peer_id == rendezvous_point {
+                    tracing::info!("Connection established with rendezvous point {}", peer_id);
+                    tracing::info!("Connected to rendezvous point, discovering nodes...");
```
jmg-duarte: Pick one, essentially, they both say the same thing

examples/peer-resolver/src/main.rs
jmg-duarte, line 1: I think the file needs a mod-level description explaining what it does/how to use it.

Eventually this is to be removed and just implemented in the client and when that time arrives, having an actual description of what is happening will be helpful

jmg-duarte (commented):
> I have created a new binary for this in the root, I am aware that this is not the place to put it but I am not sure where I should place it.

Maybe `storage-provider/client/examples`

th7nder (dismissed):

jmg-duarte (dismissed):

th7nder (approved):

jmg-duarte (approved):

PR #697, opened 2025-01-23
fix: Unlock pre commit desposit after proving

aidan46:
### Description

Unlock the pre-commit deposit after the sector has been proven. Before this, the pre-commit deposit funds were never returned to the storage provider. The tests have been updated to reflect this change.

References:
[FC 1](https://github.com/filecoin-project/builtin-actors/blob/d5ef63537774b9b144ea5e332077540f9d3a1842/actors/miner/src/lib.rs#L5567)
[FC 2](https://github.com/filecoin-project/builtin-actors/blob/d5ef63537774b9b144ea5e332077540f9d3a1842/actors/miner/src/lib.rs#L5613-L5616)

### DIFF
--- a/pallets/storage-provider/src/lib.rs
+++ b/pallets/storage-provider/src/lib.rs
@@ -677,7 +677,6 @@ pub mod pallet {
                 .pre_commit_deposits
                 .checked_sub(&pre_commit_deposit_to_unlock)
             {
-                log::info!("Unlocking {pre_commit_deposit_to_unlock:?} from pre-commit deposit");
                 sp.pre_commit_deposits = pre_commit_deposits
             } else {
                 log::error!(target: LOG_TARGET, "catastrophe, failed to subtract from pre_commit_deposits {:?} - {:?} < 0", sp.pre_commit_deposits, pre_commit_deposit_to_unlock);
--- a/pallets/storage-provider/src/lib.rs
+++ b/pallets/storage-provider/src/lib.rs
@@ -374,6 +374,8 @@ pub mod pallet {
         TooManyProofs,
         /// AuthorVRF lookup failed.
         MissingAuthorVRF,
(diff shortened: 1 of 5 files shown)


jmg-duarte (dismissed):

cernicc (dismissed):

pallets/storage-provider/src/lib.rs
```diff
@@ -667,6 +672,18 @@ pub mod pallet {
...
+                .pre_commit_deposits
+                .checked_sub(&pre_commit_deposit_to_unlock)
+            {
+                sp.pre_commit_deposits = pre_commit_deposits
+            } else {
+                log::error!(target: LOG_TARGET, "catastrophe, failed to subtract from pre_commit_deposits {:?} - {:?} < 0", sp.pre_commit_deposits, pre_commit_deposit_to_unlock);
+                return Err(Error::<T>::FailedToReturnPreCommitDeposit.into());
+            };
+            // Unlock pre commit deposit funds.
+            T::Market::unlock_pre_commit_funds(&owner, pre_commit_deposit_to_unlock)?;
```
jmg-duarte, line 686: This makes me wonder if we should keep two balance tables in the market pallet

jmg-duarte (approved):
🚀

pallets/storage-provider/src/lib.rs
aidan46: Do you mean moving the pre commit deposits into the market pallet? That makes sense as the market pallet deals with all the balances

th7nder (approved):

pallets/storage-provider/src/lib.rs
jmg-duarte: No. I mean keeping two balance tables

PR #705, opened 2025-01-24
feat(pallet-market): implement benchmarks

### LINKED ISSUE #77: Market Pallet: take it out of the dev mode
Market Pallet is in the [dev_mode](https://paritytech.github.io/polkadot-sdk/master/frame_support/attr.pallet.html#dev-mode-palletdev_mode).

Make sure to:
- set the correct weights (benchmarking)
- set the correct hash functions 
- call indexes 
- anything else that's mentioned in the docs of dev mode

Can be separated into multiple tasks.

jmg-duarte:
### Description

* Implements benchmarks for the `pallet-market`
* Applies the weights to the pallet

Closes #77


pallets/market/src/benchmarking.rs
```diff
@@ -0,0 +1,409 @@
...
+    client: T::AccountId,
+    client_pair: MultiSigner,
+) -> (
+    Vec<ClientDealProposal<T::AccountId, BalanceOf<T>, BlockNumberFor<T>, MultiSignature>>,
+    u32,
+)
+where
+    T: crate::Config,
+{
+    let mut small_rng = SmallRng::seed_from_u64(0x101097);
```
aidan46: Magic seed number? Is this so it creates the same CID every time?

primitives/src/sector/pre_commit.rs
```diff
@@ -119,9 +121,24 @@ pub mod builder {
...
+
+        pub fn raw_unsealed_cid(
+            mut self,
+            unsealed_cid: BoundedVec<u8, ConstU32<CID_SIZE_IN_BYTES>>,
+        ) -> Self {
+            self.unsealed_cid = unsealed_cid;
+            self
+        }
+
+        pub fn raw_sealed_cid(
```
aidan46: This function doesn't seem to be used anywhere

pallets/market/src/benchmarking.rs
jmg-duarte: It's just because we can't use other sources of randomness due to WASM. This was the way I found of getting some randomness for the tests.

.github/workflows/ci.yaml
```diff
@@ -5,6 +5,7 @@ on:
     types: [labeled]
   push:
     branches: [develop, main]
+  workflow_dispatch: # manual launch
```
th7nder, line 8: Does it conflict with automatic on `labeled` or on `develop`?

.github/workflows/ci.yaml
```diff
@@ -121,11 +123,27 @@ jobs:
...
+
+      - name: pallet-market — Run node benchmarks
+        run: |
+          cargo run \
+          -p polka-storage-node -r -F runtime-benchmarks -F testnet -- \
+          benchmark pallet \
+          --wasm-execution=compiled \
+          --pallet "pallet_market" \
           --extrinsic "*" \
           --steps 5 \
```
th7nder, line 148: Btw, are those numbers we chose explainable/reasonable? https://docs.polkadot.com/develop/parachains/testing/benchmarking/#run-benchmarks. On the `default tutorial` they have other ones.

Cargo.toml
```diff
@@ -80,7 +80,8 @@ indexmap = "2.2.6"
 integer-encoding = "4.0.0"
 ipld-core = "0.4.1"
 ipld-dagpb = "0.2.1"
-itertools = "0.13.0"
+# Use `use_alloc` for crates that compile to WASM
```
th7nder, line 83: How is this comment relevant? I don't understand it without your context :(

pallets/market/Cargo.toml
```diff
@@ -32,16 +35,37 @@ frame-system = { workspace = true, default-features = false }
...
+pallet-storage-provider = { workspace = true, default-features = false, optional = true }
+## Required for .chunks()
+itertools = { workspace = true, optional = true, default-features = false, features = ["use_alloc"] }
+## Required for random CIDs
+rand = { workspace = true, features = ["small_rng"], optional = true }
+## Substrate Primitives & Friends
+sp-core = { workspace = true, default_features = false, optional = true }
+sp-io = { workspace = true, optional = true }
+sp-keystore = { workspace = true, default-features = false, optional = true }
+sp-runtime = { workspace = true, optional = true }
```
th7nder: That's weird that we need to add a million dependencies for benchmarking. Why is that necessary? Wouldn't `dev-dependencies` be enough?

primitives/src/sector/pre_commit.rs
jmg-duarte: Nice catch

pallets/market/src/benchmarking.rs
```diff
@@ -0,0 +1,409 @@
...
+const COLLATERAL: u32 = 10;
+
+fn random_cid<R: RngCore>(rng: &mut R) -> Cid {
+    let mut buffer = [0u8; 32];
+    rng.fill_bytes(&mut buffer);
+
+    let hash = sha2_256(&buffer);
+    let mh = Multihash::<64>::wrap(SHA2_256_TRUNC254_PADDED, &hash).unwrap();
+    Cid::new_v1(FIL_COMMITMENT_UNSEALED, mh)
+}
```
th7nder:
```suggestion
fn commp() -> Cid {
    Commitment::<CommP>::zero(PaddedPieceSize::MIN).cid()
}

```

I don't see a reason to generate it randomly.

pallets/market/src/benchmarking.rs
th7nder: Do we need this randomness anyways? I don't think it has an impact on benchmarks, like at all. `piece_cid` contents shouldn't affect time complexity of the function being run.

pallets/market/src/benchmarking.rs
```diff
@@ -0,0 +1,409 @@
...
+    pub fn run_to_block<T>(n: u32)
+    where
+        T: crate::Config,
+        T: pallet_storage_provider::Config,
+    {
+        use frame_support::traits::Hooks;
+
+        while frame_system::Pallet::<T>::block_number() < n.into() {
+            if frame_system::Pallet::<T>::block_number() > 1u32.into() {
+                pallet_storage_provider::Pallet::<T>::on_finalize(
```
th7nder, line 153: Starting to wonder whether we should mock out the `pallet_storage_provider`. I mean, we're only using it here because `StorageProviderValidation` trait.

pallets/market/src/benchmarking.rs
```diff
@@ -0,0 +1,409 @@
...
+            MarketPallet::<T>::free(&account).unwrap(),
+            EXISTENTIAL_DEPOSIT.into()
+        );
+        assert_eq!(MarketPallet::<T>::locked(&account).unwrap(), 0u32.into());
+    }
+
+    /// Run until a particular block.
+    ///
+    /// Stolen't from: <https://github.com/paritytech/polkadot-sdk/blob/7df94a469e02e1d553bd4050b0e91870d6a4c31b/substrate/frame/lottery/src/mock.rs#L87-L98>
+    pub fn run_to_block<T>(n: u32)
```
th7nder, line 144: I know, I know, it's a stretch, BUT -> this function could be macroed as it's used across all of the pallets in the similar form.

pallets/market/src/benchmarking.rs
```diff
@@ -0,0 +1,409 @@
...
+    // When running `cargo test` the first get's called,
+    // for the proper benchmarks we get the second
+    #[cfg(test)]
+    const MAX_DEALS: u32 = 32;
+    #[cfg(not(test))]
+    const MAX_DEALS: u32 = 128;
+
+    /// `n`: number of submitted deals
+    #[benchmark]
+    fn publish_storage_deals(n: Linear<1, MAX_DEALS>) {
```
th7nder, line 243: Does the generated benchmark weight is set dynamically based on the number of deals passed to the extrinsic in the runtime?

pallets/market/src/benchmarking.rs
```diff
@@ -0,0 +1,409 @@
...
+        assert_eq!(balance_entry.free, 0u32.into());
+        assert_eq!(balance_entry.locked, 0u32.into());
+    }
+
+    // When running `cargo test` the first get's called,
+    // for the proper benchmarks we get the second
+    #[cfg(test)]
+    const MAX_DEALS: u32 = 32;
+    #[cfg(not(test))]
+    const MAX_DEALS: u32 = 128;
```
th7nder, line 239: Why do we have different values for test & ~test?

pallets/market/src/benchmarking.rs
```diff
@@ -0,0 +1,409 @@
...
+            .map(Vec::from_iter)
+            // Convert each chunk into a BoundedVec
+            .map(|sectors_for_call| {
+                BoundedVec::<
+                    SectorPreCommitInfo<BlockNumberFor<T>>,
+                    ConstU32<MAX_SECTORS_PER_CALL>
+                >::try_from(sectors_for_call).unwrap()
+            })
+            // Submit each "chunk"
+            .for_each(|pre_commit_infos| {
```
th7nder, line 338: Can you remind me why do we need to `pre_commit` and `prove` in the test to `settle_deal_payments`?
How about just setting the DealState to `Active` manually?

pallets/market/src/lib.rs
```diff
@@ -566,6 +580,79 @@ pub mod pallet {
...
         }
 
+        /// Publish a new set of storage deals (not yet included in a sector).
+        /// It saves valid deals as [`DealState::Published`] and locks up client fees and provider's collaterals.
+        /// Locked up balances cannot be withdrawn until a deal is terminated.
+        /// All of the deals must belong to a single Storage Provider.
+        /// It is permissive, if some of the deals are correct and some are not, it emits events for valid deals.
+        /// On success emits [`Event::<T>::DealPublished`] for each successful deal.
+        #[pallet::call_index(2)]
+        #[pallet::weight((T::WeightInfo::settle_deal_payments(deals.len() as u32), DispatchClass::Normal))]
```
th7nder: `settle_deal_payments` weight for `publish_storage_deals`?

th7nder (changes requested):

.github/workflows/ci.yaml
jmg-duarte: no, this opens a new option on the CI interface so that we can run manually from there

.github/workflows/ci.yaml
jmg-duarte: Yes they're explainable, here we JUST want the benchmarks to RUN. The benchmark process itself is ran manually and the parameters are placed on the generated `weights.rs` file

Cargo.toml
jmg-duarte: Itertools defaults to using `std`, we don't want `std` in pallets crates because they must compile to WASM

`use_alloc` uses `extern crate alloc` instead

pallets/market/Cargo.toml
jmg-duarte: No, benchmarks don't run under `test` when running on the node

pallets/market/src/benchmarking.rs
jmg-duarte: Because the deals will all be the same and generate conflicts

pallets/market/src/benchmarking.rs
jmg-duarte: The benchmark doesn't measure this part and the randomness MUST run under WASM which takes out most of the available randomness values. Since we're using this for testing, we don't need any assurances, making it the most adequate anyway

pallets/market/src/benchmarking.rs
jmg-duarte: It can't, different crates will have different requirements — i.e. have different `on_finalize` etc

pallets/market/src/benchmarking.rs
jmg-duarte: Yes

pallets/market/src/benchmarking.rs
jmg-duarte: Because the runtime configurations don't match the mock and fixing them all is a process in of itself

pallets/market/src/lib.rs
jmg-duarte: Good catch

pallets/market/src/benchmarking.rs
th7nder: They can be different just by changing labels for example.

pallets/market/src/benchmarking.rs
jmg-duarte: I can try but I don't see it bringing any benefits now
* This way the process is realistic & correct
* The benchmark doesn't measure this part of the set up
* Last (and least), I've already spent the time writing all the code to set this up

th7nder: Ok, you convinced me, I don't care that much.

pallets/market/src/benchmarking.rs
th7nder: Your call, not major.

pallets/market/src/benchmarking.rs
jmg-duarte: I removed the randomness

pallets/market/src/benchmarking.rs
jmg-duarte: raise this again if a similar pattern comes up again, for now - not worth it

th7nder (approved):

aidan46 (approved):

PR #706, opened 2025-01-28
feat: mater car reader read_block_metadata

cernicc:
### Description

Enable mater car readers to be able to read the blocks metadata while reading the car archive. If the `read_block_metadata` is used, that means that the block content is not loaded into memory and only the block's metadata is returned.

This is needed for the storage provider server for when the sector is being indexed. The provider will index the sectors after they are successfully prove committed. The indexing process opens unesaled sector files and reads the car blocks directly from the file. For indexing purposes, we don't need the content in the block. This pr enables the sector indexer to only read the metadata of the car block and saves that to the index.

### Checklist

### DIFF
--- a/mater/lib/src/cid.rs
+++ b/mater/lib/src/cid.rs
@@ -1,0 +1,88 @@
+use ipld_core::cid::{multihash::Multihash, CidGeneric, Error, Version};
+use tokio::io::{AsyncRead, AsyncReadExt};
+
+use crate::{async_varint::read_varint, IDENTITY_CODE};
+
+pub trait MultihashExt {
+    async fn read_async<R>(r: R) -> Result<(Self, usize), Error>
+    where
+        Self: Sized,
+        R: AsyncRead + Unpin;
+}
+
+pub trait CidExt {
+    async fn read_bytes_async<R>(r: R) -> Result<(Self, usize), Error>
+    where
+        Self: Sized,
+        R: AsyncRead + Unpin;
+
+    /// Returns Some(data) if the CID is an identity. If not, None is returned.
+    fn get_identity_data(&self) -> Option<&[u8]>;
+}
+
+impl<const S: usize> CidExt for CidGeneric<S> {
+    fn get_identity_data(&self) -> Option<&[u8]> {
+        (self.hash().code() == IDENTITY_CODE).then(|| self.hash().digest())
+    }
+
+    /// Async implementation of
+    /// https://github.com/multiformats/rust-cid/blob/eb03f566e9bfb19bad79b2691dbcb2541627c0b3/src/cid.rs#L143C12-L143C22
+    async fn read_bytes_async<R>(mut r: R) -> Result<(Self, usize), Error>
+    where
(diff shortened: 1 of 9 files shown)


mater/lib/src/v1/reader.rs
```diff
@@ -45,19 +46,38 @@ pub(crate) async fn read_block<R>(mut reader: R) -> Result<(Cid, Vec<u8>), Error
...
+    let data_size = full_block_length as usize - cid_bytes_read;
+    let mut data_buffer = vec![0; data_size];
+    reader.read_exact(&mut data_buffer).await?;
 
-    let data_start_position = full_block_cursor.position() as usize;
-    let mut full_block_buffer = full_block_cursor.into_inner();
+    Ok((cid, data_buffer))
+}
+
+pub(crate) async fn skip_block<R>(mut reader: R) -> Result<BlockMetadata, Error>
```
jmg-duarte: I feel that this is more accurate
```suggestion
pub(crate) async fn read_block_metadata<R>(mut reader: R) -> Result<BlockMetadata, Error>
```

mater/lib/src/cid.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+impl<const S: usize> CidExt for CidGeneric<S> {
+    fn get_identity_data(&self) -> Option<&[u8]> {
+        (self.hash().code() == IDENTITY_CODE).then(|| self.hash().digest())
+    }
+
+    /// Async implementation of
+    /// https://github.com/multiformats/rust-cid/blob/eb03f566e9bfb19bad79b2691dbcb2541627c0b3/src/cid.rs#L143C12-L143C22
+    async fn read_bytes_async<R>(mut r: R) -> Result<(Self, usize), Error>
```
jmg-duarte, line 30: Could you add a test for this function?

mater/lib/src/cid.rs
```diff
@@ -0,0 +1,88 @@
...
+                Ok((Self::new(version, codec, mh)?, bytes_read))
+            }
+        }
+    }
+}
+
+impl<const S: usize> MultihashExt for Multihash<S> {
+    /// Async implementation of
+    /// https://github.com/multiformats/rust-multihash/blob/90a6c19ec71ced09469eec164a3586aafeddfbbd/src/multihash.rs#L271
+    async fn read_async<R>(mut r: R) -> Result<(Self, usize), Error>
```
jmg-duarte, line 63: A test for this one too would be great

jmg-duarte (commented):
LGTM but there's some things I think could be improved

mater/lib/src/cid.rs
cernicc: Added 👍

mater/lib/src/cid.rs
cernicc: Added 👍

jmg-duarte (approved):
When I grow up, I want to be like you 🚀

th7nder (approved):

PR #710, opened 2025-01-30
fix: zero padding reader incomplete reads

### LINKED ISSUE #709: Cannot store arbitrary file because of padding
When I created a random file, using `mater convert` < 8MiB and tried to store it, invalid CommP issue arises when trying to store it in the real_world maat test.

th7nder:
### Description

Fixes #709.
On bigger files read with `ZeroPaddingReader` it started giving non-deterministic results as it was padding with 0 when it should not have.

### Checklist

  - [X] If yes, which ones?

### DIFF
--- a/lib/polka-storage-proofs/src/zero_reader.rs
+++ b/lib/polka-storage-proofs/src/zero_reader.rs
@@ -56,12 +56,12 @@ mod tests {
 
     /// Sole purpose of this reader is to simulate the file reading in the OS.
     /// When we give it a buf of a certain buf.length(), it might read n < buf.length().
-    struct IOSimulatingReader {
+    struct IncompleteReader {
         read_count: usize,
         served_data: Vec<Vec<u8>>,
     }
 
-    impl Read for IOSimulatingReader {
+    impl Read for IncompleteReader {
         fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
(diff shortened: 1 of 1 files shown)


lib/polka-storage-proofs/src/zero_reader.rs
```diff
@@ -49,6 +54,73 @@ mod tests {
 
     use super::ZeroPaddingReader;
 
+    /// Sole purpose of this reader is to simulate the file reading in the OS.
+    /// When we give it a buf of a certain buf.length(), it might read n < buf.length().
+    struct IOSimulatingReader {
+        read_count: usize,
+        served_data: Vec<Vec<u8>>,
+    }
```
jmg-duarte:
```suggestion
    /// Sole purpose of this reader is to simulate the file reading in the OS.
    /// When we give it a buf of a certain buf.length(), it might read n < buf.length().
    struct IncompleteReader {
        read_count: usize,
        served_data: Vec<Vec<u8>>,
    }
```

jmg-duarte (dismissed):
Great find, some would say "the best find", the bestest find even! 🚀

cernicc (dismissed):
💪

jmg-duarte (dismissed):

cernicc (approved):

jmg-duarte (approved):

PR #722, opened 2025-02-03
fix(storage-provider): query sectory expiration

### LINKED ISSUE #622: Query sector expiration margin from the chain

jmg-duarte:
### Description

Query the sector expiration bounds instead of using a constant.

There are some notes left on the code, please review.

Closes #622


storage-provider/server/src/pipeline/mod.rs
```diff
@@ -279,6 +284,29 @@ async fn add_piece(
     let mut sector = find_sector_for_piece(&state).await?;
 
     tracing::info!("Adding a piece...");
+
+    // NOTE(@jmg-duarte,31/01/2025): Unsure if getting the best block would be better
+    let current_block = state.xt_client.height(true).await?;
```
th7nder, line 292: Precommit depends on the last finalized block, so this should too.

th7nder (dismissed):
Great detective work!

storage-provider/server/src/pipeline/mod.rs
jmg-duarte: Clippy failed so I'll remove that too

cernicc (approved):
Nice

th7nder (approved):

PR #727, opened 2025-02-04
feat: Move bootstrap node into collator

### LINKED ISSUE #720: feat: Move P2P bootstrap into collator
Move the P2P bootstrapping node into the collator as a service

aidan46:
### Description

Move the P2P bootstrap node from the polka-storage-provider-server into the collator. All the examples have been updated to reflect this change.
The zombienet setup has been updated to include the p2p key and the p2p listen address.


node/src/cli.rs
```diff
@@ -109,3 +112,42 @@ impl RelayChainCli {
...
     }
 }
+
+#[derive(Debug, clap::Parser)]
+#[group(skip)]
+pub struct RunCmd {
+    #[clap(flatten)]
+    pub base: cumulus_client_cli::RunCmd,
+
+    /// P2P ED25519 private key
```
th7nder: P2P is too generic, I'd say something like:
```suggestion
    /// Key identifying nodes in p2p network of Storage Providers and Collators.
    /// It must be ED25519 private key, supplied as a short format XXXX or XXX or XXX.
```

(please fill the description of how it should look like)

node/src/cli.rs
```diff
@@ -109,3 +112,42 @@ impl RelayChainCli {
...
+    #[arg(long, value_parser = keypair_value_parser)]
+    pub p2p_key: Option<Keypair>,
+
+    /// Listen address that the bootstrap node binds to.
+    #[arg(long)]
+    pub p2p_listen_address: Option<Multiaddr>,
+}
+
+impl std::ops::Deref for RunCmd {
```
th7nder: Genuinely curious, why is it needed and how does this work?

node/src/cli.rs
```diff
@@ -109,3 +112,42 @@ impl RelayChainCli {
...
+    type Target = cumulus_client_cli::RunCmd;
+
+    fn deref(&self) -> &Self::Target {
+        &self.base
+    }
+}
+
+/// Parses a ED25519 private key into a Keypair.
+/// Takes in a private key or the path to a PEM file, depending on the @ prefix.
+pub(crate) fn keypair_value_parser(src: &str) -> Result<Keypair, String> {
```
th7nder: We use similar functions across the codebase, should we put somewhere and reuse it?

node/src/command.rs
```diff
@@ -225,6 +225,17 @@ pub fn run() -> Result<()> {
             let collator_options = cli.run.collator_options();
 
             runner.run_node_until_exit(|config| async move {
+                let bootstrap_config = if config.role.is_authority() {
+                    let p2p_key = cli.run.p2p_key.ok_or("Could not find p2p key")?;
```
th7nder:
```suggestion
                    let p2p_key = cli.run.p2p_key.ok_or("This node is configured as authority, so it will be used as Bootstrap node for Storage Provider & Collator network, but the key is missing. Set --p2p--key argument of the node.")?;
```

Let's try to make error messages which are shown to the user as concise and sufficient enough so they don't need to go to the source code to figure this one out.

node/src/service/mod.rs
```diff
@@ -332,6 +337,13 @@ pub async fn start_parachain_node(
         })
     };
 
+    if parachain_config.role.is_authority() {
+        let bootstrap_config = bootstrap_config.ok_or("Could not find bootstrap config")?;
+        task_manager
+            .spawn_handle()
+            .spawn("p2p", None, run_bootstrap_node(bootstrap_config));
+    }
```
th7nder:
```suggestion
    if let Ok(cfg) = bootstrap_config { 
        task_manager
            .spawn_handle()
            .spawn("p2p", None, run_bootstrap_node(bootstrap_config));
    }
```

We should be able assume that if bootstrap config is set, then the node is authority and is supposed to be the authority.

zombienet/local-testnet.toml
```diff
@@ -27,7 +27,13 @@ id = 1000
...
 # run charlie as parachain collator
 [[parachains.collators]]
-args = ["--detailed-log-output", "--pool-type=fork-aware", "-lparachain=debug,xcm=trace,runtime=trace,txpool=debug,basic-authorship=debug"]
+args = [
+  "--detailed-log-output",
+  "--p2p-key=@zombienet/private.pem",
+  "--p2p-listen-address=/ip4/127.0.0.1/tcp/62649",
+  "--pool-type=fork-aware",
+  "-lparachain=debug,xcm=trace,runtime=trace,txpool=debug,basic-authorship=debug",
+]
```
th7nder: It does also need changes to `maat/src/lib.rs` , we pass those args manually as well.

zombienet/private.pem
th7nder, line 1: Not sure whether it's a good idea to save any private keys on github.

zombienet/local-testnet.toml
th7nder: + `kube-testnet` as well.

th7nder (changes requested):

node/src/cli.rs
aidan46: I took this from [moonbeam](https://github.com/moonbeam-foundation/moonbeam/blob/26f41f7bb8ad6614a5143edc2ec121716b60b8dc/node/cli/src/cli.rs#L362). It is to prevent calling `cli.run.base` on [collator_options](https://docs.rs/cumulus-client-cli/latest/cumulus_client_cli/struct.RunCmd.html#method.collator_options) and [normalize](https://docs.rs/cumulus-client-cli/latest/cumulus_client_cli/struct.RunCmd.html#method.normalize) which are used is the codebase.

zombienet/private.pem
aidan46: I agree with that. I don't know how to generate the key and then pass it into the zombienet configuration. Do you have any ideas?

th7nder: We could specify the path in the hard path zombienet and add generation of the .pem key as part of the Justfile 🤔

node/src/cli.rs
cernicc: In my opinion, this is a hack. It's not to prevent calling of `cli.run.base` . It's used so that you don't have to explicitly call the `cli.run.base`. The call `cli.run.collator_options()` implicitly dereferences.

maat/src/lib.rs
```diff
@@ -92,6 +101,10 @@ pub fn local_testnet_config() -> NetworkConfig {
         .display()
         .to_string();
     let polka_storage_node_binary_path = binding.as_str();
+    let temp_dir = tempdir().unwrap();
+    let file_path = temp_dir.path().join("private_key.pem");
+    generate_pem_file(&file_path);
+    let p2p_arg = format!("--p2p-key={}", file_path.display());
```
jmg-duarte: IMO this becomes more explicit if you inline in the vec call, since they're all CLI args, it becomes easier to read in a single pass rather than going back here after reading the other ones

node/src/command.rs
```diff
@@ -225,6 +225,24 @@ pub fn run() -> Result<()> {
...
+                        .p2p_key
+                        .ok_or(
+                            "This node is configured as authority, so it will be used as Bootstrap node for Storage Provider & Collator network, but the key is missing. Set the --p2p-key argument of the node."
+                        )?;
+                    let p2p_listen_address = cli
+                        .run
+                        .p2p_listen_address
+                        .ok_or(
+                            "This node is configured as authority, so it will be used as Bootstrap node for Storage Provider & Collator network, but the listen address is missing. Set the --p2p-listen-address argument of the node."
+                        )?;
```
jmg-duarte, line 240: Did you make them optional just so you could provide a reason? 

Interesting approach! I'd double check if clap allows us to provide an "error" message ourselves but don't spend too much time on it, maybe leave a comment explaining that

primitives/src/p2p.rs
jmg-duarte, line 1: I suspect that this isn't no_std compatible, can you check and possibly gate?

Running the benchmarks will tell

maat/src/lib.rs
aidan46: moved the argument formatting inline to the vec.

primitives/src/p2p.rs
aidan46: Good catch, this did result in issues with benchmarking. I feature-gated this before in p2p but that did result in issues with the benchmarking. Moved it to std feature and now the benchmarks runs without issues.

node/src/command.rs
aidan46: ~I made them optional because these args are re-used for the relaychain (even though they only use relay_chain_args as far as I can tell) which is not supposed to run a bootstrap node.~

jmg-duarte: Sorry, I didn't understand, could you elaborate?

cernicc:
I am wondering. We already have a peerid which is derived from the provider's address. If I am not mistaken. That peerid is later used in deals etc. How is the p2p peer id used? I imagine those two peerid should be the same?

cernicc:
Sooo. I was thinking about the retrieval. Currently the p2p network and polkadot parachain store different peer ids for the same provider. That means that our custom p2p network only knows how to map the p2p peer id to the multiaddress. The idea for "retrieval name resolver" was that we fetch the peer id for a specific deal from the chain. But that currently wouldn't work because the one stored on the chain is different from the one used by the custom p2p network.

Please correct me if I am wrong.

aidan46:
> I am wondering. We already have a peerid which is derived from the provider's address. If I am not mistaken. That peerid is later used in deals etc. How is the p2p peer id used? I imagine those two peerid should be the same?

> Sooo. I was thinking about the retrieval. Currently the p2p network and polkadot parachain store different peer ids for the same provider. That means that our custom p2p network only knows how to map the p2p peer id to the multiaddress. The idea for "retrieval name resolver" was that we fetch the peer id for a specific deal from the chain. But that currently wouldn't work because the one stored on the chain is different from the one used by the custom p2p network.

> Please correct me if I am wrong.

@cernicc The Peer ID on chain and the Peer ID in the P2P network are both derived from the same ED25519 private key. When the storage provider first registers on-chain they [create an ED25519 private key](https://github.com/eigerco/polka-storage/blob/de7f373fddaa7cd2cf57683f9c408415fc535ff5/examples/start_sp.sh#L16), derive the [Peer ID using the polka-storage-client](https://github.com/eigerco/polka-storage/blob/de7f373fddaa7cd2cf57683f9c408415fc535ff5/examples/start_sp.sh#L22) and [register themselves on-chain with this Peer ID](https://github.com/eigerco/polka-storage/blob/de7f373fddaa7cd2cf57683f9c408415fc535ff5/examples/start_sp.sh#L34). Basically the Peer ID on-chain is something that is generated externally and is the same that the storage provider uses to register within the P2P network

node/src/command.rs
aidan46: I was wrong about the relay chain stuff, the arguments are simply optional because when any [Subcommand](https://github.com/eigerco/polka-storage/blob/de7f373fddaa7cd2cf57683f9c408415fc535ff5/node/src/cli.rs#L6) is run the P2P arguments are not needed and it will cause an error.

examples/rpc_publish.sh
```diff
@@ -29,6 +29,7 @@ CONFIG="/tmp/config.toml"
 P2P_PUBLIC_KEY="/tmp/public.pem"
 P2P_PRIVATE_KEY="/tmp/private.pem"
 P2P_ADDRESS="/ip4/127.0.0.1/tcp/62649"
+P2P_BOOTSTRAP_PEER_ID="12D3KooWJsSUCM8ZMHd6ms8YcE324raSWDHLdMN3caDSNNAoKLfH"
```
jmg-duarte: If the key is generated on the fly, wouldn't this be too?

node/src/command.rs
jmg-duarte: I'm sorry mate, but you're a step ahead and I still don't get it. If we want to force the user to register libp2p keys, why are they optional?

cargopete:
looks great. Should we update the documentation for p2p_key to specify that it is used both for on–chain registration and for the custom P2P network? This may reduce any confusion about why two different peer IDs aren’t generated

Also I like the approach of making the p2p arguments optional except when running as an authority. Perhaps we could have clap enforce these conditionally (if possible) to give users immediate feedback rather than only at runtime?

Justfile
```diff
@@ -24,8 +24,10 @@ release: lint
 release-testnet:
     cargo build --release --features polka-storage-runtime/testnet --bin polka-storage-node
 
-# Run the testnet without building
+# Generate a private key for the P2P network and
+# run the testnet without building
 run-testnet:
+    openssl genpkey -algorithm ED25519 -out /tmp/private.pem
```
th7nder: nit: maybe we should prefix this /tmp/polka-storage/private.pem, to avoid potential conflicts with different software.

th7nder (dismissed):

aidan46:
> looks great. Should we update the documentation for p2p_key to specify that it is used both for on–chain registration and for the custom P2P network? This may reduce any confusion about why two different peer IDs aren’t generated

I can expand on the docs for the P2P key.

> Also I like the approach of making the p2p arguments optional except when running as an authority. Perhaps we could have clap enforce these conditionally (if possible) to give users immediate feedback rather than only at runtime?

I would like this too, clap has a [requires](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.requires) option. Unfortunately this does not work for us because the authority is based on the [Role](https://paritytech.github.io/polkadot-sdk/master/polkadot_service/enum.Role.html) which is hidden in the [Configuration](https://paritytech.github.io/polkadot-sdk/master/polkadot_service/struct.Configuration.html) (gotten with [run_node_until_exit](https://paritytech.github.io/polkadot-sdk/master/sc_cli/struct.Runner.html#method.run_node_until_exit)) preventing us from setting the `requires` argument on it.

storage-provider/server/src/p2p/mod.rs
```diff
@@ -81,56 +57,11 @@ pub(crate) fn deser_keypair<'de, D: de::Deserializer<'de>>(d: D) -> Result<Keypa
...
-                return Err(e);
-            }
-        },
-        _ = token.cancelled() => {
-            tracing::info!("P2P node has been stopped by the cancellation token...");
-        },
-    }
-
-    Ok(())
+pub(crate) fn string_to_peer_id<'de, D: de::Deserializer<'de>>(d: D) -> Result<PeerId, D::Error> {
```
jmg-duarte: Especially given it is public, I believe this is a better name

```suggestion
pub(crate) fn deserialize_string_to_peer_id<'de, D: de::Deserializer<'de>>(d: D) -> Result<PeerId, D::Error> {
```

jmg-duarte (dismissed):
Left a suggestion, but looks good to me.

cernicc (dismissed):

zombienet/local-kube-testnet.toml
```diff
@@ -25,7 +25,12 @@ id = 1000
 
 # run charlie as parachain collator
 [[parachains.collators]]
-args = ["--detailed-log-output", "-lparachain=debug,xcm=trace,runtime=trace"]
+args = [
+  "--detailed-log-output",
+  "--p2p-key=@/tmp/private.pem",
```
th7nder: There is a mismatch, in some places it's `@/tmp/polka-storage/private.pem`, here it's still `@/tmp/private.pem`.

th7nder (changes requested):
One last thing to correct, besides that, LGTM!

zombienet/local-kube-testnet.toml
aidan46: You are right, fixed this. There were some other issues with keys not aligning too.

node/src/command.rs
aidan46: They are optional because not every command needs the keys and address to be passed in. We can make them required if we add a command (start node) or something. Currently, the Cli struct has the subcommand as an optional field but it still requires all the arguments for the RunCmd, forcing us to set the key and address as optional.

th7nder (approved):

cernicc (approved):

PR #729, opened 2025-02-04
fix: use older nixpkgs to make nix work on macos

th7nder:
### Description

https://github.com/NixOS/nixpkgs/pull/323869#issuecomment-2633773890

There was a flag added to the nixpkgs that breaks downstream builds.
Downgrading nixpkgs for now, it won't affect polkadot's version as we're not using nixpkgs for that.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -59,11 +59,11 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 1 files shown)


th7nder:
Actually, it broke something else. Closing for now, looking for another solution.

PR #731, opened 2025-02-05
fix(storage-provider): wait for deadline end on empty partition

jmg-duarte:
### Description

Before this PR:

https://github.com/user-attachments/assets/d2de2898-91a4-4986-81f7-ce4213e39958

This happened because the pre-commit for an empty deadline was being re-scheduled right after, meaning it would try to pre-commit the same deadline, over and over and over again until it passed.

Now, we wait for the deadline close before returning.


th7nder (approved):

cernicc (approved):
Great find

PR #733, opened 2025-02-05
feat(storage-provider): add time-based sector sealing

jmg-duarte:
### Description

Adds time-based sector sealing.

Schedules the Pre-Commit for later, when "later" arrives, tries to pre-commit.
The database keeps the source of truth for this operation.

Up next will be to consider the start block in the deals.

### Checklist

### DIFF
--- a/storage-provider/common/src/config/sealing.rs
+++ b/storage-provider/common/src/config/sealing.rs
@@ -8,7 +8,15 @@
 //! the full value in seconds (90s) instead of some mix of values (1m30s); this makes it easy to
 //! maintain for us and straightforward for the user to use.
 
+use std::time::Duration;
+
 use serde::{Deserialize, Deserializer, Serialize};
+
+/// Default amount of time to wait to seal an unfilled sector.
+const fn default_wait_deals_delay() -> Duration {
+    // 24 hours * 60 minutes * 60 seconds =
+    Duration::from_secs(24 * 60 * 60)
+}
 
 /// Returns the default fill percentage.
 const fn default_fill_threshold() -> u8 {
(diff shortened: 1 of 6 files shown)


cernicc (approved):

storage-provider/common/src/config/sealing.rs
```diff
@@ -54,6 +104,15 @@ pub struct SealingConfiguration {
...
     ))]
     pub fill_threshold: u8,
+
+    /// The amount of time to wait before sealing an unfilled sector, defaults to 24 hours.
+    /// If the sector is empty, it will not be sealed.
+    #[serde(
+        default = "default_wait_deals_delay",
+        deserialize_with = "duration_deserializer"
+    )]
+    #[cfg_attr(feature = "clap", arg(long, default_value = "24h", value_parser = duration_value_parser))]
```
th7nder, line 114: I think you are aware, but I just want to confirm that.
You know, that if sectors would be precommitted after 24h, this means that the `start_block` needs to be at least 24h in the future, because if sector is not precommited and proven before the start block -> it's going to be slashed.

to be precise: 24h + 1h needed for sealing and proving.

th7nder (approved):
I understand what this code does, looks messy and it's hard to follow. But I dont have any ideas for better alternatives, they'll come probably later, when we have better overview of the whole process with more puzzle pieces fit. 

Good for me for now.

storage-provider/common/src/config/sealing.rs
jmg-duarte: I placed 24h as an "easy" default, I didn't think too much about the value that it should actually be. 

But your concern is precisely the upcoming PR

PR #734, opened 2025-02-06
fix: Add required = false to p2p cli args

aidan46:
### Description

Small change to change the help section in the storage provider CLI. Previously the CLI showed p2p-key and rendezvous-point-address as required even tho it is not when passing in a config file when showing the usage.

Previous:
```
Usage: polka-storage-provider-server [OPTIONS] --p2p-key <P2P_KEY> --rendezvous-point-address <RENDEZVOUS_POINT_ADDRESS>
```
Now:
```
Usage: polka-storage-provider-server [OPTIONS]
```

### DIFF
--- a/storage-provider/server/src/config.rs
+++ b/storage-provider/server/src/config.rs
@@ -112,27 +112,27 @@ pub struct ConfigurationArgs {
 
     /// P2P Node type, can be either a bootstrap node or a registration node.
     #[serde(default = "NodeType::default")]
-    #[arg(long, default_value_t = NodeType::Bootstrap)]
+    #[arg(long, default_value_t = NodeType::Bootstrap, required = false)]
     pub(crate) node_type: NodeType,
 
     /// P2P ED25519 private key
     #[serde(deserialize_with = "deser_keypair")]
-    #[arg(long, value_parser = keypair_value_parser)]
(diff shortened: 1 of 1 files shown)


storage-provider/server/src/config.rs
```diff
@@ -112,27 +112,27 @@ pub struct ConfigurationArgs {
 
     /// P2P Node type, can be either a bootstrap node or a registration node.
     #[serde(default = "NodeType::default")]
-    #[arg(long, default_value_t = NodeType::Bootstrap)]
+    #[arg(long, default_value_t = NodeType::Bootstrap, required = false)]
```
th7nder, line 115: Are we validating that later, that if no configuration file is passed, it'll require p2p_key and others anyways?

th7nder (changes requested):

storage-provider/server/src/config.rs
jmg-duarte: It's the same logic applied to others. Especially with a default, a this parameter is not required anyway. 

This is a clap issue. Maybe if we moved to the builder approach it would work better, but for now, this is what we have

jmg-duarte (approved):

th7nder (approved):

PR #736, opened 2025-02-07
fix(storage-provider): make default body limit max sector size

jmg-duarte:
### Description

It also fixes errors that were the same.
Required to be merged before #733 

### Checklist

### DIFF
--- a/storage-provider/server/src/storage.rs
+++ b/storage-provider/server/src/storage.rs
@@ -92,8 +92,11 @@ fn configure_router(state: Arc<StorageServerState>) -> Router {
             .route(
                 "/upload/:cid",
                 put(upload)
-                    // Without disable we can't handle
-                    .layer(DefaultBodyLimit::disable()),
+                    // Limit upload size to maximum sector size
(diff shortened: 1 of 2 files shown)


storage-provider/server/src/storage.rs
```diff
@@ -80,14 +80,21 @@ pub async fn start_upload_server(
...
             .allow_headers([header::CONTENT_TYPE])
             .max_age(std::time::Duration::from_secs(3600));
 
         Router::new()
-            .route("/upload/:cid", put(upload))
+            .route(
+                "/upload/:cid",
+                put(upload)
+                    // Without disable we can't handle
+                    .layer(DefaultBodyLimit::disable()),
```
th7nder: Probably there is a reason for the default body limit, meaning DoS attacks. I'm happy to disable it now anyways I suppose.

th7nder (dismissed):

th7nder (approved):

cernicc (approved):

Issue #739, opened 2025-02-11
Storage Provider pallet benchmarks


Jinxit:
The storage provider pallet needs benchmarks in order to calculate the real weights. An initial implementation exists in #740, but there are still multiple extrinsics left to test.

- [x] `register_storage_provider` #740
- [x] `pre_commit_sectors` #740
- [x] `prove_commit_sectors` #818
- [x] `submit_windowed_post` #860
- [x] `declare_faults` #828
- [x] `declare_faults_recovered` #834
- [x] `terminate sectors` #837

jmg-duarte:
Closing this since there's no point in having a tracker for a single issue

PR #740, opened 2025-02-11
feat(storage-provider): benchmark register & pre_commit

Jinxit:
### Description

Here's a first attempt at benchmarking the storage-provider pallet.

For now I've only implemented `register_storage_provider` and `pre_commit_sectors` to get the code in front of you to review and see if you agree with how I've done it.

I've also added Justfile commands for running benchmarks.

**Why a separate crate?**

Putting the benchmarks in `pallet_storage_provider`, as is standard, would require a normal (non-dev) dependency on `pallet_market` which in turn is dependent on `pallet_storage_provider` - so it'd be cyclical. The solution (borrowed from [polkadot-sdk](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/session/benchmarking)) is to separate it in a different crate. There's some minor differences to how the benchmarks and the runtime are configured as a consequence of this, but mostly it's the same thing.

The most noticable impact of this for you is that benchmarks for the storage-provider are run with:
```sh
$ just bench-test storage-provider-benchmarks
```
for tests, but
```sh
$ just bench-node storage-provider
```
for the actual benchmarks.

### Checklist

  - [x] Separate crate
  - [x] #739
  - [x] The normal way of doing it didn't work because of the cyclic dependency


.vscode/settings.json
```diff
@@ -7,5 +7,19 @@
...
+  "rust-analyzer.cargo.features": [
+    "runtime-benchmarks"
+  ],
+  "rust-analyzer.procMacro.ignored": {
+    "frame_benchmarking": {
+      "v2::benchmarks": true,
+      "benchmarks": true,
+      "benchmark": true
+    }
+  },
```
jmg-duarte: These go out to everyone, could you explain what's happening here?

Justfile
jmg-duarte, line 1: Uhhh nice!

flake.nix
```diff
@@ -71,6 +71,7 @@
           RUST_SRC_PATH = "${rustToolchain}/lib/rustlib/src/rust/library/";
           # Workaround https://github.com/NixOS/nixpkgs/issues/370494#issuecomment-2625163369.
           CFLAGS = "-DJEMALLOC_STRERROR_R_RETURNS_CHAR_WITH_GNU_SOURCE";
+          CARGO_TARGET_DIR = "target/nix";
```
jmg-duarte: For the nix noob, what is this achieving?

Also, if I understand correctly, this breaks all our scripts

pallets/storage-provider/src/lib.rs
```diff
@@ -388,6 +386,8 @@ pub mod pallet {
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight(0)]
```
jmg-duarte: You don't need to add the weight while you have `dev_mode`, remove this or use the right value.

Same applies for the rest of the file

pallets/storage-provider/benchmarks/src/utils.rs
```diff
@@ -0,0 +1,52 @@
...
+pub fn sign_proposal<T: pallet_market::Config>(
+    pubkey: MultiSigner,
+    proposal: DealProposalOf<T>,
+) -> ClientDealProposalOf<T> {
+    let client_signature = create_ed25519_signature(&Encode::encode(&proposal), pubkey);
+    ClientDealProposal {
+        proposal,
+        client_signature,
+    }
+}
```
jmg-duarte: If these are the same as the ones in market, you can (try to [it doesn't always work correctly]) open a separate PR isolating these into the `primitives` crate (with a benchmark flag)

jmg-duarte (commented):
Looks good overall, some questions and notes though.

Regardless, good job!

.vscode/settings.json
Jinxit: Sure. I can commit this as comments too, but going by each:
- `targetDir` tells RA to build not in `target`, but in `target/rust-analyzer`. This means builds on the command line don't block RA, and vice versa, except for when they're touching the shared crate cache.
- `SKIP_WASM_BUILD` is a [polkadot flag](https://github.com/paritytech/polkadot-sdk/blob/f340e07ffd7cff69e930d19563186ff7f0fed601/substrate/utils/wasm-builder/src/lib.rs#L69-L71) to not build wasm binaries, which is pretty irrelevant for RA purposes.
- `features` toggles on extra features that are not enabled by default by the crate (in this case `runtime-benchmarks`), otherwise RA just ignores benchmark code completely in my experience.
- `procMacro.ignored` tells RA to completely ignore a certain proc-macro and pass through the inner code unmodified. I was experimenting with this to try to get better editor feedback in the benchmarks, but it's very undocumented and I'm not 100% sure it even does anything to an attribute macro. I'll take another look at it before merging.

flake.nix
Jinxit: It's doing the same as the vscode RA setting but for cargo inside the nix shell. If I remember my nix correctly, it should still _build_ to `target/` (thus not breaking scripts), but when developing with `nix develop` it'll work in a separate directory to reduce collisions and rebuilds.

Related question: What should I be running before creating a PR to make sure I didn't break anything?

aidan46: > What should I be running before creating a PR to make sure I didn't break anything?

It depends on the change but to cover everything you can run format, build, test and benchmark. If you have not touched anything in the pallet I don't think running benchmarks is needed. If you made changes to the pallet APIs you also need generate a new scale file (`just t and just generate-scale`)

pallets/storage-provider/benchmarks/src/inner.rs
aidan46, line 1: I am not a fan of this module name `inner`. Did you take this from a template somewhere? We could use more descriptive module names, the same goes for the `utils` module.

Jinxit: I took it straight from [polkadot-sdk](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/session/benchmarking/src) but I agree it's not great - but I suppose the alternative is having a benchmarks module in a benchmarks crate. I'm fine with either, what do you think?

flake.nix
th7nder: I don't think this change is needed. As for `CARGO_TARGET_DIR` for `rust-analyzer` I'm fine, but this, I don't understand it.

pallets/storage-provider/src/lib.rs
```diff
@@ -388,6 +386,8 @@ pub mod pallet {
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
```
th7nder, line 389: If we just benchmarked this crate to get proper weights, why are we setting weights to 0?
If we don't have proper weights for every extrinsic yet, let's just leave it be.

pallets/storage-provider/benchmarks/src/utils.rs
```diff
@@ -0,0 +1,52 @@
...
+    DealProposal<<Test as frame_system::Config>::AccountId, BalanceOf<Test>, BlockNumberFor<Test>>;
+
+pub fn generate_benchmark_account(name: &'static str) -> (AccountId32, MultiSigner) {
+    // Generate a deterministic seed
+    let seed = format!("//{}", name);
+
+    let signer: sp_core::ed25519::Public =
+        ed25519_generate(0.into(), Some(seed.into_bytes())).into();
+    let signer: MultiSigner = signer.into();
+    let account_id = signer.clone().into_account();
```
th7nder: Any reason for choosing ed25519 curve for signing? I mean, I don't mind, I just was used to `sr25519` being used as polkadot default and just curious.

pallets/storage-provider/benchmarks/src/utils.rs
```diff
@@ -0,0 +1,52 @@
...
+use frame_system::pallet_prelude::BlockNumberFor;
+use pallet_market::{BalanceOf, ClientDealProposal, DealProposal};
+use sp_core::ed25519;
+use sp_io::crypto::ed25519_generate;
+use sp_io::crypto::ed25519_sign;
+use sp_runtime::{traits::IdentifyAccount, AccountId32, MultiSignature, MultiSigner};
+use alloc::format;
+
+pub const ALICE: &'static str = "//Alice";
+#[allow(dead_code)]
```
th7nder: If they're unused, we should just put them in the test code/where they are used.

pallets/storage-provider/benchmarks/src/inner.rs
```diff
@@ -0,0 +1,226 @@
...
+};
+use sp_core::Get;
+use sp_runtime::{AccountId32, MultiSignature, MultiSigner};
+
+use crate::utils::{
+    generate_benchmark_account, sign_proposal, ClientDealProposalOf, DealProposalOf, ALICE, CHARLIE,
+};
+
+pub const MILLISECS_PER_BLOCK: u64 = 6000;
+pub const MINUTES: u64 = 60_000 / MILLISECS_PER_BLOCK;
```
th7nder: Those constants are defined but then not used? I see in the other part of code that we have `const MINUTES = 1`.

pallets/storage-provider/benchmarks/src/mock.rs
```diff
@@ -0,0 +1,191 @@
...
+    type MinDealDuration = MinDealDuration;
+    type MaxDealDuration = MaxDealDuration;
+    type MaxDealsPerBlock = ConstU32<500>;
+}
+
+parameter_types! {
+    // Storage Provider Pallet
+    pub const WPoStPeriodDeadlines: u64 = 10;
+    pub const WPoStProvingPeriod: BlockNumber = 40 * MINUTES;
+    pub const WPoStChallengeWindow: BlockNumber = 4 * MINUTES;
```
th7nder, line 90: I'm not sure whether those params are affecting benchmarks, however we should leave a note that those are `testnet` parameters and link to the source of them.

pallets/storage-provider/benchmarks/src/utils.rs
```diff
@@ -0,0 +1,52 @@
...
+    BalanceOf<T>,
+    BlockNumberFor<T>,
+    MultiSignature,
+>;
+pub type DealProposalOf<Test> =
+    DealProposal<<Test as frame_system::Config>::AccountId, BalanceOf<Test>, BlockNumberFor<Test>>;
+
+pub fn generate_benchmark_account(name: &'static str) -> (AccountId32, MultiSigner) {
+    // Generate a deterministic seed
+    let seed = format!("//{}", name);
```
th7nder: Bit confused, by default it formats as `"//{}", name` however from the function calls I can see that it's using `ALICE` which is `//Alice`, so it ends up being `////Alice`.

pallets/storage-provider/benchmarks/src/inner.rs
aidan46: I think the consts can be stored in the `lib.rs` file. The pallet type and config trait could be in a `pallet.rs` or `config.rs` file (I think you should keep them in the same module). I am not sure where you should place the benchmarking functions, maybe just a benchmarks module and re-export in `lib.rs`

runtime/src/lib.rs
```diff
@@ -627,6 +628,8 @@ impl_runtime_apis! {
 
             use cumulus_pallet_session_benchmarking::Pallet as SessionBench;
             impl cumulus_pallet_session_benchmarking::Config for Runtime {}
+            use pallet_storage_provider_benchmarks::Pallet as StorageProviderBench;
+            impl pallet_storage_provider_benchmarks::Config for Runtime {}
```
th7nder, line 632: Why do we need to add that here?

th7nder (changes requested):

pallets/storage-provider/benchmarks/src/utils.rs
Jinxit: It's kind of a mess in terms of dependencies and what's usable with externalities and what isn't. I gave it a shot but ultimately I think it's easier to duplicate code in this case.

pallets/storage-provider/benchmarks/src/utils.rs
Jinxit: I just used the same as the market crate, but I could switch to `sr25519` if it's preferred.

pallets/storage-provider/benchmarks/src/utils.rs
Jinxit: I'm thinking I'll probably need them for the other benchmarks 😄

flake.nix
Jinxit: Me and @jmg-duarte discussed it a bit yesterday and I'll move the vscode changes to a separate PR, and I'll remove this as it's redundant with the RA change.

pallets/storage-provider/benchmarks/src/inner.rs
Jinxit: You're absolutely right, just a leftover from earlier code.

runtime/src/lib.rs
Jinxit: Borrowed from [here](https://github.com/paritytech/polkadot-sdk/blob/e5df3306a5e741bc97b21daa60f25870be897b8c/templates/parachain/runtime/src/apis.rs#L286-L287) (corresponding to the two lines just above these), it's because of the benchmarks being in a different crate. Without them the benchmarks don't compile.

pallets/storage-provider/benchmarks/src/utils.rs
Jinxit: Good catch! Accident from combining the two methods for generating accounts. Will fix.

pallets/storage-provider/benchmarks/src/utils.rs
th7nder: Ok, no worries, I don't mind.

runtime/src/lib.rs
th7nder: ok

Jinxit:
I've pushed some changes, please take another look when you have time. I think I addressed all the comments, let me know if I missed something.

pallets/storage-provider/benchmarks/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+
+
+# Frame deps
+frame-benchmarking = { workspace = true, default-features = false, optional = true }
+frame-support = { workspace = true, default-features = false }
+frame-system = { workspace = true, default-features = false }
+
+[dev-dependencies]
+env_logger = { workspace = true }
+multihash-codetable = { workspace = true, features = ["blake2b"] }
```
jmg-duarte: I didn't see this being used

jmg-duarte (dismissed):
LGTM

aidan46 (dismissed):
LGTM!

pallets/storage-provider/benchmarks/src/benchmarking.rs
```diff
@@ -0,0 +1,204 @@
+#![cfg(feature = "runtime-benchmarks")]
```
cernicc: I think this is not needed, because the module is already gated in the `lib.rs`

pallets/storage-provider/benchmarks/src/mock.rs
```diff
@@ -0,0 +1,192 @@
+#![cfg(test)]
```
cernicc, line 1: Same reasoning

pallets/storage-provider/benchmarks/src/benchmarking.rs
jmg-duarte: Personally, i think that having both doesn't hurt, it shows the info in two places — i.e. if you open the module, it's obvious when it should compile and when you open the parent, the same applies

Jinxit: I also found it strange, but it's how it's done in the tutorials and how our other benchmarks look - but I'm not sure why.

th7nder (dismissed):

pallets/storage-provider/benchmarks/src/benchmarking/accounts.rs
```diff
@@ -0,0 +1,14 @@
+use sp_io::crypto::ed25519_generate;
+use sp_runtime::{traits::IdentifyAccount, AccountId32, MultiSigner};
+
+pub const ALICE: &'static str = "//Alice";
+pub const STORAGE_PROVIDER: &'static str = "//StorageProvider";
```
aidan46, line 5: We use these values in several crates in the repo. Maybe we should chuck them into `primitives` instead of redefining them every time :thinking: . No need to do this in this PR, I'm just making a not of it.

pallets/storage-provider/benchmarks/src/benchmarking/deal_proposals.rs
```diff
@@ -0,0 +1,32 @@
...
+        proposal,
+        client_signature,
+    }
+}
+
+pub fn create_ed25519_signature(payload: &[u8], pubkey: MultiSigner) -> MultiSignature {
+    let edpubkey = ed25519::Public::try_from(pubkey).unwrap();
+    let edsig = ed25519_sign(0.into(), &edpubkey, payload).unwrap();
+    edsig.into()
+}
```
aidan46, line 32: This is the same function as in market I believe. We could also extract this into `primitives`

Jinxit: This particular function sure, but the rest of the module is [kind of locked](https://github.com/eigerco/polka-storage/pull/740#discussion_r1952372737). I feel like it's small enough to duplicate rather than extract, especially since it's a test function with two unwraps.

pallets/storage-provider/benchmarks/src/benchmarking/accounts.rs
Jinxit: My personal guideline for this is whether we're likely to change the constant's value. If so it'd be nice to have it in a single place, but given that the constant is named the same as its value I think it'll keep the same value and thus is less complex to keep it duplicated. But it's not a hill I'll die on.

jmg-duarte (approved):
Great job figuring out all the details, looks nice 👌

aidan46 (approved):

PR #741, opened 2025-02-11
feat(mater): add unixfs support

cargopete:
adds unixfs support


mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+        // Parse the index according to the CARv2 spec. For demonstration,
+        // we assume a very simple format where each index entry is:
+        //   [CID length (u8)][CID bytes][offset (u64)][size (u64)]
+        let mut index = HashMap::new();
+        // In a real implementation you’d read until EOF or index length.
+        // Here we use a simple loop:
+        loop {
+            let cid_len = match reader.read_u8().await {
+                Ok(n) => n as usize,
+                Err(_) => break,
```
jmg-duarte: unhandled error

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+        mut reader: R,
+        index_offset: u64,
+    ) -> Result<HashMap<Cid, BlockLocation>, Error> {
+        // Seek to the start of the index.
+        reader.seek(SeekFrom::Start(index_offset)).await?;
+        // Parse the index according to the CARv2 spec. For demonstration,
+        // we assume a very simple format where each index entry is:
+        //   [CID length (u8)][CID bytes][offset (u64)][size (u64)]
+        let mut index = HashMap::new();
+        // In a real implementation you’d read until EOF or index length.
```
jmg-duarte: Isn't this a real implementation?

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+    /// For simplicity, assume the CAR header has been read and the index offset is known.
+    pub async fn load_index(
+        mut reader: R,
+        index_offset: u64,
+    ) -> Result<HashMap<Cid, BlockLocation>, Error> {
+        // Seek to the start of the index.
+        reader.seek(SeekFrom::Start(index_offset)).await?;
+        // Parse the index according to the CARv2 spec. For demonstration,
+        // we assume a very simple format where each index entry is:
+        //   [CID length (u8)][CID bytes][offset (u64)][size (u64)]
```
jmg-duarte: My reading of the spec is very different, where did you find this?

> The first byte(s) of a CARv2 index (at the "Index offset" position) contain an unsigned [LEB128](https://en.wikipedia.org/wiki/LEB128) integer ("varint") that indicates the index format type as a [Multicodec](https://github.com/multiformats/multicodec) code. The remaining bytes follow the encoding rules of that index format type. — https://ipld.io/specs/transport/car/carv2/#index-format

And then each index has its own format

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+        // In a real implementation you’d read until EOF or index length.
+        // Here we use a simple loop:
+        loop {
+            let cid_len = match reader.read_u8().await {
+                Ok(n) => n as usize,
+                Err(_) => break,
+            };
+            let mut cid_buf = vec![0u8; cid_len];
+            reader.read_exact(&mut cid_buf).await?;
+            let cid = Cid::try_from(cid_buf).map_err(|_| Error::InvalidCid)?;
```
jmg-duarte: This is not an invalid cid, it's a CID error. An invalid CID would be expecting v1 and getting v2.

This does raise the point that the error name isn't great

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+    /// Retrieve a block by its CID. This method uses the in-memory index
+    /// to seek directly to the block’s location.
+    pub async fn get_block(&mut self, cid: &Cid) -> Result<Vec<u8>, Error> {
+        if let Some(location) = self.index.get(cid) {
+            self.reader.seek(SeekFrom::Start(location.offset)).await?;
+            let mut buf = vec![0u8; location.size as usize];
+            self.reader.read_exact(&mut buf).await?;
+            Ok(buf)
+        } else {
+            Err(Error::BlockNotFound(cid.to_string()))
```
jmg-duarte: Not sure if this should be an option instead, making it more uniform with normal fallible getter APIs

mater/lib/src/stores/filestore.rs
```diff
@@ -90,6 +99,86 @@ where
...
+        index_offset.try_into().unwrap(),
+    );
+    writer.write_header(&header).await?;
+
+    let header_v1 = CarV1Header::new(vec![root.unwrap()]);
+    writer.write_v1_header(&header_v1).await?;
+
+    writer.finish().await?;
+
+    Ok(root.unwrap())
```
jmg-duarte: Same as above

mater/lib/src/stores/filestore.rs
```diff
@@ -90,6 +99,86 @@ where
...
+        single_width_index.into(),
+    ));
+    writer.write_index(&index).await?;
+
+    writer.get_inner_mut().rewind().await?;
+    let header = CarV2Header::new(
+        false,
+        car_v1_start.try_into().unwrap(),
+        (index_offset - car_v1_start).try_into().unwrap(),
+        index_offset.try_into().unwrap(),
```
jmg-duarte: Do not panic in a library

mater/lib/src/stores/filestore.rs
```diff
@@ -90,6 +99,86 @@ where
         }
     }
 
+    let index_offset = position;
```
jmg-duarte, line 104: From my understanding, this chunk is writing the index out?

mater/lib/src/unixfs/mod.rs
```diff
@@ -41,7 +42,105 @@ enum TreeNode {
...
+                    size: Some(link_info.encoded_data_length),
+                });
+                (
+                    raw_sum + link_info.raw_data_length,
+                    encoded_sum + link_info.encoded_data_length,
+                    links,
+                    sizes,
+                )
+            },
+        );
```
jmg-duarte: IMO this is much simpler and more readable:
```suggestion
        let mut total_raw_size: u64 = 0;
        let mut total_encoded_size: u64 = 0;
        let mut pb_links = Vec::with_capacity(children.len());
        let mut blocksizes = Vec::with_capacity(children.len());
        // Process all children in a single pass, gathering totals and building links and blocksizes
        for (cid, link_info) in children {
            pb_links.push(PbLink {
                cid,
                name: Some("".to_string()),
                size: Some(link_info.encoded_data_length),
            });
            blocksizes.push(link_info.raw_data_length);

            total_raw_size += link_info.raw_data_length;
            total_encoded_size += link_info.encoded_data_length;
        }
```

mater/lib/src/unixfs/mod.rs
```diff
@@ -41,7 +42,105 @@ enum TreeNode {
...
+            fanout: None,
+        };
+
+        // Encode UnixFS data
+        let mut data_buf = Vec::new();
+        {
+            let mut w = quick_protobuf::Writer::new(&mut data_buf);
+            unixfs_data.write_message(&mut w)?;
+        }
+
```
jmg-duarte, line 120: Good use of scoping 👍

mater/lib/src/unixfs/mod.rs
```diff
@@ -41,7 +42,105 @@ enum TreeNode {
...
+        // Create DAG-PB node
+        let pb_node = PbNode {
+            links: pb_links,
+            data: Some(data_buf.clone().into()),
```
jmg-duarte, line 124: Instead of cloning here, save the length on a variable and use later instead

mater/lib/src/unixfs/mod.rs
```diff
@@ -41,7 +42,105 @@ enum TreeNode {
...
+        let cid = Cid::new_v1(DAG_PB_CODE, mh);
+
+        let info = LinkInfo {
+            raw_data_length: data_buf.len() as u64,
+            encoded_data_length: encoded.len() as u64 + total_encoded_size,
+        };
+
+        Ok(((cid, encoded.into()), info))
+    }
+    fn encode_raw(self) -> Result<((Cid, Bytes), LinkInfo), Error> {
```
jmg-duarte:
```suggestion
    }
    
    fn encode_raw(self) -> Result<((Cid, Bytes), LinkInfo), Error> {
```

mater/lib/src/stores/filestore.rs
```diff
@@ -6,10 +6,19 @@ use tokio::io::{AsyncRead, AsyncReadExt, AsyncSeek, AsyncSeekExt, AsyncWrite};
...
-    multicodec::SHA_256_CODE, unixfs::stream_balanced_tree, CarV1Header, CarV2Header, CarV2Writer,
-    Error, Index, IndexEntry, MultihashIndexSorted, SingleWidthIndex,
+    multicodec::SHA_256_CODE,
+    unixfs::{stream_balanced_tree, stream_balanced_tree_unixfs},
+    CarV1Header, CarV2Header, CarV2Writer, Error, Index, IndexEntry, MultihashIndexSorted,
+    SingleWidthIndex,
 };
 
+/// Converts a source stream into a CARv2 file and writes it to an output stream.
+//
```
jmg-duarte: No need for the empty comment line
```suggestion
```

mater/lib/src/stores/filestore.rs
```diff
@@ -90,6 +99,86 @@ where
...
+    mut source: Src,
+    mut output: Out,
+    chunk_size: usize,
+    tree_width: usize,
+) -> Result<Cid, Error>
+where
+    Src: AsyncRead + Unpin,
+    Out: AsyncWrite + AsyncSeek + Unpin,
+{
+    let chunker = async_stream::try_stream! {
```
jmg-duarte, line 150: If you're using the same approach as the other balanced import:
1. Reference to the other implementation for details (someone may read this and question why it's done the way it it)
2. Maybe there's a way of having this chunk be the same for both? (no macros)

mater/lib/src/stores/mod.rs
```diff
@@ -24,24 +24,38 @@ pub enum Config {
...
     }
+
+    /// Creates a new balanced tree configuration with UnixFS wrapping (recommended).
+    pub fn balanced_unixfs(chunk_size: usize, tree_width: usize) -> Self {
+        Self::balanced(chunk_size, tree_width, false)
+    }
+
+    /// Creates a new balanced tree configuration with raw storage.
+    pub fn balanced_raw(chunk_size: usize, tree_width: usize) -> Self {
+        Self::balanced(chunk_size, tree_width, true)
```
jmg-duarte: If VSCode isn't tricking me, these are only used in tests, I think just the `balanced` is enough.

jmg-duarte (changes requested):

mater/lib/src/stores/filestore.rs
```diff
@@ -90,6 +99,86 @@ where
...
+                let chunk = buf.split_to(chunk_size);
+                yield chunk.freeze();
+            }
+
+            if read_bytes == 0 && !buf.is_empty() {
+                let chunk = buf.split();
+                yield chunk.freeze();
+                break;
+            } else if read_bytes == 0 {
+                break;
```
th7nder: I think I understand what was going on.
If `chunk_size =2`, but you read_bytes == 0, and there is still `buf.len() == 4` from the previous invocation, you should not yield it in its entirety, but continue chunking it.
It makes sense, HOWEVER.

Then there is a bug/or it should be panic in here:
```
            if read_bytes == 0 && !buf.is_empty() {
                let chunk = buf.split();
                yield chunk.freeze();
                break;
```
because if we're chunking, and loop above gave all of the nice sized chunks, why we're yielding something which is not `chunk_size`? Is this intended?

Imagine, `buf.len() == 5`, `chunk_size == 2`, it goes:
```
yield 2;
yield 2;
yield 1;
```
so the last yielded chunk won't be a nice chunk.


.....

Then I took a look at: `let mut buf = BytesMut::with_capacity(chunk_size);`, on the first look it looks like it's always going to read `source.read_buf(&mut buf).await?`, chunk_size into buf. But is it? 

Is `BytesMut::with_capacity(chunk_size)` guaranteeing that `source.read_buf(&mut buf).await?` will read always at most `chunk_size` bytes? Quick search told me that it depends on the behaviour of underlying `read_buf`, so not sure.

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+                let mut cursor = std::io::Cursor::new(&block_bytes);
+                // Propagate any error that occurs during decoding.
+                let pb_node: ipld_dagpb::PbNode =
+                    DagPbCodec::decode(&mut cursor).map_err(Error::DagPbError)?;
+                for link in pb_node.links {
+                    if !processed.contains(&link.cid) {
+                        to_process.push(link.cid);
+                    }
+                }
+            }
```
th7nder:
```suggestion
            if current_cid.codec() == crate::multicodec::DAG_PB_CODE {
                let mut cursor = std::io::Cursor::new(&block_bytes);
                // Propagate any error that occurs during decoding.
                let pb_node: ipld_dagpb::PbNode =
                    DagPbCodec::decode(&mut cursor).map_err(Error::DagPbError)?;
                for link in pb_node.links {
                    if !processed.contains(&link.cid) {
                        to_process.push(link.cid);
                    }
                }
            } else {
            	return Err(Error::UnsupportedCidCodec(current_cid.codec()));
            }
```

I think we should return an error when we have different codec than we support.

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+    {
+        // To avoid processing a block more than once.
+        let mut processed = HashSet::new();
+        // We use a stack for DFS traversal.
+        let mut to_process = vec![*root];
+
+        while let Some(current_cid) = to_process.pop() {
+            if processed.contains(&current_cid) {
+                continue;
+            }
```
th7nder: This logic is duplicate in the DFS. We're checking it already here.
```
 if !processed.contains(&link.cid) {
 	to_process.push(link.cid);
}
```

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+            processed.insert(current_cid);
+
+            // Retrieve block by CID via the index.
+            let block_bytes = self.get_block(&current_cid).await?;
+
+            // Write the raw block data. In a real UnixFS traversal you might need
+            // to reconstruct file content in order.
+            output.write_all(&block_bytes).await?;
```
th7nder: How is that not real UnixFS traversal?
I get confused, what extract_content_via_index is supposed to do?
What do we mean by index if actually we are not using index anywhere?

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,139 @@ mod unixfs;
...
+                    }
+                }
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl<R> CarBlockStore<R>
```
th7nder: Can't we pull it in under one `impl` block? they seem to be generic over the same trait bounds.

I don't know why one is `AsyncSeek`, the other `AsyncSeekExt` though.

mater/lib/src/stores/filestore.rs
```diff
@@ -173,6 +274,92 @@ mod test {
         assert_buffer_eq!(&expected, &result);
     }
 
+    #[tokio::test]
+    async fn test_filestore_unixfs_dag_structure() {
```
th7nder, line 293: Can we add a test which checks whether this is somehow similar to the reference implementation?

th7nder (changes requested):

mater/lib/src/stores/filestore.rs
jmg-duarte: Yes please, we need to CAR-ify a large-ish file (like glenda) and commit it to do a byte-check

mater/lib/src/stores/filestore.rs
cargopete: Yeah this code is indeed where we write out the CARv2 index. We record the current file position in index_offset so that the CARv2 header can point to the start of this index. Right after that, the `writer.write_index(&index)` call serializes `MultihashIndexSorted` to the file, effectively appending the index bytes at position (the `index_offset`). Later, when we rewind and write the final headers, we embed `index_offset` so that any CARv2 reader knows where the index begins.

mater/lib/src/stores/filestore.rs
cargopete: We do allow partial chunks at the end. If `read_bytes == 0` but buf still has data, we yield whatever remains (`buf.split()`), even if it’s smaller than `chunk_size`. This is intentional so we don’t discard leftover bytes.
the loop `while buf.len() >= chunk_size { ... }` already split off as many full chunks as possible. If there’s anything left, we yield that as a “final” partial chunk. In your example with buf.len() == 5 and chunk_size = 2, we do yield 2, yield 2, then yield 1—that last chunk is legitimately smaller.
`BytesMut::with_capacity(chunk_size)` sets the initial capacity to `chunk_size`, but does not guarantee `read_buf` will only read exactly chunk_size or less. `read_buf` can read as much as fits in the allocated capacity—Rust I/O doesn’t forcibly limit to `chunk_size`. That’s why we keep splitting off multiple chunks if buf ends up with more data.
So, there’s no bug or panic needed imho, partial final chunks are expected behavior for many chunking scenarios. The code aims for maximum `chunk_size`, but also gracefully handles leftovers for the last chunk.

mater/lib/src/lib.rs
cargopete: its fine imo it was a leftover comment sry

mater/lib/src/stores/filestore.rs
cargopete: added comment

mater/lib/src/lib.rs
jmg-duarte: You just added my comment to the docs but didn't change the code. ~I still don't understand how this code loads the index correctly.~ It doesn't, according to:
 https://github.com/eigerco/polka-storage/blob/a5c4eb14f1514ff7b4369194389d09c6ffdf7c02/mater/lib/src/lib.rs#L78-L80

There's a full implementation already, please use it:
https://github.com/eigerco/polka-storage/blob/9a1f8c488b53802e71695836fc306aaec63580f5/mater/lib/src/v2/index.rs#L292-L304

mater/lib/src/stores/filestore.rs
```diff
@@ -6,16 +6,24 @@ use tokio::io::{AsyncRead, AsyncReadExt, AsyncSeek, AsyncSeekExt, AsyncWrite};
...
+//   between threads and ensure they live long enough for the entire async operation
+// - `AsyncSeek`: Required for the output to write the final header after processing all blocks
+// - `Unpin`: Required because we need to move the source/output around during async operations
 async fn balanced_import<Src, Out>(
     mut source: Src,
     mut output: Out,
     chunk_size: usize,
     tree_width: usize,
-) -> Result<Cid, Error>
+) -> Result<Option<Cid>, Error>
```
jmg-duarte, line 26: Our implementation considers it invalid to have a missing root — https://ipld.io/specs/transport/car/carv1/#number-of-roots

There was an error being returned in case the root was `None`, put it back and apply the same logic to the other balanced import.

mater/lib/src/stores/filestore.rs
```diff
@@ -76,18 +84,111 @@ where
...
         let (node_cid, node_bytes) = node?;
-        let digest = node_cid.hash().digest().to_owned();
+        let digest = node_cid.hash().digest().to_vec();
+
+        // Index entry for the block
         let entry = IndexEntry::new(digest, (position - car_v1_start) as u64);
         entries.push(entry);
+
+        // Write the block
         position += writer.write_block(&node_cid, &node_bytes).await?;
```
jmg-duarte, line 100: A slice's `to_owned` implementation is a `to_vec`, but the function call was selected on purpose to *communicate intent*. The comments after repeat what the struct (`IndexEntry`) and function `write_block` already say
```suggestion
        let digest = node_cid.hash().digest().to_owned();
        let entry = IndexEntry::new(digest, (position - car_v1_start) as u64);
        entries.push(entry);
        position += writer.write_block(&node_cid, &node_bytes).await?;
```

mater/lib/src/stores/filestore.rs
```diff
@@ -137,20 +238,33 @@ where
...
+            raw,
+        } => {
+            if raw {
+                let maybe_cid = balanced_import(source, output, chunk_size, tree_width).await?;
+                let cid = maybe_cid.ok_or(Error::EmptyRootsError)?;
+                Ok(cid)
+            } else {
+                let cid = balanced_import_unixfs(source, output, chunk_size, tree_width).await?;
+                Ok(cid)
+            }
```
jmg-duarte, line 250: The signatures match already, no need for those additions
```suggestion
            if raw {
                balanced_import(source, output, chunk_size, tree_width).await
            } else {
                balanced_import_unixfs(source, output, chunk_size, tree_width).await
            }
```

mater/lib/src/stores/filestore.rs
```diff
@@ -76,18 +84,111 @@ where
...
+    let header_v1 = CarV1Header::new(root.into_iter().collect());
+    writer.write_v1_header(&header_v1).await?;
+
+    writer.finish().await?;
+    Ok(header_v1.roots.first().cloned())
+}
+
+/// Similar to [`balanced_import`], this function also creates a “balanced” CARv2 layout.
+// Unlike [`balanced_import`], it handles UnixFS blocks and ensures there is always at
+// least one root (returning an error otherwise).
```
jmg-duarte, line 131: Revert

mater/lib/src/stores/filestore.rs
```diff
@@ -76,18 +84,111 @@ where
...
+// We keep this separate from [`balanced_import`] because:
+// UnixFS Data: This function encodes blocks differently, wrapping them with UnixFS
+//   metadata rather than raw.
+// Guaranteed Root: We return an error if no root is found, whereas [`balanced_import`]
+//   allows for an empty result (`Option<Cid>`).
```
jmg-duarte, line 136: Revert

res.log
jmg-duarte, line 1: Remove

mater/lib/src/stores/filestore.rs
```diff
@@ -173,6 +289,141 @@ mod test {
...
+        let ref_path = Path::new("tests/fixtures/car_v2/spaceglenda_unixfs.car");
+
+        let output_file = File::create(&ref_path).await?;
+
+        let config = Config::balanced(DEFAULT_CHUNK_SIZE, DEFAULT_TREE_WIDTH, false);
+        let source_file = File::open(&source_path).await?;
+        create_filestore(source_file, output_file, config).await?;
+
+        Ok(())
+    }
```
jmg-duarte, line 425: This is not correct, a reference file is generated with go-car.

Remove

mater/lib/src/stores/filestore.rs
th7nder: Ok, if this behaviour with last chunk being legitimately smaller is fine, I'm cool with that.

mater/lib/src/lib.rs
th7nder: kk

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,178 @@ mod unixfs;
...
+    ///   you’ll need to extend this function accordingly.
+    ///
+    /// # Errors
+    ///
+    /// - Returns [`Error::IoError`] if any non-EOF I/O error occurs.
+    /// - Returns [`Error::InvalidCid`] if CID decoding fails.
+    /// - Returns [`Error::BlockNotFound`] if a requested block is missing (in `get_block`).
+    pub async fn load_index(
+        mut reader: R,
+        index_offset: u64,
```
th7nder, line 101: According to the spec, if `index_offset` is 0, it means that the index is not there. Are we handling that? (https://ipld.io/specs/transport/car/carv2/#index-payload)

mater/lib/src/lib.rs
```diff
@@ -18,16 +18,178 @@ mod unixfs;
...
+        // Seek to the start of the index.
+        reader.seek(SeekFrom::Start(index_offset)).await?;
+
+        let mut index = HashMap::new();
+
+        // Continuously parse index entries until EOF.
+        loop {
+            let cid_len = match reader.read_u8().await {
+                Ok(byte) => byte as usize,
+                Err(e) if e.kind() == ErrorKind::UnexpectedEof => {
```
th7nder, line 112:
```suggestion
		// Index is always at the end of a file that's why UnxpectedEof means end of the index.
                Err(e) if e.kind() == ErrorKind::UnexpectedEof => {
```

Adding this suggestion as I was very confused when I first saw this code.

mater/lib/src/stores/filestore.rs
```diff
@@ -76,18 +84,111 @@ where
...
+    tree_width: usize,
+) -> Result<Cid, Error>
+where
+    Src: AsyncRead + Unpin,
+    Out: AsyncWrite + AsyncSeek + Unpin,
+{
+    let chunker = async_stream::try_stream! {
+        let mut buf = BytesMut::with_capacity(chunk_size);
+        loop {
+            let read_bytes = source.read_buf(&mut buf).await?;
```
th7nder, line 153: If we improved this implementation, why it wasn't needed to change the same chunker in the `balanced_import`?

th7nder (changes requested):

jmg-duarte:
Closing this PR since: 
* it was not wrapping the files in the right way (diverges from the reference go-car implementation) 
* extraction was implemented in #753 
* retrieval was unblocked in the meantime

PR #744, opened 2025-02-12
feat: Implement p2p bootstrap improvement

### LINKED ISSUE #715: feat: Implement P2P bootstrap improvement
Based on the RFC described in #714, implement the needed improvements on the P2P network.

aidan46:
### Description

This PR adds gossipsub and request-response libp2p protocols to the bootstrap node.
The gossipsub protocol is used for bootstrap nodes to exchange peer information between each other, storing the information in a hashmap.
The request-response protocol is used for clients to request a peer id to multiaddrs mapping.

#### Layout of the P2P connections

p2p

### Testing

- build and run the zombienet with `just t`
- run the david collator with `just run-collator`
- run the start_sp script in example `example/start_sp.sh`
- wait for registration to complete, while running the script keep note of the generated peer id, we will look up the multiaddrs using this peer ID.
```shell
examples/start_sp.sh
Generated new peer ID for //Charlie: 12D3KooWLDeoZ3MCdCgjdoEtymitioS7Q1r2u3CUme45XMQpxLPg
```
`12D3KooWLDeoZ3MCdCgjdoEtymitioS7Q1r2u3CUme45XMQpxLPg` will be our resolve-id
- check the collator david's log and search for the multiaddr and peer ID:
```shell
2025-02-12 15:05:25.944  INFO tokio-runtime-worker libp2p_swarm: [Parachain] local_peer_id=12D3KooWPNWqWXMGqufZErv8v6dBamCHNpFv58WXgWShfdB8tUCz <- we need this
2025-02-12 15:05:25.944  INFO tokio-runtime-worker polka_storage_node::service::p2p::bootstrap: [Parachain] Starting P2P bootstrap node at /ip4/127.0.0.1/tcp/1337 <- we need this
2025-02-12 15:05:25.944  INFO tokio-runtime-worker polka_storage_node::service::p2p::bootstrap: [Parachain] Dialing /ip4/127.0.0.1/tcp/62649    
2025-02-12 15:05:25.944  INFO tokio-runtime-worker polka_storage_node::service::p2p::bootstrap: [Parachain] Listening on /ip4/127.0.0.1/tcp/1337    
``` 
`12D3KooWPNWqWXMGqufZErv8v6dBamCHNpFv58WXgWShfdB8tUCz` will be out bootstrap-id
`/ip4/127.0.0.1/tcp/1337` will be our bootstrap-addr

- build and run the peer-resolver example with the values you extracted.
```
cargo build -p polka-storage-provider-client --example peer-resolver
...
RUST_LOG=info ./target/debug/examples/peer-resolver --bootstrap-addr "/ip4/127.0.0.1/tcp/1337" --bootstrap-id "12D3KooWPNWqWXMGqufZErv8v6dBamCHNpFv58WXgWShfdB8tUCz" --resolve-id "12D3KooWLDeoZ3MCdCgjdoEtymitioS7Q1r2u3CUme45XMQpxLPg"
2025-02-12T14:24:48.128725Z  INFO libp2p_swarm: local_peer_id=12D3KooWEgFkcaEYFyYent4UXwekLKPR8fqX2LTSNmqH7YsbSnos
Attempting to get multiaddrs for peer PeerId("12D3KooWLDeoZ3MCdCgjdoEtymitioS7Q1r2u3CUme45XMQpxLPg")
2025-02-12T14:24:48.142060Z  INFO peer_resolver: Connected to 12D3KooWPNWqWXMGqufZErv8v6dBamCHNpFv58WXgWShfdB8tUCz
2025-02-12T14:24:48.263090Z  INFO peer_resolver: Received response with id 1 from 12D3KooWPNWqWXMGqufZErv8v6dBamCHNpFv58WXgWShfdB8tUCz
Got multiaddrs [/ip4/127.0.0.1/tcp/46706] for peer PeerId("12D3KooWLDeoZ3MCdCgjdoEtymitioS7Q1r2u3CUme45XMQpxLPg")
```

These are quite a few steps. Should I write a script to test all of this?

### Improvements

Currently, bootstrap nodes do not recover their in memory db in the event that they go down.
This recovery can either be done by requesting all peers through the gossipsub protocol or adding an on-disk database that the bootstrap node writes to.

### DIFF
--- a/node/src/service/p2p/bootstrap.rs
+++ b/node/src/service/p2p/bootstrap.rs
@@ -18,7 +18,7 @@ use libp2p::{
 use log::{debug, error, info, warn};
 use primitives::p2p::{
     PeerIdRequest, PeerInfo, PeerInfoResponse, DEFAULT_REGISTRATION_TTL, GOSSIP_TOPIC,
-    IDENTIFY_PROTOCOL_VERSION, REQUEST_RESPONSE_STREAM_PROTOCOL
+    IDENTIFY_PROTOCOL_VERSION, REQUEST_RESPONSE_STREAM_PROTOCOL,
 };
 
 use crate::service::p2p::P2PError;
--- a/node/src/service/p2p/bootstrap.rs
+++ b/node/src/service/p2p/bootstrap.rs
@@ -16,11 +16,12 @@ use libp2p::{
     tcp, yamux, Multiaddr, PeerId, StreamProtocol, Swarm, SwarmBuilder,
 };
 use log::{debug, error, info, warn};
-use primitives::p2p::{PeerIdRequest, PeerInfo, PeerInfoResponse};
+use primitives::p2p::{
+    PeerIdRequest, PeerInfo, PeerInfoResponse, DEFAULT_REGISTRATION_TTL, GOSSIP_TOPIC,
+    IDENTIFY_PROTOCOL_VERSION, REQUEST_RESPONSE_STREAM_PROTOCOL
+};
 
-use crate::service::p2p::{P2PError, DEFAULT_REGISTRATION_TTL};
-
-const GOSSIP_TOPIC: &str = "registrar";
+use crate::service::p2p::P2PError;
 
 #[derive(NetworkBehaviour)]
 pub struct BootstrapBehaviour {
@@ -76,7 +77,7 @@ impl BootstrapConfig {
                     ),
                     // The identify behaviour is used to share the external address and the public key with connecting clients.
                     identify: identify::Behaviour::new(identify::Config::new(
-                        "identify/1.0.0".to_string(),
+                        IDENTIFY_PROTOCOL_VERSION.to_string(),
                         key.public(),
                     )),
                     gossipsub: gossipsub::Behaviour::new(
@@ -85,7 +86,7 @@ impl BootstrapConfig {
                     )?,
                     request_response: request_response::cbor::Behaviour::new(
                         [(
-                            StreamProtocol::new("/resolver/1.0.0"),
+                            StreamProtocol::new(REQUEST_RESPONSE_STREAM_PROTOCOL),
                             ProtocolSupport::Full,
                         )],
                         request_response::Config::default(),
@@ -175,20 +176,11 @@ fn on_rendezvous_event(
             let encoded_peer_info = match bincode::serialize(&peer_info) {
                 Ok(info) => info,
                 Err(..) => {
-                    error!("Failed to serialize peer_info");
+                    error!(peer_info:?; "Failed to serialize peer_info");
                     return;
                 }
             };
-            registrations
-                .entry(peer)
-                .and_modify(|info| {
-                    for addr in peer_info.multiaddrs.iter() {
-                        if !info.multiaddrs.contains(addr) {
-                            info.multiaddrs.push(addr.clone())
-                        }
-                    }
-                })
-                .or_insert(peer_info);
+            insert_or_update_registrations(registrations, peer_info);
             // Send registration information to other bootstrap nodes.
             match swarm
                 .behaviour_mut()
@@ -196,7 +188,7 @@ fn on_rendezvous_event(
                 .publish(IdentTopic::new(GOSSIP_TOPIC), encoded_peer_info)
             {
                 Ok(..) => info!("Successfully published new peer info for peer {peer}"),
-                Err(..) => error!("Failed to publish new peer info for peer {peer}"),
+                Err(e) => error!(e:?; "Failed to publish new peer info for peer {peer}"),
             }
         }
         other => debug!("Encountered other rendezvous event: {other:?}"),
@@ -222,7 +214,7 @@ fn on_gossipsub_event(
             let peer_info: PeerInfo = match bincode::deserialize(&message.data) {
                 Ok(info) => info,
                 Err(..) => {
-                    error!("Received invalid peer info");
+                    error!(message:? = message.data; "Received invalid peer info from peer {peer_id:?}");
                     return;
                 }
             };
@@ -230,16 +222,7 @@ fn on_gossipsub_event(
                 "Got registration: {:?} with id: {} from peer: {:?}",
                 peer_info, id, peer_id
             );
-            registrations
-                .entry(peer_info.peer_id)
-                .and_modify(|info| {
-                    for addr in peer_info.multiaddrs.iter() {
-                        if !info.multiaddrs.contains(addr) {
-                            info.multiaddrs.push(addr.clone())
-                        }
-                    }
-                })
-                .or_insert(peer_info);
+            insert_or_update_registrations(registrations, peer_info);
         }
         other => debug!("Encountered other gossipsub event: {other:?}"),
     }
@@ -298,4 +281,22 @@ fn on_request_response_event(
             debug!("Request with id {request_id} sent to {peer}")
         }
     }
+}
+
+/// Take the HashMap and update the entry based on peer_info.peer_id
+/// or insert a new entry.
+fn insert_or_update_registrations(
+    registrations: &mut HashMap<PeerId, PeerInfo>,
+    peer_info: PeerInfo,
+) {
+    registrations
+        .entry(peer_info.peer_id)
+        .and_modify(|info| {
+            for addr in peer_info.multiaddrs.iter() {
+                if !info.multiaddrs.contains(addr) {
+                    info.multiaddrs.push(addr.clone())
+                }
+            }
+        })
+        .or_insert(peer_info);
 }
--- a/node/src/service/p2p/bootstrap.rs
+++ b/node/src/service/p2p/bootstrap.rs
@@ -112,7 +112,7 @@ pub(crate) async fn bootstrap(
     for addr in bootstrap_addresses {
         info!("Attempting to dial peer at {addr}");
         if swarm.dial(addr.clone()).is_err() {
-            warn!("Failed to dial peer");
+            warn!("Failed to dial peer at address {addr}");
         }
     }
     swarm
--- a/node/src/service/p2p/bootstrap.rs
+++ b/node/src/service/p2p/bootstrap.rs
@@ -173,7 +173,7 @@ fn on_rendezvous_event(
                 multiaddrs: registration.record.addresses().to_vec(),
             };
             // Serialize PeerInfo
-            let encoded_peer_info = match bincode::serialize(&peer_info) {
+            let encoded_peer_info = match serde_json::to_vec(&peer_info) {
                 Ok(info) => info,
                 Err(..) => {
                     error!(peer_info:?; "Failed to serialize peer_info");
@@ -208,10 +208,12 @@ fn on_gossipsub_event(
             message_id: id,
             message,
         } => {
+            // Got a message from ourselves, return early.
             if peer_id == local_peer_id {
                 return;
             }
-            let peer_info: PeerInfo = match bincode::deserialize(&message.data) {
+            // Deserialize peer info
+            let peer_info: PeerInfo = match serde_json::from_slice(&message.data) {
                 Ok(info) => info,
                 Err(..) => {
                     error!(message:? = message.data; "Received invalid peer info from peer {peer_id:?}");
--- a/node/src/service/p2p/bootstrap.rs
+++ b/node/src/service/p2p/bootstrap.rs
@@ -14,7 +14,7 @@ use libp2p::{
     tcp, yamux, Multiaddr, PeerId, StreamProtocol, Swarm, SwarmBuilder,
 };
 use log::{debug, error, info, warn};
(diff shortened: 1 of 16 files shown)


node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +104,166 @@ impl BootstrapConfig {
...
     mut swarm: Swarm<BootstrapBehaviour>,
     addr: Multiaddr,
+    bootstrap_addresses: Vec<Multiaddr>,
 ) -> Result<(), P2PError> {
     info!("Starting P2P bootstrap node at {addr}");
     swarm.listen_on(addr)?;
+    for addr in bootstrap_addresses {
+        info!("Dialing {addr}");
+        swarm.dial(addr)?;
+    }
```
th7nder: What if one of the addressess fail to dial? I don't think we should fail the entire bootstrap if one of the nodes is unreachable.

node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +104,166 @@ impl BootstrapConfig {
...
+        rendezvous::server::Event::PeerRegistered { peer, registration } => {
+            info!(
+                "Peer {} registered for namespace '{}' for {} seconds",
+                peer, registration.namespace, registration.ttl
+            );
+            let peer_info = PeerInfo {
+                peer_id: peer,
+                multiaddrs: registration.record.addresses().to_vec(),
+            };
+            let encoded_peer_info = match bincode::serialize(&peer_info) {
```
th7nder: Any reason for using bincode instead of SCALE/serde which we already depend on anyways? I mean, we don't need to add an additional dependency in this way.

node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +104,166 @@ impl BootstrapConfig {
...
+            };
+            registrations.insert(peer, peer_info);
+            match swarm
+                .behaviour_mut()
+                .gossipsub
+                .publish(IdentTopic::new(GOSSIP_TOPIC), encoded_peer_info)
+            {
+                Ok(..) => info!("Successfully published new peer info"),
+                Err(..) => error!("Failed to publish new peer info"),
             }
```
th7nder:
```suggestion
                Ok(..) => info!("Successfully published new peer info for peer {}", peer_id),
                Err(..) => error!("Failed to publish new peer info for peer {}", peer_id),
            }
```

node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +104,166 @@ impl BootstrapConfig {
...
+                SwarmEvent::ConnectionEstablished { peer_id, .. } => {
+                    info!("Connected to {}", peer_id);
+                }
+                SwarmEvent::ConnectionClosed { peer_id, .. } => {
+                    info!("Disconnected from {}", peer_id);
+                }
+                SwarmEvent::Behaviour(BootstrapBehaviourEvent::Rendezvous(event)) => on_rendezvous_event(&mut swarm, event, &mut registrations),
+                SwarmEvent::Behaviour(BootstrapBehaviourEvent::Gossipsub(event)) => on_gossipsub_event(event, *swarm.local_peer_id(), &mut registrations),
+                SwarmEvent::Behaviour(BootstrapBehaviourEvent::RequestResponse(event)) => on_request_response_event(&mut swarm, event, &registrations),
+                other => debug!("Encountered event: {other:?}"),
```
th7nder, line 139: That's funny that we just can pass shared exclusive references to those things. 
It means that we're handling everything in the same task.
I guess for know we don't need to introduce the event loop pattern with spawning on messages and synchronization, because all of the message handling is so fast that it won't be blocking?

But then if we're blocking the thread on `RequestResponse`, then a client which is purposefully slow with its responses can DoS the other messages handling.

node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +104,166 @@ impl BootstrapConfig {
...
+                info!("Looking up {id:?}");
+                match registrations.get(&id) {
+                    Some(peer_info) => {
+                        if swarm
+                            .behaviour_mut()
+                            .request_response
+                            .send_response(channel, peer_info.clone())
+                            .is_err()
+                        {
+                            error!("Failed to send peer info to {peer:?}");
```
th7nder: Shouldn't we send a response like: 'sorry bro but I don't have it'?

storage-provider/client/examples/peer-resolver.rs
```diff
@@ -1,133 +1,114 @@
...
+                    if let Message::Response { request_id, response } = message {
+                        tracing::info!("Received response with id {request_id} from {peer}");
+                        return Ok(response);
+                    },
+                    libp2p::request_response::Event::OutboundFailure {
+                        peer,
+                        request_id,
+                        error,
+                    } => tracing::warn!("Failed to send message with id {request_id} to {peer}: {error}"),
+                    libp2p::request_response::Event::InboundFailure {
```
th7nder: When InboundFailure is sent, after a timeout?
What happens after, is it stuck in an infinite loop? When should we kill the process?

th7nder (changes requested):
Overall looks good, I just have some minor concerns about error handling and serialization.

node/src/service/p2p/bootstrap.rs
aidan46: The bootstrap node does not expect a response. It only responds to requests. So, a client being purposefully slow does not really affect the bootstrap node.

storage-provider/client/examples/peer-resolver.rs
aidan46: Killed the process after encountering an error.

node/src/service/p2p/bootstrap.rs
aidan46: I changed the type to be an enum with `Found` and `NotFound` so the client will get a response whether the peer ID is registered or not.

node/src/service/p2p/bootstrap.rs
aidan46: I need `bincode` to serialize/deserialize from/to bytes. `serde_json` (which is imported) does allow to serialize as a JSON byte vector and deserialize from bytes of JSON text but I am not sure if this is the right approach. Wdyt?

th7nder: I mean, honestly, I'd go with JSON text as not to add an additional dependency (if we're not going with SCALE).
Others may have different opinions.

node/src/service/p2p/bootstrap.rs
th7nder: Ok, i'm fine with that then.

node/src/service/p2p/bootstrap.rs
```diff
@@ -37,22 +56,47 @@ impl BootstrapConfig {
...
+
+                Ok(BootstrapBehaviour {
+                    // Rendezvous server behaviour for serving new peers to connecting nodes.
+                    rendezvous: rendezvous::server::Behaviour::new(
+                        rendezvous::server::Config::default()
+                            .with_max_ttl(DEFAULT_REGISTRATION_TTL), // Max TTL of 24 hours
+                    ),
+                    // The identify behaviour is used to share the external address and the public key with connecting clients.
+                    identify: identify::Behaviour::new(identify::Config::new(
+                        "identify/1.0.0".to_string(),
```
th7nder: nit: I can see that rendezvous doesn't have this specified and I'd rather have those as constant somewhere.

node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +105,197 @@ impl BootstrapConfig {
...
     addr: Multiaddr,
+    bootstrap_addresses: Vec<Multiaddr>,
 ) -> Result<(), P2PError> {
     info!("Starting P2P bootstrap node at {addr}");
     swarm.listen_on(addr)?;
+    for addr in bootstrap_addresses {
+        info!("Attempting to dial peer at {addr}");
+        if swarm.dial(addr.clone()).is_err() {
+            warn!("Failed to dial peer");
+        }
```
th7nder:
```suggestion
    for addr in bootstrap_addresses {
        info!("Attempting to dial peer at {addr}");
        if swarm.dial(addr.clone()).is_err() {
            warn!("Failed to dial peer: {}", addr);
        }
```

Btw, what happens if a bootstrap node later joins a network? Will it still be like connected/connect somehow to the bootstrap nodes and still exchange information?

zombienet/local-kube-testnet.toml
```diff
@@ -36,3 +37,18 @@ image = "polkadotstorage.azurecr.io/parachain-node:0.1.0"
 name = "charlie"
 rpc_port = 42069
 validator = true
+
+# run david as parachain collator
+[[parachains.collators]]
+args = [
+  "--bootstrap-addresses=/ip4/127.0.0.1/tcp/62649",
+  "--detailed-log-output",
+  "--p2p-key=@/tmp/zombienet/private.pem",
```
th7nder:
```suggestion
  "--p2p-key=@/tmp/zombienet/charlie-private.pem",
```

th7nder: Why are we adding david to the kube configuration, but not adding it to the `local-testnet.toml`?

th7nder (changes requested):

node/src/service/p2p/bootstrap.rs
aidan46: The connection is established when one of the nodes connect to each other. So if a node is already running and another one dials (connects) the running node the information will still be shared.

Note: This PR does not implement any catch-up or recovery, meaning if a bootstrap node joins late, after registrations have happened, it will have incomplete data.
To keep this PR sizeable I have not implemented this in here but this is certainly an improvement we have to make.

zombienet/local-kube-testnet.toml
aidan46: I forgot to remove david from the kube config. I have created a separate config for david `local-david-collator.toml`. This is kinda as a bandaid because when running it all in 1 config the bootstrap nodes do not seem to connect to each other (not sure why)

th7nder (dismissed):

primitives/src/p2p.rs
```diff
@@ -19,3 +20,59 @@ pub fn keypair_value_parser(src: &str) -> Result<Keypair, String> {
     };
     Keypair::ed25519_from_bytes(key.to_bytes()).map_err(|e| e.to_string())
 }
+
+/// Struct holds peer information.
+/// - PeerId: Registered Peer ID.
+/// - multiaddrs: Vec of multiaddresses the peer has registered.
+#[cfg(feature = "std")]
```
cernicc: p2p module is already behind a feature flag. I think you can remove `#[cfg(feature = "std")]` in this file.

node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +105,197 @@ impl BootstrapConfig {
...
+            );
+            let peer_info = PeerInfo {
+                peer_id: peer,
+                multiaddrs: registration.record.addresses().to_vec(),
+            };
+            // Serialize PeerInfo
+            let encoded_peer_info = match bincode::serialize(&peer_info) {
+                Ok(info) => info,
+                Err(..) => {
+                    error!("Failed to serialize peer_info");
```
cernicc: I think this should never happen. But we should still add the info to the message if printing a log.

```suggestion
                    error!(?peer_info, "Failed to serialize peer_info");
```

node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +105,197 @@ impl BootstrapConfig {
...
+            message,
+        } => {
+            if peer_id == local_peer_id {
+                return;
             }
-            other => debug!("Encountered event: {other:?}"),
+            let peer_info: PeerInfo = match bincode::deserialize(&message.data) {
+                Ok(info) => info,
+                Err(..) => {
+                    error!("Received invalid peer info");
```
cernicc:
```suggestion
                    error!(message = ?message.data, "Received invalid peer info");
```

node/src/service/p2p/bootstrap.rs
```diff
@@ -61,46 +105,197 @@ impl BootstrapConfig {
...
+                    return;
+                }
+            };
+            info!(
+                "Got registration: {:?} with id: {} from peer: {:?}",
+                peer_info, id, peer_id
+            );
+            registrations
```
cernicc: Lets extract this to a function. It can be used in `on_rendezvous_event` too.

node/src/service/p2p/bootstrap.rs
aidan46: It woks a bit different in the [log library](https://docs.rs/log/latest/log/kv/index.html) but I understand your point. Will add the info to the log, same for the de-serialization.

cernicc (approved):

th7nder (approved):

Issue #703, opened 2025-01-24
Transactions sent in parallel don't work


th7nder:
```
    tokio::join!(
        register_storage_provider(client.clone(), charlie_kp.clone()),
        set_porep_verifying_key(
            client.clone(),
            charlie_kp.clone(),
            VerifyingKey::from_raw_bytes(porep_vk_scale),
        ),
        set_post_verifying_key(
            client.clone(),
            charlie_kp.clone(),
            VerifyingKey::from_raw_bytes(post_vk_scale),
        ),
        // Deal collateral + pre_commit_deposit (1)
        add_balance(client.clone(), charlie_kp.clone(), 12_500_000_001),
        add_balance(client.clone(), alice_kp.clone(), 25_000_000_0000),
    );
```

```
2025-01-24T13:45:59.819538Z DEBUG register_storage_provider{address="5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y"}: storagext::runtime::client: Waiting for finalization, Previous nonce: 0, next nonce: 0
2025-01-24T13:46:00.825749Z DEBUG set_porep_verifying_key{address="5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y"}: storagext::runtime::client: Waiting for finalization, Previous nonce: 0, next nonce: 0
2025-01-24T13:46:01.831169Z DEBUG set_post_verifying_key{address="5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y"}: storagext::runtime::client: Waiting for finalization, Previous nonce: 0, next nonce: 0
2025-01-24T13:46:02.835048Z DEBUG add_balance{address="5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y" amount=12500000001}: storagext::runtime::client: Waiting for finalization, Previous nonce: 0, next nonce: 0
2025-01-24T13:46:06.250544Z DEBUG add_balance{address="5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY" amount=250000000000}: storagext::runtime::client: Waiting for finalization, Previous nonce: 0, next nonce: 0
```

It probably has to do with our mechanism of setting nonces.
We're depending on the nonce returned from the:
```
        let current_nonce = self
            .legacy_rpc
[...]

PR #748, opened 2025-02-13
fix: post proof batching

### LINKED ISSUE #738: Accommodate splitting batches in Windowed PoSt

cernicc:
### Description

The issue is that we can only publish 10 proofs in a single extrinsic. 1 proof = 1 partition. So in this pr we changed so that the proofs are generated only for 10 partitions at most. 

This was tested by lowering the limit and checking if the prooving still works.

### Checklist

### DIFF
--- a/storage-provider/common/src/deadline.rs
+++ b/storage-provider/common/src/deadline.rs
@@ -1,9 +1,10 @@
 use std::sync::Arc;
 
+use futures::future::join_all;
 use itertools::Itertools;
 use polka_storage_proofs::{
     match_post_proof,
-    porep::sealer::{BlstrsProof, SubstrateProof},
+    porep::sealer::SubstrateProof,
     post::{self, generate_window_post, PoStParameters, ReplicaInfo},
 };
 use primitives::{
@@ -17,10 +18,7 @@ use storagext::{
     types::storage_provider::{PartitionState, PoStProof, SubmitWindowedPoStParams},
     RandomnessClientExt, StorageProviderClientExt, SystemClientExt,
 };
-use subxt::{
-    ext::{codec::Encode, futures::future::join_all},
-    tx::Signer,
-};
+use subxt::{ext::codec::Encode, tx::Signer};
 use tokio::task::JoinError;
 
 use crate::sector::ProvenSector;
@@ -71,15 +69,15 @@ impl Deadline {
             .ok_or(DeadlineError::DeadlineNotFound(self.deadline_index))
     }
 
-    pub async fn submit_windowed_post<SectorStorage>(
+    pub async fn submit_windowed_post<F>(
         &self,
         xt_client: Arc<storagext::Client>,
         xt_keypair: &storagext::multipair::MultiPairSigner,
         post_params: Arc<PoStParameters>,
-        sector_storage: SectorStorage,
+        sector_storage: F,
     ) -> Result<(), DeadlineError>
     where
-        SectorStorage: Fn(SectorNumber) -> Option<ProvenSector>,
+        F: Fn(SectorNumber) -> Option<ProvenSector>,
     {
         let deadline = self.get_info(xt_client.clone(), xt_keypair).await?;
 
@@ -184,7 +182,7 @@ impl Deadline {
         Ok(())
     }
 
-    async fn generate_and_submit<SectorStorage>(
+    async fn generate_and_submit<F>(
         &self,
         xt_client: Arc<storagext::Client>,
         xt_keypair: &storagext::multipair::MultiPairSigner,
@@ -193,10 +191,10 @@ impl Deadline {
         prover_id: [u8; 32],
         randomness: [u8; 32],
         post_params: Arc<PoStParameters>,
-        sector_storage: SectorStorage,
+        sector_storage: F,
     ) -> Result<(), DeadlineError>
     where
-        SectorStorage: Fn(SectorNumber) -> Option<ProvenSector>,
+        F: Fn(SectorNumber) -> Option<ProvenSector>,
     {
         // Generate proofs for partitions
         let (partitions, proofs) = self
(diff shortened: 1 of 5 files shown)


storage-provider/common/src/deadline.rs
```diff
@@ -158,8 +140,96 @@ impl Deadline {
...
+                let partitions = partitions.collect();
+                tracing::info!("Proving PoSt partitions... {:?}", partitions);
+
+                let post_params = Arc::clone(&post_params);
+                self.generate_proofs_for_partitions(
+                    partitions,
+                    prover_id,
+                    randomness,
+                    post_params,
+                    &sector_storage,
```
th7nder: I'm not sure whether this is the right approach for parallelizing this work.
The `generate_window_post` function is heavily dependent on the GPU, so basically only 1 at a time can run, because each windowed post has an exclusive lock on the GPU.
This means that basically all of the chunks will run sequentially.

Then, we're running `submit_windowed_post` extrinsics for those partitions sequentially. 
Each submit windowed post takes like ~30secs at least (because of the finalization).

I mean, given that we have deadline of length 30 minutes in the prod (currently, it might change), that's not so bad.
However, we couldn't submit them in parallel yet #703. 
Besides this lil' parallel caveat that needs to be fixed (or rechecked, I added a small comment to that).

I think it should the future we're awaiting is more like `generate_window_post -> submit_window_post` in parallel.

th7nder (changes requested):
I agree with the solution for batching, I don't agree on the parallelization yet.

storage-provider/common/src/deadline.rs
cernicc: I've changed it up a little.

storage-provider/server/src/pipeline/mod.rs
```diff
@@ -475,7 +475,7 @@ async fn submit_windowed_post(
     {
         tracing::error!("failed to submit post for deadline, {}", e);
     } else {
-        tracing::info!("submitted post successfully");
+        tracing::info!("completed post submition");
```
th7nder:
```suggestion
        tracing::info!("completed post submission");
```

th7nder (dismissed):

th7nder (dismissed):

storage-provider/common/src/deadline.rs
```diff
@@ -217,4 +237,70 @@ impl Deadline {
...
+                        post_proof,
+                        &post_params,
+                        randomness,
+                        prover_id,
+                        replicas
+                    )
+                )
+            })
+        }
+        .await??;
```
jmg-duarte: Nit: IMO, it's less weird like this
```suggestion
            tokio::task::spawn_blocking(move || {
                match_post_proof!(
                    post_proof,
                    generate_window_post::<_>(
                        post_proof,
                        &post_params,
                        randomness,
                        prover_id,
                        replicas
                    )
                )
            })
            .await??
        };
```

storage-provider/common/src/deadline.rs
```diff
@@ -217,4 +237,70 @@ impl Deadline {
...
+
+        // Map proofs to our internal type
+        let proofs = proofs
+            .into_iter()
+            .map(|p| PoStProof {
+                post_proof: self.post_proof,
+                proof_bytes: codec::Encode::encode(
+                    &TryInto::<SubstrateProof>::try_into(p.clone()).expect(
+                        "converstion between rust-fil-proofs and polka-storage-proofs to work",
```
jmg-duarte: The expect doesn't show up in the error message so the text feels kinda weird without it
```suggestion
                        "converstion between rust-fil-proofs and polka-storage-proofs should work",
```

storage-provider/common/src/deadline.rs
```diff
@@ -217,4 +237,70 @@ impl Deadline {
...
+    async fn generate_proofs_for_partitions<SectorStorage>(
+        &self,
+        partitions: Vec<(PartitionNumber, PartitionState)>,
+        prover_id: [u8; 32],
+        randomness: [u8; 32],
+        post_params: Arc<PoStParameters>,
+        sector_storage: SectorStorage,
+    ) -> Result<(Vec<PartitionNumber>, Vec<PoStProof>), DeadlineError>
+    where
+        SectorStorage: Fn(SectorNumber) -> Option<ProvenSector>,
```
jmg-duarte: This generic name is kinda weird, why `SectorStorage` when it's not storing anything?

storage-provider/common/src/deadline.rs
```diff
@@ -9,14 +10,18 @@ use primitives::{
...
 };
 use storagext::{
     runtime::runtime_types::primitives::pallets::DeadlineInfo,
     types::storage_provider::{PartitionState, PoStProof, SubmitWindowedPoStParams},
     RandomnessClientExt, StorageProviderClientExt, SystemClientExt,
 };
-use subxt::{ext::codec::Encode, tx::Signer};
-use tokio::task::{JoinError, JoinHandle};
+use subxt::{
+    ext::{codec::Encode, futures::future::join_all},
```
jmg-duarte: This looks weird, why not use the proper crates?

storage-provider/common/src/deadline.rs
```diff
@@ -158,42 +140,80 @@ impl Deadline {
...
+                    randomness,
+                    post_params,
+                    &sector_storage,
+                )
             })
             .collect::<Vec<_>>();
 
-        tracing::info!("Wait for block {} for open deadline", deadline.start,);
+        // Proving results
+        let proving_results = join_all(proving_futures).await;
```
jmg-duarte, line 173: Following Konrad's previous comment, how do we know that join_all runs them one by one and doesn't content on the GPU lock?

jmg-duarte (dismissed):
Some nits and questions, but overall looks good

storage-provider/common/src/deadline.rs
cernicc: The suggestion doesn't work. It would work if the await is moved inside the scope. But the `??` needs to be left outside. That is also weird 😄 I've moved the `await??` below. I think that is the least weirdest approach.

storage-provider/common/src/deadline.rs
cernicc: It is weird. Good catch. Auto import

storage-provider/common/src/deadline.rs
cernicc: Not sure why it was named that. I've copied the signature from the previous function.
I've changed the type name to F. It should be better now.

storage-provider/common/src/deadline.rs
cernicc: Before the Konrads comment. I've first generated all proofs and then submitted those proofs to the chain sequentially. 
Konrad's suggestion was, that we should start on the next proof generation while waiting for the submission of the previous proof.

In case of the CPU the proofs are generated in parallel. In case of the GPU, if I understood Konrad correctly, the lock is being hold while the single proof is being generated. I haven't tried the GPU case.

jmg-duarte (approved):

th7nder (approved):

PR #749, opened 2025-02-13
feat(sp-server): start_block based sealing

jmg-duarte:
### Description

Takes into account the start block of a deal when considering the pre-commit.

The implementation is as simple as I could make it given the constant tug-of-war of several requirements, but basically:
* We now always try to schedule a pre-commit for a given sector
* If no pre-commit has been scheduled yet, we schedule it
* If a pre-commit has been scheduled, then we check if the *estimated* target deadline for that task
  * If the deadline for the new task happens *before* the old one, we cancel the old task and replace it in our mapping
  * If the deadline for the new task happens *after* the old one, we don't do anything

The original implementation used abort handles but turns out, they don't work like I expected — I expected that if a task was suspended at an await point and got cancelled, it would be dropped from the runtime and forgotten but no, instead the task will still run and only at the next await point it will be cancelled. 

The new implementation simply suspends waiting for a cancellation value with a timeout (instead of a sleep), then we "race" the cancel versus the timeout, whichever runs first decides whether the pre commit happens or not.

### Checklist

Logs for review:
[sp_server.log.2025-02-13.txt](https://github.com/user-attachments/files/18786282/sp_server.log.2025-02-13.txt)

### DIFF
--- a/storage-provider/common/src/config/sealing.rs
+++ b/storage-provider/common/src/config/sealing.rs
@@ -117,8 +117,8 @@ pub struct SealingConfiguration {
     #[cfg_attr(feature = "clap", arg(long, default_value = "6h", value_parser = duration_value_parser))]
     pub wait_deals_delay: Duration,
 
-    /// The amount of time before a sector's earliest deal expiration; once hit, the sector is
-    /// sealed & pre-committed.
+    /// The amount of time before a sector's earliest deal start; once hit, the sector is sealed &
+    /// pre-committed.
     #[serde(
         default = "default_pre_commit_submission_slack",
         deserialize_with = "duration_deserializer"
--- a/storage-provider/common/src/config/sealing.rs
+++ b/storage-provider/common/src/config/sealing.rs
@@ -14,13 +14,17 @@ use serde::{Deserialize, Deserializer, Serialize};
 
 /// Default amount of time to wait to seal an unfilled sector.
 const fn default_wait_deals_delay() -> Duration {
-    // 24 hours * 60 minutes * 60 seconds =
-    Duration::from_secs(24 * 60 * 60)
+    Duration::from_secs(6 * 60 * 60)
 }
 
 /// Returns the default fill percentage.
 const fn default_fill_threshold() -> u8 {
     95
+}
+
+/// Default amount of time before the expiration of the sectors earliest deal.
+const fn default_pre_commit_submission_slack() -> Duration {
+    Duration::from_secs(60 * 60)
 }
 
 fn validate_percentage(percentage: u8) -> Result<u8, String> {
@@ -105,14 +109,22 @@ pub struct SealingConfiguration {
     ))]
     pub fill_threshold: u8,
 
-    /// The amount of time to wait before sealing an unfilled sector, defaults to 24 hours.
-    /// If the sector is empty, it will not be sealed.
+    /// The amount of time to wait before sealing an unfilled sector, defaults to 6 hours.
     #[serde(
         default = "default_wait_deals_delay",
         deserialize_with = "duration_deserializer"
     )]
-    #[cfg_attr(feature = "clap", arg(long, default_value = "24h", value_parser = duration_value_parser))]
+    #[cfg_attr(feature = "clap", arg(long, default_value = "6h", value_parser = duration_value_parser))]
     pub wait_deals_delay: Duration,
+
+    /// The amount of time before a sector's earliest deal expiration; once hit, the sector is
+    /// sealed & pre-committed.
+    #[serde(
+        default = "default_pre_commit_submission_slack",
+        deserialize_with = "duration_deserializer"
+    )]
+    #[cfg_attr(feature = "clap", arg(long, default_value = "1h", value_parser = duration_value_parser))]
+    pub pre_commit_submission_slack: Duration,
 }
 
 // This default is implemented for when `sealing_configuration` is missing from the config file,
@@ -124,6 +136,7 @@ impl Default for SealingConfiguration {
         Self {
             fill_threshold: default_fill_threshold(),
             wait_deals_delay: default_wait_deals_delay(),
+            pre_commit_submission_slack: default_pre_commit_submission_slack(),
         }
     }
 }
--- a/storage-provider/server/src/pipeline/add_piece.rs
+++ b/storage-provider/server/src/pipeline/add_piece.rs
@@ -9,7 +9,7 @@ use primitives::{
     sector::SectorNumber,
 };
 use storagext::{types::market::DealProposal, SystemClientExt};
-use tokio::task::JoinHandle;
+use tokio::sync::oneshot;
 use tokio_util::task::TaskTracker;
 use tracing::Instrument;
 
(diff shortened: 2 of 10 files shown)


storage-provider/server/src/pipeline/mod.rs
```diff
@@ -436,6 +293,15 @@ async fn precommit(
 ) -> Result<(), PipelineError> {
     tracing::info!("Starting pre-commit");
 
+    {
+        // We remove ourselves from the scheduled pre-commits
+        let mut scheduled_pre_commits = state.scheduled_pre_commits.lock().await;
+        if scheduled_pre_commits.remove(&sector_number).is_none() {
+            tracing::warn!(%sector_number, "No task was found!");
+            // Maybe we should do an early return in this case?
```
aidan46: I think an early return is appropriate here since the pre commit is not found in the scheduled ones.

storage-provider/server/src/pipeline/add_piece.rs
```diff
@@ -0,0 +1,248 @@
...
+use tokio_util::task::TaskTracker;
+use tracing::Instrument;
+
+use crate::{
+    pipeline::{PipelineError, PipelineState},
+    rpc::SECS_PER_BLOCK,
+};
+
+/// Finds a sector to which a piece will fit and adds it to the sector.
+/// This function is *cancellation safe* as if future is dropped,
```
th7nder: I'd argue whether this is cancellation safe. 
If it's dropped at any await point after `state.db.insert_unsealed_sector(sector_number, &sector)?;` then it never ends up being scheduled.

storage-provider/server/src/pipeline/add_piece.rs
```diff
@@ -0,0 +1,248 @@
...
+    if fill_percentage >= fill_threshold {
+        tracing::debug!(
+            %sector_number,
+            "Occupation level at {}%, above limit of {}% - pre-committing",
+            fill_percentage,
+            fill_threshold,
+        );
+        // TODO(@th7nder,30/10/2024): simplification, as we're always scheduling a precommit just after adding a piece and creating a new sector.
+        // Ideally sector won't be finalized after one piece has been added and the precommit will depend on the start_block?
+        return state.send_pre_commit(sector_number);
```
th7nder: This todo no longer applies, I think.

storage-provider/common/src/config/sealing.rs
```diff
@@ -105,14 +109,22 @@ pub struct SealingConfiguration {
...
+    /// The amount of time to wait before sealing an unfilled sector, defaults to 6 hours.
     #[serde(
         default = "default_wait_deals_delay",
         deserialize_with = "duration_deserializer"
     )]
-    #[cfg_attr(feature = "clap", arg(long, default_value = "24h", value_parser = duration_value_parser))]
+    #[cfg_attr(feature = "clap", arg(long, default_value = "6h", value_parser = duration_value_parser))]
     pub wait_deals_delay: Duration,
+
+    /// The amount of time before a sector's earliest deal expiration; once hit, the sector is
```
th7nder: Earliest deal start, not expiration - I suppose.

storage-provider/server/src/pipeline/add_piece.rs
```diff
@@ -0,0 +1,248 @@
...
+
+            tracing::info!(%sector_number, "Awoken from sleep, submitting pre-commit task!");
+            match state.send_pre_commit(sector_number) {
+                Ok(()) => tracing::info!(%sector_number, "Successfully submitted task!"),
+                // Not sure what we can do if the task fails, maybe try to re-submit?
+                // Not even sure when this can happen asides from an OOM since we're using an unbounded channel
+                Err(err) => tracing::error!(%sector_number, "Failed to submit task: {}", err),
+            }
+        }
+        .instrument(span),
```
th7nder, line 245:
```suggestion
    let _ = tracker.spawn(
        async move {
            tokio::select! {
                _ = tokio::time::sleep(when) => {
                    tracing::debug!(%sector_number, "No cancellation signal was received");
                    tracing::info!(%sector_number, "Awoken from sleep, submitting pre-commit task!");
                    match state.send_pre_commit(sector_number) {
                        Ok(()) => tracing::info!(%sector_number, "Successfully submitted task!"),
                        Err(err) => tracing::error!(%sector_number, "Failed to submit task: {}", err),
                    }
                }
                _ = cancellation_receiver => {
                    tracing::debug!(%sector_number, "Received cancellation signal, not sending message.");
                }
            }
        }
        .instrument(span),
```

Have you considered doing it like this?

storage-provider/server/src/pipeline/mod.rs
```diff
@@ -79,6 +80,18 @@ pub struct PipelineState {
     // transactions + get_for_update(exclusive: true)
     // This is not trivial and requires a refactor on the DB side!
     pub add_piece_serializer: Mutex<()>,
+
+    // Store the estimated date of execution of the task and its abort handle for re-scheduling
+    pub scheduled_pre_commits: Mutex<HashMap<SectorNumber, (DateTime<Utc>, oneshot::Sender<()>)>>,
```
th7nder, line 85: I'm wondering why the [`CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) wasn't used here instead of `oneshot::Sender<()>`.

storage-provider/server/src/pipeline/mod.rs
```diff
@@ -436,6 +293,15 @@ async fn precommit(
 ) -> Result<(), PipelineError> {
     tracing::info!("Starting pre-commit");
 
+    {
+        // We remove ourselves from the scheduled pre-commits
+        let mut scheduled_pre_commits = state.scheduled_pre_commits.lock().await;
+        if scheduled_pre_commits.remove(&sector_number).is_none() {
+            tracing::warn!(%sector_number, "No task was found! Not pre-committing.");
```
th7nder:
```suggestion
            tracing::warn!(%sector_number, "No task was found! Skipping pre-committing as it must have been precommited before");
```

storage-provider/server/src/pipeline/add_piece.rs
```diff
@@ -0,0 +1,248 @@
...
+                tracing::error!("Failed to send cancellation value");
+            }
+
+            tracing::debug!(%sector_number, "Existing deadline & task have been replaced: new_deadline = {}", deadline);
+        }
+        Some((old_deadline, _)) => {
+            tracing::debug!(%sector_number, "Existing deadline is newer than the new one: old_deadline = {}, new_deadline = {}", old_deadline, deadline);
+        }
+        None => {
+            tracing::debug!(%sector_number, "No deadline existing existed, creating it.");
```
th7nder:
```suggestion
            tracing::debug!(%sector_number, "Precommit wasn't scheduled, scheduling it.");
```

storage-provider/server/src/pipeline/add_piece.rs
```diff
@@ -0,0 +1,248 @@
...
+/// that task will be cancelled and replaced with the new one.
+/// Otherwise, this function is a no-op.
+#[tracing::instrument(skip_all)]
+async fn schedule_pre_commit(
+    state: Arc<PipelineState>,
+    tracker: TaskTracker,
+    sector_number: SectorNumber,
+    when: Duration,
+) {
+    let deadline = Utc::now() + when;
```
th7nder:
```suggestion
    let precommit_target_time = Utc::now() + when;
```

I just don't like `deadline`, it's confusing me with those deadlines in the PoSt.

th7nder (changes requested):
I can see this working, can't come up with something better given the cards we have. Left some code suggestions.

Internally I was thinking if maybe we could get rid of this `start_block` from the protocol totally, but then there is a question when should we expect to submit a proof for the file that was uploaded and it might break a million of other stuff. So just leaving this as 'food for thought'.

storage-provider/server/src/pipeline/add_piece.rs
```diff
@@ -0,0 +1,248 @@
...
+    tracing::debug!(
+        "Searching for sector for piece with size: {}",
+        deal.piece_size
+    );
+    // Find the first sector with enough space for the piece
+    let sector = state.db.iter_unsealed_sectors().find(|res| match res {
+        Ok(unsealed_sector) => {
+            let free_space = unsealed_sector.free_space();
+            tracing::debug!(sector_number = %unsealed_sector.sector_number, free_space = free_space, "Checking sector...");
+            free_space > deal.piece_size
```
cernicc: Can we use `>=`?

storage-provider/server/src/pipeline/add_piece.rs
```diff
@@ -0,0 +1,248 @@
...
+            .inspect(|sector| {
+                tracing::debug!(
+                    sector_number = %sector.sector_number,
+                    "Found sector for piece!",
+                );
+            })
+            .map_err(PipelineError::from);
+    }
+
+    // NOTE(@jmg-duarte,03/02/2025): comment below no longer applies but im keeping it until
```
cernicc, line 129: Is the NOTE and TODO still relevant?

storage-provider/server/src/pipeline/mod.rs
jmg-duarte: Because the CancellationToken allows to create a tree of tokens, enabling misuse. This is my attempt at reducing said misuse

storage-provider/server/src/pipeline/add_piece.rs
jmg-duarte: Yes, good catch

storage-provider/server/src/pipeline/add_piece.rs
jmg-duarte: You're right, it leaves the DB in an inconsistent state. Without a bigger refactor on the DB its really hard to ensure this is cancellation safe.

storage-provider/server/src/pipeline/add_piece.rs
jmg-duarte: Kinda, we still create sectors non-stop but I'm not sure what algorithm to implement

storage-provider/server/src/pipeline/mod.rs
th7nder: Okk

th7nder (approved):

cernicc (approved):

PR #752, opened 2025-02-14
fix: support multiple partitions in PoRep

### LINKED ISSUE #747: Adjust proofs in PoRep for bigger sector sizes

th7nder:
### Description

Fixes #747.
When a PoRep is configured with multiple partitions (meaning many more challenges to prove) we didn't support this case.
It's still untested, as we don't have configuration with those multiple partitions.
It'll be followed-up in #751.

### Checklist

  - [x] If yes, which ones?
  - [ ] If yes, which ones? / List them here
     - [x] #751  


pallets/storage-provider/src/lib.rs
```diff
@@ -1634,14 +1634,19 @@ pub mod pallet {
...
+        >,
     ) -> Result<(), DispatchError> {
         let max_proof_size = precommit.info.seal_proof.proof_size();
 
         // Check proof size
-        if proof.len() > max_proof_size {
-            log::error!(target: LOG_TARGET, "sector proof size {} exceeds max {}", proof.len(), max_proof_size);
-            return Err(Error::<T>::InvalidProof)?;
+        for proof in proofs.iter() {
+            if proof.len() > max_proof_size {
```
jmg-duarte:
```suggestion
			if proofs.iter().any(|p| p.len() > max_proof_size)  {
```

jmg-duarte (dismissed):
lgtm

pallets/market/src/benchmarking.rs
```diff
@@ -343,13 +343,16 @@ mod benchmarks {
...
+        proofs.try_push(BoundedVec::new()).unwrap();
+
         // Similar pattern to before
         (0..n)
             // Create ProveCommitSectors from the `n` "index"
             .map(|n| ProveCommitSector {
                 sector_number: n.try_into().unwrap(),
-                // Empty proof is considered valid under the DummyProofVerifier
-                proof: BoundedVec::new(),
+                proofs: proofs.clone(),
```
jmg-duarte: I'm not a fan of this new style because proofs is mut and is getting cloned inside the map, looks kinda weird.

I feel that we could just use the mut inside the map closure, makes it more self contained and the intent clearer

jmg-duarte (dismissed):

pallets/market/src/benchmarking.rs
th7nder: I wanted to just use `bounded_vec![bounded_vec![]]` however they're not available in the runtime benchmarks :(

pallets/storage-provider/src/lib.rs
```diff
@@ -1634,13 +1634,16 @@ pub mod pallet {
...
+            ConstU32<MAX_PROOFS_PER_BLOCK>,
+        >,
     ) -> Result<(), DispatchError> {
         let max_proof_size = precommit.info.seal_proof.proof_size();
 
         // Check proof size
-        if proof.len() > max_proof_size {
-            log::error!(target: LOG_TARGET, "sector proof size {} exceeds max {}", proof.len(), max_proof_size);
+        if proofs.iter().any(|p| p.len() > max_proof_size) {
+            log::error!(target: LOG_TARGET, "sector proof size {} exceeds max {}", proofs.len(), max_proof_size);
```
aidan46: You are checking the length of the inner vector with a single proof but logging that the size of the aggregated proof is too large. Or am I reading this incorrectly? You could check if [find](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) returns `Some(proof)` and log the length of the inner proof.

jmg-duarte: Great catch & fix

aidan46 (approved):

jmg-duarte (approved):

PR #753, opened 2025-02-14
feat(mater): tree-aware content extraction

### LINKED ISSUE #669: feat: Mater content extraction from unixfs dag
Extend our content extraction by enabling `mater` to also be able to extract content from `.car` files that store unixfs dags.

The current extraction only works in cases when the car file contains ordered blocks with arbitrary content. The blocks are expected to be in the correct order. The extraction then iterates over all blocks in sequence and extracts that raw content. This works great for many cases, but in our case we'll have content wrapped in an unixfs dag, the extraction process should know how to walk that graph and extract the content.

jmg-duarte:
### Description

Fixes the content extraction to be tree aware, it's not the most efficient implementation but it's simple & straightforward.
I removed the CLI error enum because it was getting in the way of debugging and it wasn't covering anything that `mater::Error` wasn't already.

I'm looking for better names too.

Closes https://github.com/eigerco/polka-storage/issues/669

### Checklist


mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,143 @@
...
+            .ok_or(Error::WrongNumberOfRoots)
+    }
+
+    /// Indexes the file.
+    ///
+    /// *Always* seeks to the start of the file before proceeding with indexing.
+    /// This function performs indexing *naively*, it will sequentially read all block "headers"
+    /// (Cid, data start offset and data length) without loading the blocks into memory.
+    async fn index(&mut self) -> Result<(), Error> {
+        // Indexing must alwasy be made from the start
```
cernicc:
```suggestion
        // Indexing must always be made from the start
```

mater/lib/src/v1/reader.rs
```diff
@@ -60,6 +60,8 @@ pub(crate) async fn read_block_metadata<R>(mut reader: R) -> Result<BlockMetadat
 where
     R: AsyncRead + AsyncSeek + Unpin,
 {
+    let block_offset = reader.stream_position().await?;
```
cernicc, line 63: Lets calculate the `data_offset_source` from the `block_offset`, `full_block_length_bytes_read` and `cid_bytes_read`. That way we eliminate one syscall

cernicc (commented):
Let him cook

mater/cli/src/extract.rs
```diff
@@ -14,25 +11,18 @@ pub(crate) async fn extract_file_from_car(
...
-
-    let mut reader = CarV2Reader::new(BufReader::new(source_file));
-    reader.extract_content(&mut output_file).await?;
-
-    output_file.flush().await?;
+    let mut loader = FileLoader::from_path(input_path)
+        .await
+        .map_err(|_| Error::InvalidCarFile)?;
 
+    let root = loader.root().await?;
```
aidan46: Why not move this into the `copy_cid` function? Is the `root` function used anywhere else?

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,212 @@
...
+        loader.index().await?;
+        Ok(loader)
+    }
+}
+
+impl FileLoader<Cursor<Vec<u8>>> {
+    /// Creates a [`FileLoader`] from a vector of bytes.
+    pub async fn from_vec(vec: Vec<u8>) -> Result<Self, Error> {
+        let mut loader = Self {
+            reader: v2::Reader::new(Cursor::new(vec)),
```
aidan46, line 49:
```suggestion
    pub async fn from_bytes(bytes: Vec<u8>) -> Result<Self, Error> {
        let mut loader = Self {
            reader: v2::Reader::new(Cursor::new(bytes)),
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,212 @@
...
+        loader.index().await?;
+        Ok(loader)
+    }
+}
+
+impl<R> FileLoader<R>
+where
+    R: AsyncRead + AsyncSeek + Unpin,
+{
+    // pub async fn from_buffer(buffer: Vec<u8>) -> Result<>
```
aidan46: Left over comment

mater/lib/src/file_reader.rs
jmg-duarte: I don't want to say from bytes because there's [`bytes::Bytes`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html) and it could also mean `&[u8]`

mater/cli/src/extract.rs
jmg-duarte: Not really, but I'm not sure if this is going to be helpful for other functions, that's why its here. But I have a compromise

mater/lib/src/file_reader.rs
aidan46: I agree with that statement, you could implement for `Cursor<T>` (since [Cursor::new](https://doc.rust-lang.org/std/io/struct.Cursor.html#method.new) has no trait bounds) or `Cursor<T: AsRef<[u8]>>` (this could work for `Vec<u8>` and `Bytes`) but keeping it like this is fine too.

jmg-duarte: Your suggestion would hide a clone, this one forces the clone on the caller, which is more explicit IMO

mater/cli/src/extract.rs
jmg-duarte: Take a look again 👁️ 👁️

aidan46 (dismissed):

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+        self.reader.read_pragma().await?;
+        self.reader.read_header().await?;
+        // We assume there's a single root
+        self.reader
+            .read_v1_header()
+            .await?
+            .roots
+            .first()
+            .copied()
+            .ok_or(Error::WrongNumberOfRoots)
```
th7nder: I think it'll return error only if roots is == 0, 
if roots > 1 it'll still work.

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /// Indexes the file.
+    ///
+    /// *Always* seeks to the start of the file before proceeding with indexing.
+    /// This function performs indexing *naively*, it will sequentially read all block "headers"
+    /// (Cid, data start offset and data length) without loading the blocks into memory.
+    async fn index(&mut self) -> Result<(), Error> {
```
th7nder: nit: judging by private field `index` and this function name, I'd say that this function returns private field `index`.
```suggestion
    async fn create_index(&mut self) -> Result<(), Error> {
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /// Traverse the tree under the given [`Cid`], yielding the data in the leaves.
+    fn load_cid<'a>(
+        &'a mut self,
+        cid: &'a Cid,
+    ) -> impl Stream<Item = Result<(Cid, Vec<u8>), Error>> + 'a {
+        try_stream! {
+            let partial_node = self.index.get(&cid).ok_or(Error::InvalidCid)?;
+            let mut stack = VecDeque::new();
```
th7nder: this is not a stack, it's a queue. 
You're pushing back and popping from front.
This is BFS.
```suggestion
            let mut queue = VecDeque::new();
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+use ipld_core::{cid::Cid, codec::Codec};
+use ipld_dagpb::{DagPbCodec, PbNode};
+use tokio::{
+    fs::File,
+    io::{AsyncRead, AsyncSeek, AsyncSeekExt, AsyncWriteExt},
+};
+
+use crate::{multicodec, v1::BlockMetadata, v2, Error};
+
+/// CAR file loader.
```
th7nder:
```suggestion
/// Extracts the raw data from a CARv2 file.
/// It expects the CAR file to have only 1 root.
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+            let block_metadata = self.reader.read_block_metadata().await?;
+            self.index
+                .insert(block_metadata.cid, block_metadata.try_into()?);
+        }
+
+        Ok(())
+    }
+
+    /// Traverse the tree under the given [`Cid`], yielding the data in the leaves.
+    fn load_cid<'a>(
```
th7nder:
```suggestion
    fn tree_stream<'a>(
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+    ///
+    /// This is equivalent to reading the stream of blocks from [`Self::load_cid`] into a writer.
+    async fn copy_cid<W>(&mut self, cid: &Cid, mut w: W) -> Result<(), Error>
+    where
+        W: AsyncWriteExt + Unpin,
+    {
+        let loader = self.load_cid(cid);
+        tokio::pin!(loader);
+        while let Some((cid, block)) = loader.try_next().await? {
+            println!("{}", cid);
```
th7nder:
```suggestion
            tracing::debug!(LOG_TARGET: "mater", "loading block for cid {}", cid);
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+                .insert(block_metadata.cid, block_metadata.try_into()?);
+        }
+
+        Ok(())
+    }
+
+    /// Traverse the tree under the given [`Cid`], yielding the data in the leaves.
+    fn load_cid<'a>(
+        &'a mut self,
+        cid: &'a Cid,
```
th7nder, line 108: Just to be clear, because we're supporting only 1 Cid we could've skipped that, but this prepares the implementation for the multi-roots?

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    /// Writes the content tree for the given [`Cid`] into `w`.
+    ///
+    /// This is equivalent to reading the stream of blocks from [`Self::load_cid`] into a writer.
+    async fn copy_cid<W>(&mut self, cid: &Cid, mut w: W) -> Result<(), Error>
```
th7nder:
```suggestion
    async fn copy_tree<W>(&mut self, cid: &Cid, mut w: W) -> Result<(), Error>
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+impl TryFrom<BlockMetadata> for PartialNode {
+    type Error = Error;
+
+    fn try_from(value: BlockMetadata) -> Result<Self, Self::Error> {
+        match value.cid.codec() {
+            multicodec::RAW_CODE => Ok(Self::Leaf(value)),
+            multicodec::DAG_PB_CODE => Ok(Self::Stem(value)),
+            _ => Err(Error::InvalidCid),
```
th7nder:
```suggestion
            unknown_codec => Err(Error::InvalidCid(unknown_codec),
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// Ensures that duplicated blocks
+    #[tokio::test]
+    async fn read_duplicated_blocks() {
+        let raw_input = std::iter::repeat(0).take(4096).collect::<Vec<u8>>();
+
+        let mut bs = Blockstore::with_parameters(Some(1024), None);
+        bs.read(Cursor::new(raw_input.clone())).await.unwrap();
+
+        // 1519 is the expected CAR file size after deduplicating and writing the CAR file
+        let mut out_car_buffer = Vec::with_capacity(1519);
```
th7nder, line 215: how did you get this number?

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+        assert_eq!(raw_input, out_check.into_inner());
+    }
+
+    #[tokio::test]
+    async fn read_wrapped() {
+        let spaceglenda_original = tokio::fs::read("tests/fixtures/original/spaceglenda.jpg")
+            .await
+            .unwrap();
+        let mut spaceglenda_wrapped_loader =
+            FileLoader::from_path("tests/fixtures/car_v2/spaceglenda_wrapped.car")
```
th7nder: How was this file created?

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+                    PartialNode::Stem(metadata) => {
+                        self.reader
+                            .get_inner_mut()
+                            .seek(std::io::SeekFrom::Start(metadata.block_offset))
+                            .await?;
+                        let (_, block) = self.reader.read_block().await?;
+
+                        let pb_node: PbNode = DagPbCodec::decode_from_slice(block.as_slice())?;
+                        for link in pb_node.links {
+                            let partial_node = self.index.get(&link.cid).ok_or(Error::InvalidCid)?;
```
th7nder: This error should not be the same as it's also used in `TryFrom` below.

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+impl FileLoader<Cursor<Vec<u8>>> {
+    /// Creates a [`FileLoader`] from a vector of bytes.
+    pub async fn from_vec(vec: Vec<u8>) -> Result<Self, Error> {
+        let mut loader = Self {
+            reader: v2::Reader::new(Cursor::new(vec)),
+            index: HashMap::new(),
+        };
+        loader.index().await?;
```
th7nder: What if the file is already indexed?

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,246 @@
...
+use ipld_dagpb::{DagPbCodec, PbNode};
+use tokio::{
+    fs::File,
+    io::{AsyncRead, AsyncSeek, AsyncSeekExt, AsyncWriteExt},
+};
+
+use crate::{multicodec, v1::BlockMetadata, v2, Error};
+
+/// CAR file loader.
+pub struct FileLoader<R>
```
th7nder:
```suggestion
pub struct CarExtractor<R>
```

th7nder (changes requested):

mater/lib/src/file_reader.rs
jmg-duarte: Re-indexed, there's no caching + it's naive (meaning it ignores the file index) and this is the constructor after all, so it assumes "nothing was here before"

mater/lib/src/file_reader.rs
jmg-duarte: Great catch, I'll make this method work differently and move the error somewhere else to ease up on expanding the struct

mater/lib/src/file_reader.rs
jmg-duarte: This reads the tree at whatever level you want as long as it has been indexed. So it's not only about multi roots, but also partial file reads

mater/lib/src/file_reader.rs
jmg-duarte: Great catch, this was a stack until I realized my mistake in using said stack

mater/lib/src/file_reader.rs
jmg-duarte: Great catch, this just needs to leave, no point in adding tracing

mater/lib/src/file_reader.rs
jmg-duarte: I built the file on the side, it's just a mini optimization since we know the target size

mater/lib/src/file_reader.rs
jmg-duarte: With go-car

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,247 @@
...
+            reader: v2::Reader::new(file),
+            index: HashMap::new(),
+        };
+        loader.naive_build_index().await?;
+        Ok(loader)
+    }
+}
+
+impl CarExtractor<Cursor<Vec<u8>>> {
+    /// Creates a [`FileLoader`] from a vector of bytes.
```
th7nder:
```suggestion
    /// Creates a [`CarExtractor`] from a vector of bytes.
```

mater/lib/src/file_reader.rs
```diff
@@ -0,0 +1,247 @@
...
+pub struct CarExtractor<R>
+where
+    R: AsyncRead + AsyncSeek + Unpin,
+{
+    reader: v2::Reader<R>,
+    index: HashMap<Cid, PartialNode>,
+}
+
+impl CarExtractor<File> {
+    /// Creates a [`FileLoader`] from the given file path.
```
th7nder:
```suggestion
    /// Creates a [`CarExtractor`] from the given file path.
```

th7nder (dismissed):

th7nder (dismissed):

cernicc (dismissed):

th7nder (approved):

cernicc (approved):

Issue #757, opened 2025-02-17
Improve fetch extract UX


jmg-duarte:
Currently polka fetch downloads a CAR file, we can improve the UX on the user end by leveraging a bit of the UNIX philosophy:

```
polka-fetch ... | mater-cli extract -
```

PR #761, opened 2025-02-17
feat(mater-cli): add support for stdin input

jmg-duarte:
### Description

Partial fix for #757, will need fetch to stream the file to stdout.
Also "brings to light" the issue that is if the file needs to be fully loaded into memory.

Test it with 
```
# extract
cat mater/lib/tests/fixtures/car_v2/spaceglenda_wrapped.car | cargo r -r -p mater-cli -- extract -
# convert
cat mater/lib/tests/fixtures/original/spaceglenda.jpg | cargo r -r -p mater-cli -- convert --overwrite -
```

### DIFF
--- a/mater/cli/src/convert.rs
+++ b/mater/cli/src/convert.rs
@@ -9,15 +9,18 @@ pub(crate) async fn convert_file_to_car(
     output_path: &PathBuf,
     overwrite: bool,
 ) -> Result<Cid, Error> {
-    let source_file = File::open(input_path).await?;
     let output_file = if overwrite {
         File::create(output_path).await
     } else {
         File::create_new(output_path).await
     }?;
-    let cid = create_filestore(source_file, output_file, Config::default()).await?;
 
-    Ok(cid)
+    if input_path.as_os_str() == "-" {
+        create_filestore(tokio::io::stdin(), output_file, Config::default()).await
+    } else {
+        let source_file = File::open(input_path).await?;
+        create_filestore(source_file, output_file, Config::default()).await
+    }
 }
 
 /// Tests for file conversion.
--- a/mater/cli/src/extract.rs
+++ b/mater/cli/src/extract.rs
@@ -1,7 +1,10 @@
 use std::path::PathBuf;
 
 use mater::{CarExtractor, Error};
-use tokio::fs::File;
+use tokio::{
+    fs::File,
+    io::{AsyncReadExt, BufReader},
+};
 
 /// Extracts a file to `output_path` from the CARv2 file at `input_path`
 pub(crate) async fn extract_file_from_car(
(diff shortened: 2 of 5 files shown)


mater/cli/src/convert.rs
```diff
@@ -9,15 +9,18 @@ pub(crate) async fn convert_file_to_car(
...
-    let source_file = File::open(input_path).await?;
     let output_file = if overwrite {
         File::create(output_path).await
     } else {
         File::create_new(output_path).await
     }?;
-    let cid = create_filestore(source_file, output_file, Config::default()).await?;
 
-    Ok(cid)
+    if input_path.as_os_str() == "-" {
```
aidan46, line 18: `"-"` means that there is a pipe?

mater/cli/src/extract.rs
```diff
@@ -15,10 +18,29 @@ pub(crate) async fn extract_file_from_car(
...
+        // NOTE(@jmg-duarte,17/02/2025): this approach is bound to be inefficient for large inputs
+        // given that we're forced to load the whole input into memory upfront
+        // a possible alternative could be implementing a Reader that takes an AsyncRead
+        // and makes it AsyncSeek by keeping read contents in memory and forward seeks (where
+        // forward means that said part of the stream hasn't been loaded into memory yet) start
+        // loading file as needed before returning the new position
+
+        // Arbitrary value, it's just so the buffer doesn't constantly resize
+        // while loading the first bytes
+        let mut buffer = Vec::with_capacity(16 * 1024);
```
aidan46: I would put this in a const, easier to evaluate and change later if needed.

mater/cli/src/convert.rs
aidan46: Read the other file, I guess it does TIL

jmg-duarte: It means it should read from stdin:
* https://community.spiceworks.com/t/using-as-a-filename-in-linux-bash-shell/799135
* https://unix.stackexchange.com/questions/16357/usage-of-dash-in-place-of-a-filename

mater/cli/src/main.rs
```diff
@@ -74,9 +83,18 @@ async fn main() -> Result<(), Error> {
             overwrite,
         } => {
             let output_path = output_path.unwrap_or_else(|| {
-                let mut new_path = input_path.clone();
-                new_path.set_extension("");
-                new_path
+                // If we let the output become `-.car` it isn't only weird
+                // terminals are annoying with it
+                if input_path.as_os_str() == "-" {
+                    let file_name = "stdin.car";
```
cernicc: No extension should be set

mater/cli/src/main.rs
```diff
@@ -74,9 +83,18 @@ async fn main() -> Result<(), Error> {
...
+                    let mut path = PathBuf::with_capacity(file_name.len());
+                    path.set_file_name(file_name);
+                    path
+                } else {
+                    let mut new_path = input_path.clone();
+                    new_path.set_extension("car");
```
cernicc: Same here

mater/cli/src/main.rs
```diff
@@ -74,9 +83,18 @@ async fn main() -> Result<(), Error> {
             overwrite,
         } => {
             let output_path = output_path.unwrap_or_else(|| {
-                let mut new_path = input_path.clone();
-                new_path.set_extension("");
-                new_path
+                // If we let the output become `-.car` it isn't only weird
```
cernicc:
```suggestion
                // If we let the output become `-` it isn't only weird
```

mater/cli/src/main.rs
jmg-duarte: Good catch, copy paste, you know how it is

mater/cli/src/main.rs
jmg-duarte: Here should actually be `""` so we remove the extension instead

cernicc (approved):

aidan46 (approved):

PR #763, opened 2025-02-18
rfc: Add deal parameters

### LINKED ISSUE #760: rfc: Deal parameters
Write an RFC with the proposal for deal parameters by the storage provider, allowing for automatic deal making. This proposal should include parameters the storage provider sets on-chain, how the storage provider will act on the proposed deal and how it changes the user experience from the client-side.

aidan46:
### Description

This PR adds an RFC for adding deal parameters that the storage provider can set. Allowing use-case like automatic deal making and auction houses.

Closes #760


rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+All these requirements make sense but do not give the storage provider any control over what is important for them, profit 💰.
+Allowing the storage provider to set terms for deals they accept will increase transparency on storage deal prices and allows for storage providers to automatically accept deals.
+This, in turn, will improve the storage client experience by making the price of storage more available.
+Storage clients also no longer need to propose a deal and wait for it to be accepted by the storage provider.
+They can propose a deal within the parameters set by the storage provider and be sure that the deal will be accepted.
+
+## Deal Parameters
+
+Storage providers will be able to set their deal parameters by sharing these parameters after registering as a storage provider on the polka-storage chain.
+The deal parameters are not shared during registration because we do not want to force storage providers to use this feature.
```
jmg-duarte, line 24: I understand your question on the meeting now. This wouldn't even make sense in general because the parameters may change over time — the parameters should have an expiration date as well

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+A new extrinsic will be added to the storage provider pallet in the polka-storage chain to allow storage providers to advertise their deal parameters on-chain.
+Storage clients will be able to query deal parameters of all storage providers so they can find the best deal for them.
+Deal parameters could have endless options, a good starting point is for storage providers to set a lower and upper bound for storage price per block and the duration of a deal.
+These parameters can be expanded to include things such as collateral bounds and different price bounds depending on the size of the deal.
```
jmg-duarte: You should note that the provider should be able to register multiple parameters at the same time.
Example: for deals that have a really long duration, they may want to do a lower price

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+```rust
+struct DealPriceBound<Balance> {
+    lower: Balance,
+    upper: Option<Balance>,
+}
+
+struct DealDurationBound<BlockNumber> {
+    lower: BlockNumber,
+    upper: Option<BlockNumber>,
+}
```
jmg-duarte: I feel that this is slightly better (the Option on lower is to be discussed, I'm not sure if forcing it or not is a good idea)
```
struct Bound<T> { lower: Option<T>, upper: Option<T> }
type DealPriceBounds = Bound<Balance>;
type DealDurationBounds = Bound<BlockNumber>;
```

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+</details>
+
+## Pallet Changes
+
+### Storage Provider Pallet
+
+The storage provider pallet will need some changes to support deal parameters for automatic deal making.
```
jmg-duarte: I'm not sure if we should have this implemented in the SP crate but another crate might not be a great idea either

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+</details>
+
+## Storage Provider
+
+The storage provider server and client will need some changes to accommodate supporting deal parameters.
+
+The storage provider client will mostly stay the same, only adding an interface to retrieve deal parameters.
+
+The storage provider server will check the proposed deal parameters to see if the values fall within what they accept and process deals in the same way.
+New config values will need to be added to the server so that the storage provider can supply the deal parameters.
```
jmg-duarte: For the "alpha" only, ideally we'd have a control interface for the SP

rfcs/005-deal-parameters.md
th7nder: I don't agree that they should have an expiration date. If they're controllable by the storage provider, then storage provider should not have to do any additional job to extend them.

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+Storage providers will be able to set their deal parameters by sharing these parameters after registering as a storage provider on the polka-storage chain.
+The deal parameters are not shared during registration because we do not want to force storage providers to use this feature.
+A new extrinsic will be added to the storage provider pallet in the polka-storage chain to allow storage providers to advertise their deal parameters on-chain.
+Storage clients will be able to query deal parameters of all storage providers so they can find the best deal for them.
+Deal parameters could have endless options, a good starting point is for storage providers to set a lower and upper bound for storage price per block and the duration of a deal.
+These parameters can be expanded to include things such as collateral bounds and different price bounds depending on the size of the deal.
+
+<details open>
+
+<summary><b>Deal Parameters</b></summary>
```
th7nder, line 33: I think we're also missing the Sector Size specification. It'll probably matter, price_per_block for smaller sectors probably will be higher?

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+</details>
+
+## Pallet Changes
+
+### Storage Provider Pallet
+
+The storage provider pallet will need some changes to support deal parameters for automatic deal making.
+We do not want to force storage providers to set deal parameters so we need a new extrinsic to register deal parameters for a storage provider, `register_deal_parameters(origin: OriginFor<T>, deal_parameters: DealParameters<BalanceOf<T>, BlockNumberFor<T>>)`.
```
th7nder:
```suggestion
We do not want to force storage providers to set deal parameters so we need a new extrinsic to register deal parameters for a storage provider, `publish_deal_parameters(origin: OriginFor<T>, deal_parameters: DealParameters<BalanceOf<T>, BlockNumberFor<T>>)`.
```

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+## Storage Provider
+
+The storage provider server and client will need some changes to accommodate supporting deal parameters.
+
+The storage provider client will mostly stay the same, only adding an interface to retrieve deal parameters.
+
+The storage provider server will check the proposed deal parameters to see if the values fall within what they accept and process deals in the same way.
+New config values will need to be added to the server so that the storage provider can supply the deal parameters.
+These values should align with what is registered on-chain.
+This means that whenever the storage provider server is (re)started, it should check that the given deal parameters align with what is stored on-chain and update the values if this is not the case.
```
th7nder: Not sure whether restart makes sense, why would an SP need to restart only if some 'runtime values' change.
I'd say in the beginning it should get the params from the chain always before accepting a deal, later we can add some privileged RPC to 'force refresh'

rfcs/005-deal-parameters.md
th7nder: From the 'conceptual perspective' it fits in the `Market` Pallet more.

th7nder (commented):
If we had to go with it as a MVP, I'd approve it. Left some comments to consider, but overall looks good. I don't know, when I was thinking about it the first time it felt more complex, maybe I was overthinking.

I think that the choice that SPs are setting those parameters by themselves is simplifying this work a lot. We'd need to suggest 'some prices for them' as a default for starters, to encourage SP to onboard, but I think from the technical standpoint it'll work.

rfcs/005-deal-parameters.md
aidan46: I tend to agree that the market pallet might be more suited for deal parameters as it handles all things deal related.

rfcs/005-deal-parameters.md
aidan46: Having the bound generic is a great idea. The reason I did not set the lower as an option because for me it does not make sense for the SP to have a lower bound on the duration.

rfcs/005-deal-parameters.md
aidan46: I agree with @th7nder, if the SP wants to update their parameters they are free to do so. Also, if they do have an expiry, and the SP has 'fine-tuned' their parameters, we enforce gas on them at every expiry.

rfcs/005-deal-parameters.md
aidan46: What I meant is if the SP restarts the polka-storage-provider-server it should check if the config values it set align with those on-chain. I think that getting the params from the chain might be a better place to start tho.

rfcs/005-deal-parameters.md
aidan46: Yeah I listed some more possible values but the JSON was just where to start. I can update the JSON to include more parameters.

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,138 @@
...
+Allowing the storage provider to set terms for deals they accept will increase transparency on storage deal prices and allows for storage providers to automatically accept deals.
+This, in turn, will improve the storage client experience by making the price of storage more available.
+Storage clients also no longer need to propose a deal and wait for it to be accepted by the storage provider.
+They can propose a deal within the parameters set by the storage provider and be sure that the deal will be accepted.
+
+## Deal Parameters
+
+Storage providers will be able to set their deal parameters by sharing these parameters after registering as a storage provider on the polka-storage chain.
+The deal parameters are not shared during registration because we do not want to force storage providers to use this feature.
+A new extrinsic will be added to the storage provider pallet in the polka-storage chain to allow storage providers to advertise their deal parameters on-chain.
```
cernicc:
```suggestion
A new extrinsic will be added to the market pallet in the polka-storage chain to allow storage providers to advertise their deal parameters on-chain.
``` ?

File: rfcs/005-deal-parameters.md (hunk 7 above)
Comment (cernicc) [inline rfcs/005-deal-parameters.md]: Shouldn't each storage provider set at least a price? Unless they will be ok with accepting any deal. But that can also mean free deals.

File: rfcs/005-deal-parameters.md [hunk 10]
```diff
@@ -0,0 +1,118 @@
...
+struct Bound<T> {
+    lower: Option<T>,
+    upper: Option<T>,
+}
+
+type DealPriceBounds = Bound<Balance>;
+type DealDurationBounds = Bound<BlockNumber>
+
+struct DealParameters {
+    price: DealPriceBound,
```
cernicc: What is the reason for the price bound? The client will probably always choose the min price for the specific provider?

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,118 @@
...
+    duration: DealDurationBound,
+}
+
+#[pallet::storage]
+pub type DealParametersTable<T: Config> =
+    StorageMap<_, _, T::AccountId, DealParameters>
```
cernicc: Hmm. If we want to support cases where longer deals are discounted etc. The value should be a vector of DealParameters. In that case the provider can check if the deal fits into any of the rules specified.

aidan46: You are right. Although, I think it would be better to hold a mapping for each storage provider where:
```rust
Map<SectorSize, DealParameters>
```
The key for the map is the sector size since that is most likely what the SP will base their price/duration on.

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,118 @@
...
+<details open>
+
+<summary><b>Deal Parameters</b></summary>
+
+```json
+{
+    "8MiB": {
+        "price_per_block": {
+            // Price is in the smallest unit (plancks)
+            "minimum": 1_000_000,
```
Jinxit: Does it make sense to include the unit in the field name, either on `minimum`/`maximum` or `price_per_block`, or is that always plancks in the Polkadot world?

Also applies for `duration`.

rfcs/005-deal-parameters.md
aidan46: That makes sense

rfcs/005-deal-parameters.md
cernicc: > that is most likely what the SP will base their price/duration on.

Not necessarily. I as a sp could also discount the deals that are longer for the same sector size. So I can incentivize the longer deals etc.

rfcs/005-deal-parameters.md
Jinxit: Is it ever going to be an auction type situation, or are we counting on contention for space being low enough to always accept the first available deal?

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,122 @@
...
+This will be a signed extrinsic that takes in the deal parameters and published these parameters with the storage provider calling the extrinsic.
+
+The deal parameters will be stored in a `StorageMap` where the `AccountId` is the key and the `DealParameters` is the value.
+
+<details open>
+
+<summary><b>Deal Parameter Types</b></summary>
+
+```rust
+struct Bound<T> {
```
Jinxit: This type is only used for DealDurationBounds now, maybe inline it?

rfcs/005-deal-parameters.md
Jinxit: And with the change to only using a minimum price I suppose this json needs updating as well.

aidan46: It does make sense to add parsing to support DOT and hours, days, etc. Maybe not for the MVP but I'll add a note.

rfcs/005-deal-parameters.md
aidan46: > Is it ever going to be an auction type situation, or are we counting on contention for space being low enough to always accept the first available deal?

Implementing these deal parameters will allow for an auction house to be created. Providers setting their minimum price vs clients proposing price.

rfcs/005-deal-parameters.md
jmg-duarte: I agree with your points on expiration but then we need to keep in mind removing them

rfcs/005-deal-parameters.md
jmg-duarte: That's kinda their issue though

rfcs/005-deal-parameters.md
aidan46: Okay, that makes sense too. Makes parsing the values a bit more complicated but gives more control to the SP

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,119 @@
...
+
+```rust
+struct DealDurationBound<BlockNumber> {
+    lower: Option<BlockNumber>,
+    upper: Option<BlockNumber>,
+}
+
+struct DealParameters<Balance, BlockNumber> {
+    minimum_price: Balance,
+    duration: DealDurationBound<BlockNumber>,
```
Jinxit, line 76: And also this

rfcs/005-deal-parameters.md
```diff
@@ -0,0 +1,119 @@
...
+}
+```
+
+</details>
+
+## Market Pallet
+
+### Publishing Deal Parameters
+
+The market provider pallet will need some changes to support deal parameters for automatic deal making.
```
jmg-duarte: Market provider?

jmg-duarte (dismissed):
For a PoC, LGTM

rfcs/005-deal-parameters.md
aidan46: Oopsie

rfcs/005-deal-parameters.md
aidan46: I don't really understand what you mean by removing them. When should the deal parameters be removed?

rfcs/005-deal-parameters.md
aidan46: I think we can add enforcing a minimum price if there is a need for it

cernicc (approved):

rfcs/005-deal-parameters.md
jmg-duarte: As an SP 
* I add deal parameters
* I gain popularity and have too many people requesting my service
* I want to change the parameters - remove + insert
Or
* I want to remove a parameter because that class of deals no longer interests me

jmg-duarte (approved):
LGTM we just need to finish discussing the parameter removal/modification

PR #764, opened 2025-02-18
fix(pipeline): precommit scheduling

cernicc:
### Description

If the sector was immediately full, we started the precommit process by sending a message directly. The bug was that in this case we didn't have the scheduled task saved. So later we early returned [here](https://github.com/eigerco/polka-storage/blob/c6c7dea84d2ff41fff70f218839e210a4e784b93/storage-provider/server/src/pipeline/mod.rs#L312).

I've fixed the bug by scheduling a precommit immediately if the sector is full.

### Checklist

### DIFF
--- a/storage-provider/server/src/pipeline/add_piece.rs
+++ b/storage-provider/server/src/pipeline/add_piece.rs
@@ -61,7 +61,11 @@ pub async fn add_piece(
             fill_percentage,
             fill_threshold,
         );
-        return state.send_pre_commit(sector_number);
+
+        // Schedule a new precommit to be executed immediately
+        let when = Duration::from_secs(0);
+        schedule_pre_commit(state.clone(), tracker, sector_number, when).await;
+        return Ok(());
     }
     tracing::debug!(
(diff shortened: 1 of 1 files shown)


storage-provider/server/src/pipeline/add_piece.rs
```diff
@@ -61,7 +61,11 @@ pub async fn add_piece(
             fill_percentage,
             fill_threshold,
         );
-        return state.send_pre_commit(sector_number);
+
+        // Schedule a new precommit to be executed immediately
+        let when = Duration::from_secs(0);
+        schedule_pre_commit(state.clone(), tracker, sector_number, when).await;
```
jmg-duarte:
```suggestion
        schedule_pre_commit(state.clone(), tracker, sector_number, Duration::ZERO).await;
```

jmg-duarte (dismissed):
Great find!

jmg-duarte (approved):

th7nder (approved):

PR #765, opened 2025-02-18
fix(mater): remove unnecessary PartialNode abstraction

jmg-duarte:
### Description

Looking back it was obvious, but hindsight 20/20.

Simplifies the API for the UX fix, more fixes to come to avoid a mega PR.


cernicc (approved):
Nice

th7nder (approved):

PR #768, opened 2025-02-19
chore: VS Code settings

Jinxit:
### Description

Decided to split this from CI changes to make it an easier thing to get merged. There are now two main changes:
1. RA uses a separate target directory, meaning RA builds no longer cause rebuilds in nix, and vice versa.
2. The settings file has been renamed to `settings.default.json`, meaning developers can copy that file as a starting point but have their own settings that won't be committed to the repo.

If you don't make a copy of `settings.default.json` as `settings.json` after this is merged, you won't have the right settings in VS Code.

### Checklist

  - [x] See above.
- [ ] ~~If there are follow-ups, have you created issues for them?~~
- [ ] ~~Have you tested this solution?~~
- [ ] ~~Were there any alternative implementations considered?~~


jmg-duarte (approved):
LGTM. I wonder if there's a way to tell git "dont actually delete settings.json locally, just remove from cache"

cernicc (approved):

PR #769, opened 2025-02-20
ci: Run CI on all 'ready-to-review' PR commits

Jinxit:
### Description

Hopefully this will make merging PRs faster because the workflow will already be running when you reopen the page rather than having to re-tag.

This will trigger more builds than before, but if it becomes a problem we can revisit this and maybe set a manual approval before building instead of using the tag.

### Checklist

  - [x] See above
- [ ] ~~If there are follow-ups, have you created issues for them?~~
  - [ ] ~~If yes, which ones? / List them here~~
  - I don't think I can since the PR doesn't contain code changes, but I'm open to suggestions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,7 +19,7 @@ concurrency:
 
 jobs:
   check-files:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     outputs:
       skip: ${{ steps.check-files.outputs.skip }}
     steps:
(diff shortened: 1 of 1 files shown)


cernicc (dismissed):

Jinxit:
I also (on @jmg-duarte's recommendation) changed the GH runner steps to run on our self-hosted runner.

cernicc (approved):

jmg-duarte (approved):

PR #772, opened 2025-02-20
feat(mater): replace blockstore with a more general abstraction

jmg-duarte:
### Description

Provides more liberty over the content being written and aims to be more extensible.
Furthermore, it removes some dependencies, adds a better duplication demo.

*Names are not final, a bigger re-org is coming in another PR*

The 🥩 & 🥔🥔 of the PR is in mater/lib/src/stores/blockstore.rs


cernicc (dismissed):
I like the new Blockwritter approach.

mater/lib/benches/benchmark.rs
```diff
@@ -131,53 +135,25 @@ fn generate_content(params: &Params) -> Vec<u8> {
     bytes
 }
 
-/// Read content to a Blockstore. This function is benchmarked.
-async fn read_content_benched(content: &[u8], mut store: Blockstore) {
+/// Read/Write content to a Blockstore. This function is benchmarked.
+async fn read_write_content_benched<W>(content: &[u8], mut store: Blockwriter<W>)
```
th7nder, line 139: Why are we benching both read & write at the same time? They're kind-of separate concerns.

mater/lib/src/chunker.rs
```diff
@@ -0,0 +1,54 @@
...
+            // as it may be able to return more bytes later, *however*,
+            // this means there is no right way of knowing when the reader is fully exhausted!
+            // If we need to support a case like that, we just need to track how many times
+            // the reader returned 0 and break at a certain point
+            if source.read_buf(&mut buf).await? == 0 {
+                // EOF but there's still content to yield -> yield it
+                if buf.len() > 0 {
+                    let chunk = buf.split();
+                    yield chunk.freeze();
+                }
```
th7nder: I think we should implement this logic which @pete-eiger has added in some of his PRs. 
Problem being: what if the `buf.len() > chunk_size` at this point.

mater/lib/src/file_reader.rs
```diff
@@ -174,24 +174,18 @@ where
...
-        let raw_input = std::iter::repeat(0).take(4096).collect::<Vec<u8>>();
-
-        let mut bs = Blockstore::with_parameters(Some(1024), None);
-        bs.read(Cursor::new(raw_input.clone())).await.unwrap();
-
-        // 1519 is the expected CAR file size after deduplicating and writing the CAR file
-        let mut out_car_buffer = Vec::with_capacity(1519);
-        bs.write(&mut out_car_buffer).await.unwrap();
-        assert_eq!(out_car_buffer.len(), 1519);
+        let raw_input = tokio::fs::read("tests/fixtures/original/zero")
```
th7nder, line 181: What's inside 'zero'? I feel like the raw input previously was more clear in the intent.

mater/lib/src/stores/blockstore.rs
```diff
@@ -1,370 +1,235 @@
...
-        let mut result_buffer = vec![];
-        store.write(&mut result_buffer).await.unwrap();
+        let mut store = Blockwriter::in_memory();
+        store.write_from(file).await.unwrap();
+        let output = store.finish().await.unwrap().into_inner();
 
-        let car_contents = tokio::fs::read("tests/fixtures/car_v2/spaceglenda.car")
-            .await
-            .unwrap();
+        tokio::fs::write("/tmp/zero.car", &output).await.unwrap();
```
th7nder: Do we need this write?

mater/lib/src/stores/blockstore.rs
```diff
@@ -1,370 +1,235 @@
...
-            let (cid, bytes) = block?;
-            self.insert(cid, bytes, true);
-
-            // If the stream is exhausted, we know the current block is the root
-            if tree.peek().await.is_none() {
-                // The root should always be indexed, there's no official spec saying it should though, it just makes sense.
-                // So, if the insert line is changed, the root should be placed in the `indexed` structure here
-                self.root = Some(cid);
-            }
+        if !self.started {
```
th7nder, line 82: What's up with this started stuff? Why would we support calling this function twice?

th7nder (changes requested):

mater/lib/benches/benchmark.rs
jmg-duarte: Because the blockwriter was implemented in a way that you dont need to hold everything in memory before dumping to the writer.

Imagine you're car-ifying a file with 10GB, the previous approach would load it fully into memory and only then write it to disk.

mater/lib/src/file_reader.rs
jmg-duarte: It's a bunch of zero bytes, I can keep the previous approach but I need to keep the `zero.car` at least

mater/lib/src/stores/blockstore.rs
jmg-duarte: No, it was debug

mater/lib/src/stores/blockstore.rs
jmg-duarte: Adding two files to the CAR. It creates two roots, even though that specific case is not fully supported this is here because when extending the API it will be necessary, though with a few adjustments.

mater/lib/src/chunker.rs
jmg-duarte: When we read the source for `reserve`([1](https://github.com/tokio-rs/bytes/blob/e0f3a25faeca8babae5b4c5a810b8655adbcb96e/src/bytes_mut.rs#L592), [2](https://github.com/tokio-rs/bytes/blob/e0f3a25faeca8babae5b4c5a810b8655adbcb96e/src/bytes_mut.rs#L608)), we see the following checks:
https://github.com/tokio-rs/bytes/blob/e0f3a25faeca8babae5b4c5a810b8655adbcb96e/src/bytes_mut.rs#L612
https://github.com/tokio-rs/bytes/blob/e0f3a25faeca8babae5b4c5a810b8655adbcb96e/src/bytes_mut.rs#L688
Which our buffer doesn't meet (I checked with the debugger), as soon as we `split`/`split_to` the buffer gets promoted to `Arc`:
https://github.com/tokio-rs/bytes/blob/master/src/bytes_mut.rs#L1051-L1066
Which then makes it fail the `is_unique` check:
https://github.com/tokio-rs/bytes/blob/e0f3a25faeca8babae5b4c5a810b8655adbcb96e/src/bytes_mut.rs#L1466-L1478

This excludes the case where the buffer capacity is doubled:
https://github.com/tokio-rs/bytes/blob/e0f3a25faeca8babae5b4c5a810b8655adbcb96e/src/bytes_mut.rs#L726-L736
Leaving us to the following:
https://github.com/tokio-rs/bytes/blob/e0f3a25faeca8babae5b4c5a810b8655adbcb96e/src/bytes_mut.rs#L762-L768
Where our allocated capacity is bigger than the `repr` values (you can test in [the Playground](https://github.com/tokio-rs/bytes/blob/master/src/bytes_mut.rs#L762-L765)).

I can still add it if you guys think it is necessary

th7nder: I mean, we're relying on the internal implementation of `BytesMut`, not on the API guarantees that it's like that. If it works that's fine, what if it stops?

th7nder (dismissed):

th7nder (approved):

cernicc (approved):

PR #774, opened 2025-02-21
feat(mater): add mater::ReadOnlyBlockstore

jmg-duarte:
### Description

Adds `CarReadOnlyBlockstore` (still looking for better names, but there still needs to be a PR finishing organizing this chaos).

Doesn't remove the `FileBlockstore` just yet because we still need the read and write blockstore.

Requires #773


mater/lib/src/file_reader.rs
```diff
@@ -17,39 +17,40 @@ use crate::{multicodec, v1::BlockMetadata, v2, Error};
...
 
+impl<R> CarExtractor<R> {
+    /// Creates a new [`CarExtractor`] from the given reader.
+    pub async fn new(reader: R) -> Result<Self, Error>
+    where
+        R: AsyncRead + AsyncSeek + Unpin,
+    {
+        let mut self_ = Self {
+            reader: v2::Reader::new(reader),
+            index: HashMap::with_capacity(1),
```
th7nder, line 33: What's the point o  `with_capacity(1)`?

mater/lib/src/file_reader.rs
```diff
@@ -170,6 +171,224 @@ where
...
+    where
+        R: AsyncRead + AsyncSeek + Unpin + Send + Sync,
+    {
+        async fn get<const S: usize>(
+            &self,
+            cid: &ipld_core::cid::CidGeneric<S>,
+        ) -> blockstore::Result<Option<Vec<u8>>> {
+            let cid = to_blockstore_cid(cid)?;
+            self.inner
+                .write()
```
th7nder, line 267: .write() lock for .get()?
It's counterintuitive, took me a while, but I get it. We need to rewind underlying reader to read data, hence the RwLock. I guess it is what it is.

th7nder (dismissed):

mater/lib/src/file_reader.rs
cernicc: True. You can argue here that `with_capacity(1)` doesn't optimize anything. Because the number of allocations will be the same as if you used `new`.

mater/lib/src/file_reader.rs
```diff
@@ -170,6 +171,224 @@ where
...
+                    let (_, block) = self.reader.read_block().await?;
+                    Ok(Some(block))
+                }
+                None => Ok(None),
+            }
+        }
+    }
+
+    /// A read-only [`blockstore::Blockstore`] implementation of [`CarExtractor`].
+    pub struct CarReadOnlyBlockstore<R> {
```
cernicc: What do you think of `ReadOnlyBlockstore`?

cernicc (dismissed):

mater/lib/src/file_reader.rs
jmg-duarte: > Because the number of allocations will be the same as if you used new.

No it won't, as soon as you insert an element the HashMap needs to resize, forcing an extra allocation, see this example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=aa9fe4806e6139dc3f6b88d808353226

cernicc: True. But isn't the number of allocations still the same? Only the time of when the allocation was made is different? My point is made for this example (when the `with capacity` <= `initial resizing size`). It's different in cases when you have a larger initial capacity because you "skip" the resizings. e.g. If we would have `with_capacity(4)`, then the first allocation would be optimized away (the first allocation is for the 3 elements). 

In the example you pasted, there is  "hidden" allocation that happens before `with_capacity` returns.

jmg-duarte: Rok's right — https://docs.rs/hashbrown/0.15.2/hashbrown/struct.HashMap.html#method.new
I also tested with valgrind and saw the same.

My intuition was wrong & TIL

th7nder (approved):

cernicc (approved):

Issue #742, opened 2025-02-12
Delia - Replication


th7nder:
Make 3 deals instead of 1 automatically in Delia.
Details have to be figured out.

jmg-duarte:
Dumping some things here before writing a proper post-mortem.

* We wanted the client to be serve-able from GitHub Pages, that way it could also be served from IPFS (like Polkadot JS) with no issues.
* To this end, I started implementing Libp2p on the client
* First issue was the lack of WS support on the server — this was "easy", just add it
* Then the client wasn't able to retrieve the right multiaddress since we used rendezvous and the registered address was TCP only, not the WS, since the servers didn't use WS between each other — Started changing into Kademlia, the DHT offers persistent queries and network wide (gossip is ephemeral, it would need an extra store, etc)
* Start implementing Libp2p JS on the client side in an [isolated demo](https://github.com/jmg-duarte/libp2p-rs-js-dht)
* Libp2p JS simply doesn't work, I've [asked for help to no avail](https://github.com/libp2p/js-libp2p/discussions/3058) 
* I tried WASM with WebSockets but apparently, in the browser, they require WSS instead, though it didn't fail and the Identify + the beginning of Kademlia worked
  * According to the logs, the client is closing the connection for some reason

neriumpete:
> Libp2p JS simply doesn't work, I've https://github.com/libp2p/js-libp2p/discussions/3058

hm, if only there was someone in the team that warned you 15 times that js-libp2p wont work and that you should use js-waku instead :)

neriumpete:
> First issue was the lack of WS support on the server — this was "easy", just add it

again, something that wouldn't have been an issue if there was someone to recommend using waku.

jmg-duarte:
This has been solved a while back already

PR #776, opened 2025-02-23
rfc: delia replication

cargopete:
RFC for #742


rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Introduction
+
+Polka Storage relies on replicating data across multiple storage providers to ensure high availability and fault tolerance. Currently, Delia handles storage deals on a one-to-one basis: a client calculates a piece CID, proposes a deal, uploads a file, and then publishes the deal. For true replication, it is desirable for a client to automatically generate multiple identical deals—specifically three—so that the same piece of data is stored redundantly.
+
+This RFC outlines the necessary modifications to the client-side deal flow to support the automatic submission of three deals. The changes will involve:
+
+- Looping through the existing deal proposal flow three times.
```
th7nder, line 16: How are we selecting the individual storage providers from the list? Round-robin?

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+- Making parallel JSON‑RPC calls against individual storage provider servers.
+- Aggregating status feedback in the user interface.
+- Addressing provider selection: either automatically choosing three providers or allowing the user to manually select three.
```
th7nder: I think we should automatically select 3 and allow to replace any of them in the UI.

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+   After file upload, the client:
+   - Encodes the deal proposal via the storage provider server’s HTTP endpoint (e.g. `/encode_proposal`).
+   - Uses the Polkadot extension to sign the encoded proposal.
+   - Publishes the deal via a JSON‑RPC call (e.g. `v0_publish_deal`).
+
+### Provider Selection
+
+Currently, the FE allows the user to select one storage provider from a list. With the multi-deal approach for replication, we need to consider:
+
+- **Option A:** Automatically select three storage providers from the available list (for example (purely hypothetical atm), the top three based on performance or reputation), so that data is replicated across different nodes.
```
th7nder: We don't have performance/reputation system in place, we didn't even design it. I'd say 'random 3 different storage providers' is a good starting point. If we need to, we'll make additional RFC to improve this.

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Open Questions for Discussion
+
+1. **Batch Size Configuration:**  
+   Should we allow clients to specify a different number of deals (other than 3) for replication purposes, or is 3 the fixed value for the initial implementation?
+2. **Provider Selection Mode:**  
+   - Should the client automatically select three storage providers from the available list to maximize replication, or allow the user to manually choose three?
+   - If automatic, what criteria should be used?
+3. **Partial Failures:**  
+   What is the expected behavior if one or more proposals fail while others succeed? Should the client automatically retry the failed proposals, or simply notify the user?
```
th7nder: We need to try-force to make sure at least 3 storage providers have this data stored with some max-retry atttempy mechanism.

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+- **Error Handling:**  
+  If one proposal fails while the others succeed, the client should clearly report the error and optionally offer a retry mechanism, ensuring that data replication is maintained.
+
+### Documentation
+
+- Update the Delia documentation to reflect the new multi-deal submission process with an emphasis on replication.
+- Include examples showing how a client initiates three simultaneous proposals to achieve redundancy.
+- Document any new UI changes related to replication.
+
+## Implementation Details
```
th7nder, line 72: It's not part of *this task* but keep in mind that we'll need to probably some day retrieval which is 'replication-aware'.

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+- **Error Handling:**  
+  Aggregate results from the three RPC calls. Decide on a strategy for partial failures (e.g., retry the failed proposals or notify the user for manual intervention).
+
+## Open Questions for Discussion
+
+1. **Batch Size Configuration:**  
+   Should we allow clients to specify a different number of deals (other than 3) for replication purposes, or is 3 the fixed value for the initial implementation?
+2. **Provider Selection Mode:**  
+   - Should the client automatically select three storage providers from the available list to maximize replication, or allow the user to manually choose three?
+   - If automatic, what criteria should be used?
```
th7nder: In my opiinon, automatic, random unique SP selection, parameters hard-coded to 3 in the beginning.

th7nder (commented):
I think this is the right direction, some minor comments.

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+  Submitting multiple deals automatically streamlines the process and reduces manual intervention, which is especially important in a decentralized system where network conditions and provider availability may vary.
+
+- **User Experience:**  
+  Clients gain higher confidence in data durability when the system automatically issues multiple proposals, knowing that their data will be replicated across the network.
+
+- **Foundation for Further Enhancements:**  
+  This feature complements the automated deal acceptance framework (see RFC‑005) and opens the door for future optimizations such as dynamic batch sizing and advanced provider selection strategies.
+
+- **Market Dynamics:**  
+  Increasing deal submission frequency may improve price discovery and transparency, ultimately benefiting both storage providers and clients.
```
cernicc: `Increasing deal submission frequency may improve price discovery and transparency` What does this mean?

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
+# RFC-006: Automated Multi-Deal Submission in Delia for Replication
+
+**Author:** Pete — @pete-eiger  
+**Date:** 2025-02-24
+
+## Abstract
+
+This RFC proposes modifications to the Delia storage deal flow so that, instead of submitting a single deal proposal, the client automatically submits three identical deal proposals. By looping through the proposal process three times and executing the necessary RPC calls in parallel, the client not only streamlines the storage deal process but also increases data replication in the protocol. This replication is a core feature of the Polka Storage solution, ensuring that data is redundantly stored across multiple providers for improved availability, durability, and resilience in a decentralized environment. This change lays the groundwork for further enhancements such as automated deal acceptance and flexible provider selection.
```
jmg-duarte: > This change lays the groundwork for further enhancements such as automated deal acceptance and flexible provider selection.

This is orthogonal to this RFC

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+   - Construct an identical deal proposal (using the calculated piece CID, client address, and other necessary parameters).
+   - Send an RPC request (e.g. via `v0_propose_deal`) to the storage provider’s RPC API.
+   - Execute the calls in parallel since each storage provider server operates independently.
+
+2. **Upload File:**  
+   For each deal CID returned from the proposal call, the client uploads the file via an HTTP PUT request to the corresponding endpoint (e.g. `/upload/{dealCid}`).
+
+3. **Publish Deal:**  
+   After file upload, the client:
+   - Encodes the deal proposal via the storage provider server’s HTTP endpoint (e.g. `/encode_proposal`).
```
cernicc: I think we shouldn't use `delia` featured endpoints anymore. Before implementing this, we should check what can be merged into the provider server and remove the feature `delia`. I think we should do that before building on top of it. It's going to be harder to remove later

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+  
+- **Signature & Publishing:**  
+  The encoding and signing process remains unchanged; each deal proposal is individually signed and published.
+  
+- **Error Handling:**  
+  Aggregate results from the three RPC calls. Decide on a strategy for partial failures (e.g., retry the failed proposals or notify the user for manual intervention).
+
+## Open Questions for Discussion
+
+1. **Batch Size Configuration:**
```
cernicc: We can make it configurable. 3 can be a default.

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+  This feature complements the automated deal acceptance framework (see RFC‑005) and opens the door for future optimizations such as dynamic batch sizing and advanced provider selection strategies.
+
+- **Market Dynamics:**  
+  Increasing deal submission frequency may improve price discovery and transparency, ultimately benefiting both storage providers and clients.
+
+## Proposed Changes
+
+### Client Modifications
+
+1. **Propose Deal (Three Times):**
```
cernicc: There is work being done on the deal acceptance parameterization. Each sp will be able to specify which deals it's prepared to accept. After the delia has chosen the batch of proposed providers. The user should get an option to reject or approve the aggregated deal.

rfcs/006-delia-replication.md
cargopete: changing this to "3 automatically chosen providers"

rfcs/006-delia-replication.md
cargopete: removing since it's actually not related to replication itself

rfcs/006-delia-replication.md
cargopete: makes sense, although technically these are unrelated workstreams i.e.  Delia expects certain endpoints from the storage provider, this will be the case regardless of how the SP server is started (with which features). I will raise this in Slack but I don't think it's a blocker for this RFC

rfcs/006-delia-replication.md
cargopete: I think I prefer to have it hardcoded to 3 at least for the first version of this feature landing in Delia, later on we can add more customization options in general

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Introduction
+
+Polka Storage relies on replicating data across multiple storage providers to ensure high availability and fault tolerance. Currently, Delia handles storage deals on a one-to-one basis: a client calculates a piece CID, proposes a deal, uploads a file, and then publishes the deal. For true replication, it is desirable for a client to automatically generate multiple identical deals—specifically three—so that the same piece of data is stored redundantly.
```
jmg-duarte: > Polka Storage relies on replicating data across multiple storage providers to ensure high availability and fault tolerance.

It doesn't "rely" it hasn't been implemented

rfcs/006-delia-replication.md
cernicc: It's fine if you leave it in. I just don't understand what you meant with the statement.

rfcs/006-delia-replication.md
jmg-duarte: That doesn't address @th7nder's question

rfcs/006-delia-replication.md
cargopete: Hmm I don't quite get this, couldn't this be automated on the Delia side? After all Delia will know the SP params so we will be able to select those SPs that we know will accept whatever parameters the user has inputted into Delia. Or am I wrong?

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+  Ensuring that data is stored in multiple locations is a cornerstone of a decentralized storage system. Automatically submitting three identical deals increases redundancy and resilience, so that if one deal or provider fails, the data remains available elsewhere.
+
+- **Efficiency:**  
+  Submitting multiple deals automatically streamlines the process and reduces manual intervention, which is especially important in a decentralized system where network conditions and provider availability may vary.
+
+- **User Experience:**  
+  Clients gain higher confidence in data durability when the system automatically issues multiple proposals, knowing that their data will be replicated across the network.
+
+- **Foundation for Further Enhancements:**  
+  This feature complements the automated deal acceptance framework (see RFC‑005) and opens the door for future optimizations such as dynamic batch sizing and advanced provider selection strategies.
```
jmg-duarte: This RFC/feature does not "open the door" to any of these

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,113 @@
...
+- Aggregating status feedback in the user interface.
+- Addressing provider selection: either automatically choosing three providers or allowing the user to manually select three.
+
+## Motivation
+
+- **Replication:**  
+  Ensuring that data is stored in multiple locations is a cornerstone of a decentralized storage system. Automatically submitting three identical deals increases redundancy and resilience, so that if one deal or provider fails, the data remains available elsewhere.
+
+- **Efficiency:**  
+  Submitting multiple deals automatically streamlines the process and reduces manual intervention, which is especially important in a decentralized system where network conditions and provider availability may vary.
```
jmg-duarte: While this sentence isn't false, it's not relevant to the document as it's not part of the motivation behind adding replication. We could very well add manual replication which would be better (since it provides more control & information to the end-user) in cases where "availabilty may vary" as you put it.

In Portuguese we call this "encher chouriços"

rfcs/006-delia-replication.md
jmg-duarte: It's not related, Pete is right in removing

rfcs/006-delia-replication.md
jmg-duarte: So it isn't really an open question then

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,94 @@
...
+  Use async patterns (e.g., `Promise.all`) to issue the three RPC calls concurrently.
+  
+- **Signature & Publishing:**  
+  The encoding and signing process remains unchanged; each deal proposal is individually signed and published.
+  
+- **Error Handling:**  
+  Aggregate results from the three RPC calls. If there is a failed upload attempt, Delia will retry 3 times (for each failed provider), if all 3 are unsuccessful it will select another storage provider at random and repeat this, with the goal of ensuring that the data is stored by at least 3 different storage providers.
+
+## Future work
+There will be work needed to make data retrieval be "replication-aware".
```
jmg-duarte: Not relevant to this PR. A proper future work here would be "implement N-ary replication instead of forcing 3 copies"

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,94 @@
...
+## Implementation Details
+
+- **Looping Mechanism:**  
+  Modify the existing deal proposal code to iterate three times. Each iteration constructs the same deal proposal and calls the RPC endpoint.
+  
+- **Parallel Execution:**  
+  Use async patterns (e.g., `Promise.all`) to issue the three RPC calls concurrently.
+  
+- **Signature & Publishing:**  
+  The encoding and signing process remains unchanged; each deal proposal is individually signed and published.
```
jmg-duarte: These don't seem relevant

rfcs/006-delia-replication.md
cargopete: was before i thought about it more ;> now isn't, yes

rfcs/006-delia-replication.md
cargopete: sorry I thought automatic meant random. So no, not round-robin, random via something like

```js
function getRandomElements(arr, count) {
  const shuffled = arr.slice(); // copy array
  for (let i = shuffled.length - 1; i > 0; i--) {
    const j = Math.floor(Math.random() * (i + 1));
    [shuffled[i], shuffled[j]] = [shuffled[j], shuffled[i]]; // swap elements
  }
  return shuffled.slice(0, count);
}

// Example usage:
const providers = ['Provider1', 'Provider2', 'Provider3', 'Provider4'];
const randomProviders = getRandomElements(providers, 3);
console.log(randomProviders);
```

rfcs/006-delia-replication.md
cargopete: relies on the concept, will add the future connotation to it

rfcs/006-delia-replication.md
cernicc: I agree. I thought we should resolve it before the actual RFC implementation. But still the endpoint `encode_proposal` only encodes the proposal. That will probably be a front-ends job. The RFC content currently depends on the endpoints provided only when the feature is enabled. Technically the discussion is relevant for this RFC.

rfcs/006-delia-replication.md
cernicc: Answer copied from slack:

Each SP will be able to configure its own params. Those params will basically specify which deal variations are they prepared to accept. e.g. if the sector size is N i only accept deals that pay at least M much. The price is specified by the client. But it's basically resolved from the info retrieved from the server side. Imagine that you take three cheapest providers. We would still have to check what are their prices and inform the user before they proceed with those selected providers.
Please correct me if my understanding is wrong.

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,82 @@
+# RFC-006: Automated Multi-Deal Submission in Delia for Replication
+
+**Author:** Pete — @pete-eiger  
+**Date:** 2025-02-24
+
+## Abstract
+
+This RFC proposes modifications to the Delia storage deal flow so that, instead of submitting a single deal proposal, the client automatically submits three identical deal proposals. By looping through the proposal process three times and executing the necessary RPC calls in parallel, the client not only streamlines the storage deal process but also increases data replication in the protocol. This replication is a core feature of the Polka Storage solution, ensuring that data is redundantly stored across multiple providers for improved availability, durability, and resilience in a decentralized environment.
```
aidan46: Replication is opt-in right?

rfcs/006-delia-replication.md
```diff
@@ -0,0 +1,82 @@
...
+## Introduction
+
+Polka Storage will rely on replicating data across multiple storage providers to ensure high availability and fault tolerance. Currently, Delia handles storage deals on a one-to-one basis: a client calculates a piece CID, proposes a deal, uploads a file, and then publishes the deal. For true replication, it is desirable for a client to automatically generate multiple identical deals—specifically three—so that the same piece of data is stored redundantly.
+
+This RFC outlines the necessary modifications to the client-side deal flow to support the automatic submission of three deals. The changes will involve:
+
+- Looping through the existing deal proposal flow three times.
+- Making parallel JSON‑RPC calls against individual storage provider servers.
+- Aggregating status feedback in the user interface and implementing a retry mechanism.
+- Provider selection process update - instead of the user manually selecting one provider, Delia will automatically select 3 different storage providers at random. Users will still have the option to replace any of them in the UI.
```
aidan46: I think with the deal parameters it could select an SP based on best price. Random is good for a PoC tho

rfcs/006-delia-replication.md
cargopete: ok, I think we can go ahead and agree on the framing in Slack then (`any actual implementation work for replication should be done after I try to move all of that into the FE itself then. Not a blocker for the RFC then.
`), wdyt?

rfcs/006-delia-replication.md
pete-eiger: agreed, we can change the criteria (i.e. set it to best price maybe) later but let's start with random

cargopete:
seems like the team is preferring the default "selection algorithm" to select for the cheapest N number (user provided value) of storage providers. This will be the default.

Of course if for some reason the user is not happy with this automatic selection, they will be able to select/deselect from the list of providers.

rfcs/006-delia-replication.md
cernicc: Agreed. We should probably change the next line  ```   - Encodes the deal proposal via the storage provider server’s HTTP endpoint (e.g. `/encode_proposal`).```

th7nder (approved):

jmg-duarte (approved):
LGTM

PR #777, opened 2025-02-24
feat(mater): replace the Reader/Writer structs with traits

jmg-duarte:
### Description

This is a "big" change with a simple core: replace the struct (which implemented a trait-like pattern) with (proper) traits.

This addresses a fundamental limitation of the crate that is: you either write *or* read.
One symptom of this was @cernicc (rightful) usage of "internal-er" free functions instead of using the reader/writer.


th7nder (approved):

cernicc (approved):
This feels a lot better. Nice

Issue #771, opened 2025-02-20
feat: Add deal parameters check to sp-server


aidan46:
Add a check that retrieves deal parameters information from the chain and compares the incoming deal with these parameters to check if the storage provider accepts them.

PR #778, opened 2025-02-25
feat: Add deal parameters

### LINKED ISSUE #770: feat: Add deal parameters to market pallet
Add an extrinsic for storage providers to publish deal parameters to the market pallet. This change should also include checking if incoming deals align with the published deal parameters.

aidan46:
### Description

This PR adds support for deal parameters in the market pallet.
2 new extrinsics have been added:
- `publish_deal_parameters`: Used by the SP to publish or update their deal parameters.
- `remove_deal_parameters`: Used by the SP to remove set deal parameters, I am not sure if this extrinsic is needed.

The deal parameters are stored in a [StorageMap](https://docs.rs/frame-support/latest/frame_support/storage/types/struct.StorageMap.html) and are checked before deal publishing.
Ideally, this check should always pass because the SP should check the parameters before publishing them, follow-up issue #771.

Tests have been added to cover deal parameter publishing and removal. Success and failure tests have been added to check deal parameter validation.

`storagext` has been updated to include:
- `publish_deal_parameters`
- `remove_deal_parameters`
- `retrieve_deal_parameters`

### Testing

- Build everything with `just build-binaries-all`
- Run the `start_sp.sh` script with `examples/start_sp.sh`
- Add deal parameters:
```shell
RUST_LOG=info ./target/release/storagext-cli --sr25519-key "//Charlie" market publish-deal-parameters --deal-parameters '{ "minimum_price_per_block": 5, "deal_duration": { "lower": 1, "upper": 100 }}'
```
- Query deal parameters:
```shell
RUST_LOG=info ./target/release/storagext-cli market retrieve-deal-parameters 5FLSigC9HGRKVhB9FiEo4Y3koPsNmBmLJbpXg2mp1hXcS59Y
DealParameters { minimum_price_per_block: 5, deal_duration: DealDurationBound { lower: 1, upper: 100 } }
```
- Remove deal parameters
```shell
RUST_LOG=info ./target/release/storagext-cli --sr25519-key "//Charlie" market remove-deal-parameters
```

### Checklist

  - [x] If yes, which ones? / List them here: #771


docs/src/architecture/pallets/market.md
```diff
@@ -176,6 +185,27 @@ storagext-cli --sr25519-key "//Alice" market settle-deal-payments 97 1010 1337 4
...
+| ------------------------- | --------------------------------------------------------------------------------- | -------------------------------------------------------- |
+| `minimum_price_per_block` | The minimum price per block the SP is willing to accept                           | Positive integer, in [Plancks](../../glossary.md#planck) |
+| `deal_duration`           | The optional lower and upper bound of a deal duration the SP is willing to accept | `DealDurationBound`                                      |
+
+Where `DealDurationBound` consists of:
+
+| Name  | Description                                        | Type             |
+| ----- | -------------------------------------------------- | ---------------- |
+| lower | Shortest deal duration the SP is willing to accept | Positive integer |
+| upper | Longest deal duration the SP is willing to accept  | Positive integer |
```
jmg-duarte:
```suggestion
| `lower` | Shortest deal duration the SP is willing to accept | Positive integer |
| `upper` | Longest deal duration the SP is willing to accept  | Positive integer |
```

pallets/market/src/lib.rs
```diff
@@ -804,6 +838,40 @@ pub mod pallet {
 
             Ok(())
         }
+
+        #[pallet::call_index(4)]
+        pub fn publish_deal_parameters(
```
jmg-duarte, line 837: Create the benchmarks for this too please

pallets/market/src/lib.rs
```diff
@@ -804,6 +838,40 @@ pub mod pallet {
...
+            SPDealParameters::<T>::insert(&provider, &deal_parameters);
+            Self::deposit_event(Event::<T>::DealParametersUpdated {
+                provider,
+                deal_parameters,
+            });
+            Ok(())
+        }
+
+        #[pallet::call_index(5)]
+        pub fn remove_deal_parameters(origin: OriginFor<T>) -> DispatchResult {
```
jmg-duarte, line 865: Same as above

pallets/market/src/lib.rs
```diff
@@ -804,6 +838,40 @@ pub mod pallet {
...
+            let provider = ensure_signed(origin)?;
+            ensure!(
+                T::StorageProviderValidation::is_registered_storage_provider(&provider),
+                Error::<T>::StorageProviderNotRegistered
+            );
+            // Check if params exist, remove and insert.
+            if SPDealParameters::<T>::contains_key(&provider) {
+                SPDealParameters::<T>::remove(&provider)
+            }
+            SPDealParameters::<T>::insert(&provider, &deal_parameters);
```
jmg-duarte: Couldn't this be a mutate?

pallets/market/src/lib.rs
```diff
@@ -1182,6 +1253,69 @@ pub mod pallet {
...
+                            log::error!(
+                                target: LOG_TARGET,
+                                "Deal duration for deal between {:?} and {:?} doesn't fall between the set bounds. {deal_duration:?} < {lower:?} || {deal_duration:?} > {upper:?}",
+                                proposal.provider,
+                                proposal.client
+                            );
+                            return Err(Error::<T>::InvalidDealParameters.into());
+                        }
+                    }
+                }
```
jmg-duarte: This should be encapsulated in a `impl DealParameter` and here it would just be `parameters.check_against(lower, upper)` or something similar

jmg-duarte (commented):
Overall, looks good!

Justfile
```diff
@@ -215,3 +217,16 @@ bench-node pallet steps="5" repeat="1":
...
+    cargo run \
+        -p polka-storage-node -r -F runtime-benchmarks -F testnet -- \
+        benchmark pallet \
+        --wasm-execution=compiled \
+        --pallet "pallet_{{pallet}}" \
+        --extrinsic "*" \
+        --steps "{{steps}}" \
+        --repeat "{{repeat}}" \
+        --template node/benchmark_template.hbs \
+        --output "{{output_file}}"
```
jmg-duarte: Maybe?
```suggestion
        --output "pallets/{{pallet}}/src/weights.rs"
```
Or so it's separated before you add it
```suggestion
        --output "pallets/{{pallet}}/weights.rs"
```

docs/src/architecture/pallets/market.md
```diff
@@ -2,15 +2,24 @@
...
-- [Extrinsics](#extrinsics)
-  - [`add_balance`](#add_balance)
-  - [`withdraw_balance`](#withdraw_balance)
-  - [`settle_deal_payments`](#settle_deal_payments)
-  - [`publish_storage_deals`](#publish_storage_deals)
-- [Events](#events)
-- [Errors](#errors)
-- [Constants](#constants)
+- [Market Pallet](#market-pallet)
+  - [Table of Contents](#table-of-contents)
```
jmg-duarte: These two are redundant

docs/src/architecture/pallets/market.md
```diff
@@ -176,6 +185,27 @@ storagext-cli --sr25519-key "//Alice" market settle-deal-payments 97 1010 1337 4
...
+
+| Name    | Description                                        | Type             |
+| ------- | -------------------------------------------------- | ---------------- |
+| `lower` | Shortest deal duration the SP is willing to accept | Positive integer |
+| `upper` | Longest deal duration the SP is willing to accept  | Positive integer |
+
+### `remove_deal_parameters`
+
+A signed extrinsic allows a storage provider to remove the deal parameter they have previously set.
+If no deal parameters are present this extrinsic will not fail.
```
jmg-duarte: This *might* be more accurate (you're the judge)
```suggestion
If no deal parameters are present this extrinsic is a no-op.
```

docs/src/architecture/pallets/market.md
```diff
@@ -176,6 +185,27 @@ storagext-cli --sr25519-key "//Alice" market settle-deal-payments 97 1010 1337 4
...
+| ------------------------- | --------------------------------------------------------------------------------- | -------------------------------------------------------- |
+| `minimum_price_per_block` | The minimum price per block the SP is willing to accept                           | Positive integer, in [Plancks](../../glossary.md#planck) |
+| `deal_duration`           | The optional lower and upper bound of a deal duration the SP is willing to accept | `DealDurationBound`                                      |
+
+Where `DealDurationBound` consists of:
+
+| Name    | Description                                        | Type             |
+| ------- | -------------------------------------------------- | ---------------- |
+| `lower` | Shortest deal duration the SP is willing to accept | Positive integer |
+| `upper` | Longest deal duration the SP is willing to accept  | Positive integer |
```
jmg-duarte: Maybe we should note that they're optional.

docs/src/architecture/pallets/market.md
```diff
@@ -176,6 +185,27 @@ storagext-cli --sr25519-key "//Alice" market settle-deal-payments 97 1010 1337 4
...
 [^settle_deal_payments]: Read more about the `settle-deal-payments` command in [_Storagext CLI/Subcommand `market`/`settle-deal-payments`_](../../storagext-cli/market.md#settle-deal-payments)
 
+### `publish_deal_parameters`
+
+This extrinsic lets the storage provider publish or update deal parameters.
+
+| Name                      | Description                                                                       | Type                                                     |
+| ------------------------- | --------------------------------------------------------------------------------- | -------------------------------------------------------- |
+| `minimum_price_per_block` | The minimum price per block the SP is willing to accept                           | Positive integer, in [Plancks](../../glossary.md#planck) |
+| `deal_duration`           | The optional lower and upper bound of a deal duration the SP is willing to accept | `DealDurationBound`                                      |
```
jmg-duarte: In tandem with the comment below, I believe this makes more sense
```suggestion
| `deal_duration`           | The lower and upper bounds of a deal duration the SP is willing to accept | `DealDurationBound`                                      |
```

docs/src/architecture/pallets/market.md
```diff
@@ -247,5 +277,5 @@ The Market Pallet actions can fail with following errors:
 | ------------------ | ----------------------------------------------------------------------------- | ------------------------- |
 | `MaxDeals`         | How many deals can be published in a single batch of `publish_storage_deals`. | 128                       |
 | `MaxDealsPerBlock` | Maximum deals that can be scheduled to start at the same block.               | 128                       |
-| `MinDealDuration`  | Minimum time an activated deal should last.                                   | 5 Minutes (50 Blocks)    |
+| `MinDealDuration`  | Minimum time an activated deal should last.                                   | 5 Minutes (50 Blocks)     |
```
jmg-duarte, line 278: Any chance we could get a md formatting tool to run so we stop having these?

pallets/market/src/benchmarking.rs
jmg-duarte, line 1: One thing left each:
* For the `publish` a replacement call would be great. You can do this using `Linear<1, 2>` and according to the number, publish and publish&replace
* For the `remove` is the same, remove with something there and with nothing.

The idea here is to cover both cases since they would have a different weight

pallets/market/src/lib.rs
```diff
@@ -300,6 +300,47 @@ pub mod pallet {
...
+                    return proposed_deal_duration > lower || proposed_deal_duration < upper
+                }
+            }
+        }
+
+        /// Validates that the price is higher than the SP set minimum.
+        fn validate_storage_price(&self, proposed_storage_price: Balance) -> bool {
+            return proposed_storage_price >= self.minimum_price_per_block;
+        }
+    }
```
jmg-duarte: This looks so neat, great job!

pallets/market/src/lib.rs
```diff
@@ -300,6 +300,47 @@ pub mod pallet {
...
+    #[derive(Clone, Eq, PartialEq, Encode, Decode, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+    pub struct DealParameters<Balance, BlockNumber> {
+        pub minimum_price_per_block: Balance,
+        pub deal_duration: DealDurationBound<BlockNumber>,
+    }
+
+    impl<Balance, BlockNumber> DealParameters<Balance, BlockNumber>
+    where
+        Balance: PartialOrd,            // For comparison
+        BlockNumber: PartialOrd + Copy, // Copy so we can match the Option
```
jmg-duarte: Instead of the comments you left, I'd write something like:
```suggestion
    // `Balance` and `BlockNumber` are not directly related to their `Config` counterparts,
    // the structure is generic enough that you could use it for other purposes,
    // hence we just restrict the generics to the strictly necessary subset of traits.
    //
    // Additional note: `BlockNumber` will usually be a number, especially in our network, 
    // as such, the `Copy` trait should come for free and not require any extra work on 
    // further implementations; furthermore, the actual `BlockNumber` does require `Copy`
    // https://docs.rs/sp-runtime/40.1.0/sp_runtime/traits/trait.BlockNumber.html
    impl<Balance, BlockNumber> DealParameters<Balance, BlockNumber>
    where
        Balance: PartialOrd,           
        BlockNumber: PartialOrd + Copy,
```

pallets/market/src/lib.rs
```diff
@@ -524,6 +581,8 @@ pub mod pallet {
         DealDurationOutOfBounds,
         /// Deal's piece_cid is invalid.
         InvalidPieceCid,
+        /// Deal's parameters do no fall between what the SP has set with `publish_deal_parameters`
+        InvalidDealParameters,
```
jmg-duarte: This is confusing on multiple levels.

* "Invalid parameters" would be closer to sending a string to a place expecting a number.
* The message seems targeted at who ever is pushing parameters, not "numbers being checked" for those parameters

I feel that something like this would be a bit more accurate:
```suggestion
        /// The proposed deal parameters' falls outside the parameter bounds set by the Storage Provider.
        OutOfBoundsDeal,
```

pallets/market/src/lib.rs
```diff
@@ -1112,6 +1206,9 @@ pub mod pallet {
                     return Err(e.into());
                 }
 
+                // Safety check on deal parameters, these should have been checked before publishing
```
jmg-duarte: Nit: I'm afraid the "should have been" may lead someone to mistakenly think this line isn't necessary 
```suggestion
                // Safety check on deal parameters, these should be checked by the submitter before publishing
```

pallets/market/src/lib.rs
```diff
@@ -1182,6 +1279,42 @@ pub mod pallet {
...
         }
 
+        /// Validates that proposals fall between the parameters set by the storage provider.
+        /// Checks for deal duration and minimum price.
+        /// returns Ok(()) if the proposal is valid, `Err(InvalidDealParameters)` if invalid.
+        fn validate_deal_parameters(
+            provider: &T::AccountId,
+            proposal: &DealProposal<T::AccountId, BalanceOf<T>, BlockNumberFor<T>>,
+        ) -> DispatchResult {
+            if let Some(params) = SPDealParameters::<T>::get(provider) {
```
jmg-duarte: Reducing nesting here is free & makes the code easier to read 
```suggestion
            let Some(params) = SPDealParameters::<T>::get(provider) else { return Ok(()) };
```

jmg-duarte (commented):
Looking great, just needs a bit of polish

Justfile
aidan46: I thought the sp pallet had different placement for the weights but I just noticed there are none there yet so we can place it in `pallets/{{pallet}}/src/weights.rs`

docs/src/architecture/pallets/market.md
aidan46: Auto formatting :angry:

docs/src/architecture/pallets/market.md
aidan46: [It is only a no-op if it returns an error](https://crates.parity.io/frame_support/macro.assert_noop.html). I did consider making this fail explicitly if there are no parameters present, might save on gas costs if SP double removes on accident.

docs/src/architecture/pallets/market.md
jmg-duarte: I know, I used to use the same plugin :P

docs/src/architecture/pallets/market.md
aidan46: [markdownlint](https://www.npmjs.com/package/markdownlint) is an option for this. My VSCode uses this for liting too

docs/src/architecture/pallets/market.md
jmg-duarte: Then, if there are no parameters, it replaces the existing ones with "empty ones"?

aidan46: No it doesn't replace, I attempts to [remove](https://docs.rs/frame-support/latest/frame_support/storage/types/struct.StorageMap.html#method.remove), which is infallible so I am assuming if nothing is present it does nothing, I didn't look into the implementation.

Justfile
```diff
@@ -215,3 +217,16 @@ bench-node pallet steps="5" repeat="1":
         --steps "{{steps}}" \
         --repeat "{{repeat}}" \
         --template node/benchmark_template.hbs
+
+# Generate the benchmark weights
+generate-weights pallet steps="5" repeat="1":
```
th7nder, line 222: Why are we adding this in this PR? Isn't there another command that does that already?

pallets/market/src/lib.rs
```diff
@@ -1182,6 +1324,44 @@ pub mod pallet {
...
+        /// returns Ok(()) if the proposal is valid, `Err(OutOfBoundsDeal)` if invalid.
+        fn validate_deal_parameters(
+            provider: &T::AccountId,
+            proposal: &DealProposal<T::AccountId, BalanceOf<T>, BlockNumberFor<T>>,
+        ) -> DispatchResult {
+            let Some(params) = SPDealParameters::<T>::get(provider) else {
+                return Ok(());
+            };
+            // Check that deal proposal falls between SPs set params
+            let deal_duration = proposal.end_block - proposal.start_block;
```
th7nder: We sure this never underflows?

th7nder (dismissed):

Justfile
aidan46: The command above just runs the benchmarks but doesn't update the weights in the pallet.

pallets/market/src/lib.rs
aidan46: If this underflows the [sanity check](https://github.com/eigerco/polka-storage/blob/c50ab0f4c2bf30ee6bd94ad6bebabf334e12b148/pallets/market/src/lib.rs#L1045-L1048) before the validation will catch it and reject the deal.

pallets/market/src/deal_parameters.rs
```diff
@@ -0,0 +1,189 @@
...
+        (Some(lower), Some(upper)) => DealDurationBound { lower, upper },
+    };
+    DealParameters {
+        minimum_price_per_block: offchain_params.minimum_price_per_block,
+        deal_duration: deal_duration,
+    }
+}
+
+// An attempt to make a generic DealParameters struct but because it needs to impl TypeInfo to
+// Be added in the StorageMap as a value because <T: Config> does not implement TypeInfo
```
jmg-duarte: Yeah, I had similar issues before. Don't worry about it

pallets/market/src/deal_parameters.rs
```diff
@@ -0,0 +1,189 @@
...
+            && self.deal_duration.lower < self.deal_duration.upper
+            && (self.deal_duration.lower..=self.deal_duration.upper).contains(&chain_min_duration)
+            && (self.deal_duration.lower..=self.deal_duration.upper).contains(&chain_max_duration)
+    }
+}
+
+/// Converts [`OffchainDealParameters`] to [`DealParameters`]
+/// and fills in the `None` with the given chain constants.
+/// Conversion function instead of a `From` implementation because
+/// we need the chain constants in [`crate::Config`]
```
jmg-duarte: This isn't a doc comment
```suggestion
// Conversion function instead of a `From` implementation because
// we need the chain constants in [`crate::Config`]
```

pallets/market/src/deal_parameters.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub fn validate(
+        &self,
+        chain_min_duration: BlockNumber,
+        chain_max_duration: BlockNumber,
+    ) -> bool {
+        Balance::zero() < self.minimum_price_per_block
+            && self.deal_duration.lower < self.deal_duration.upper
+            && (self.deal_duration.lower..=self.deal_duration.upper).contains(&chain_min_duration)
+            && (self.deal_duration.lower..=self.deal_duration.upper).contains(&chain_max_duration)
+    }
```
jmg-duarte: You're separating two steps that don't need to be separated — the conversion from "Offchain" to "Onchain" must be done in a single swift swoop

If you do the following:
```
impl OffchainPricingParams {
    fn validate(self, min, max) -> Result<OnChain, Error> {
        let min = self.min.unwrap_or(min);
		let max = self.max.unwrap_or(max);
		if min > max { return Error }
		return Ok(Onchain { min, max })
    }
}
```
You can always construct valid OnChain variants, because then you can do
```
impl OffchainDealParams {
    fn validate(self, min, max) -> Result<OnChain, Error> {
        if self.price <= 0 { return Error }
		return OnChain { price: self.price, duration: self.duration.validate(min, max)? }
    }
}
```

The key idea is ["making invalid states unrepresentable"](https://geeklaunch.io/blog/make-invalid-states-unrepresentable/).

jmg-duarte (commented):
Super close, I'm clearly not being clear enough about the offchain -> onchain idea, hope this helps!

jmg-duarte (dismissed):
Good job mate!

th7nder (approved):

cernicc (approved):

PR #785, opened 2025-02-27
feat: refactor p2p node & merge retrieval functionality

cernicc:
### Description

The PR looks bigger than it is. It refactors the p2p part of the storage provider node and merges retrieval server functionality to the same `Behaviour`. That was needed because, when the provider node registers its public multiaddress to the rendezvous node. That multiaddress is returned by the rendezvous node when the client requests a peer_id to multiaddress mapping.

This brings us a step closer to resolving the content based on the deal_id. Next part is indexing deals with cooresponding payload cids. That mapping should probably be exposed over request_response protocol. The p2p protocol for that is needed because client will already be connected to the provider's node before starting the block retrieval. --- The missing part for the client is figuring out which block to request first.

### Checklist

            **Many. This is the least ugly one.**


storage-fetch/src/client.rs
```diff
@@ -21,17 +21,16 @@ use tokio::{
 };
 use tracing::{debug, error, info, instrument, trace};
 
-use crate::{new_swarm, Behaviour, BehaviourEvent, InitSwarmError};
+use crate::p2p::{new_swarm, Behaviour, BehaviourEvent, InitSwarmError};
+
+// use crate::p2p::{new_swarm, Behaviour, BehaviourEvent, InitSwarmError};
```
aidan46: A leftover comment

storage-provider/server/src/p2p/mod.rs
```diff
@@ -1,52 +1,265 @@
...
+                tracing::info!("P2P worker completed");
+            }
+        }
+
+        Ok(())
+    }))
+}
+
+/// Arguments used to configure the [`P2p`].
+pub struct P2pArgs<B>
```
aidan46, line 65: Shouldn't we add the `registration_ttl` to this too? Or just set 24 hours for all registrations?

storage-provider/server/src/p2p/mod.rs
```diff
@@ -1,52 +1,265 @@
...
-pub(crate) struct P2PState {
-    /// P2P ED25519 private key
-    pub(crate) p2p_key: Keypair,
+/// Dials rendezvous nodes. The `Ok` indicates that we successfully started a
+/// dialing procedure with the nodes. It doesn't indicate that we successfully
+/// connected to the nodes.
+fn dial_rendezvous_nodes<B>(
+    swarm: &mut Swarm<Behaviour<B>>,
+    nodes: &[(PeerId, Multiaddr)],
+) -> Result<Vec<(PeerId, Multiaddr)>, P2pError>
```
aidan46: You do not seem to use this result. Was the intention to store the available bootstrap nodes back into the worker?

storage-provider/server/src/p2p/mod.rs
```diff
@@ -1,52 +1,265 @@
...
 
-    /// PeerID of the bootstrap node used by the registration node.
-    /// Optional because it is not used by the bootstrap node.
-    pub(crate) rendezvous_point: PeerId,
+/// Request registration with the rendezvous nodes.
+fn request_registration<B>(swarm: &mut Swarm<Behaviour<B>>, nodes: &[(PeerId, Multiaddr)])
+where
+    B: Blockstore,
+{
+    for (rendezvous_peer, rendezvous_addr) in nodes {
```
aidan46, line 243: I don't think we need to register with every bootstrap node. The bootstrap nodes share this information with each other. A better approach might be, if you are passing in bootstrap nodes, is to implement registration retry if it fails on the first bootstrap node.

storage-provider/server/src/p2p/mod.rs
cernicc: I didn't see any point of customizing the registration ttl. The [REGISTRATION_TTL](https://github.com/eigerco/polka-storage/blob/b92cd122d45092d3be0942559f2e37966731d229/storage-provider/server/src/p2p/mod.rs#L26) is used

storage-provider/server/src/p2p/mod.rs
cernicc: It is used [here](https://github.com/eigerco/polka-storage/blob/b92cd122d45092d3be0942559f2e37966731d229/storage-provider/server/src/p2p/mod.rs#L125). This fails in cases when we can't start connecting to any of the nodes. If Ok is returned, that doesn't mean that the nodes are available. It only means that we can start connecting to them.

storage-provider/server/src/p2p/mod.rs
cernicc: I see your point. The thing is, that in this case we would need to track the registration state in the worker. We'll probably have to implement that anyway. I just didn't want for the pr to get even bigger. But yeah. Ideally, the worker would track registration state and react on that.

storage-provider/server/src/p2p/blockstore/chaos.rs
jmg-duarte, line 1: You could #[cfg(test)] this or something

storage-provider/server/src/p2p/mod.rs
```diff
@@ -1,52 +1,264 @@
...
+            _ => {}
+        }
+    }
 
-    /// TTL of the p2p registration in seconds
-    pub(crate) registration_ttl: u64,
+    #[instrument(level = "trace", skip(self))]
+    fn on_identify_event(&mut self, event: IdentifyEvent) {
+        match event {
+            // once `/identify` did its job, we know our external address
```
jmg-duarte: our? Isn't it the *other* peer?

storage-provider/server/src/p2p/mod.rs
```diff
@@ -1,52 +1,264 @@
...
+fn dial_rendezvous_nodes<B>(
+    swarm: &mut Swarm<Behaviour<B>>,
+    nodes: &[(PeerId, Multiaddr)],
+) -> Result<Vec<(PeerId, Multiaddr)>, P2pError>
+where
+    B: Blockstore,
+{
+    let mut dialling = vec![];
+    for (rendezvous_peer, rendezvous_addr) in nodes {
+        // Dial the the node if needed
```
jmg-duarte:
```suggestion
        // Dial the node if needed
```

storage-provider/server/src/p2p/mod.rs
```diff
@@ -1,52 +1,264 @@
...
+        if !swarm.is_connected(rendezvous_peer) {
+            if let Err(err) = swarm.dial(rendezvous_addr.clone()) {
+                warn!(?err, %rendezvous_peer, %rendezvous_addr, "rendezvous node dialing error");
+                continue;
+            }
+        }
+
+        // Add successful dials to available nodes
+        dialling.push((*rendezvous_peer, rendezvous_addr.clone()));
```
jmg-duarte: If we're connected, won't it add a duplicate?

storage-provider/server/src/p2p/mod.rs
jmg-duarte, line 1: I feel that the worker needs a few docs explaining what we're expecting it to happen at a network level // between peers, it's not entirely clear when reading the code

storage-provider/server/src/p2p/blockstore/chaos.rs
cernicc: It's [here](https://github.com/eigerco/polka-storage/blob/dafafd447f1db5aac2ec7d1be18c20ac3e907dad/storage-provider/server/src/p2p/blockstore/mod.rs#L1)

storage-provider/server/src/p2p/mod.rs
cernicc: sp node first connects to the rendezvous node. Then sp node identifies with the rendezvous node. The rendezvous node identifies with the sp node too. Soo. When our node receives the `IdentifyEvent::Received`, this is a message received from the rendezvous node. The `observed_addr` in the message is set by the rendezvous node and indicates the multiaddr that was observed by the rendezvous node when we connected. The `observed_addr` is later used to register with the rendezvous nodes.

storage-provider/server/src/p2p/mod.rs
cernicc: It would yeah. But in this case it wouldn't matter.
I removed the part of returning an error if we couldn't dial the nodes. I'll add a proper failure handling in the next pr.

storage-provider/server/src/p2p/mod.rs
cernicc: I've refactored this part a little bit. I'll add a proper error handling in the next PR.

storage-provider/server/src/p2p/mod.rs
cernicc: I've added a comment. Please let me know if you have something else in mind.

storage-provider/server/src/p2p/blockstore/chaos.rs
jmg-duarte: Ah ok, but see, it isn't clear cut :P

storage-provider/server/src/p2p/mod.rs
```diff
@@ -57,44 +261,10 @@ pub(crate) fn deser_keypair<'de, D: de::Deserializer<'de>>(d: D) -> Result<Keypa
     keypair_value_parser(&src).map_err(de::Error::custom)
 }
 
-/// Parses a string to an optional Peer ID.
-/// Used in the [`ConfigurationArgs`] rendezvous_point field.
+/// Parses a string to an Peer ID.
```
jmg-duarte: It's not "in-scope" but it's "making this file better" if you change these two that are not used here, to where they're used (`config.rs`)

storage-provider/server/src/p2p/mod.rs
jmg-duarte: We resolved this on Slack, but I think we can briefly cover why it's "our" address going here, just so it makes it easier for anyone reading this code without a bunch of libp2p experience

storage-provider/server/src/p2p/mod.rs
```diff
@@ -1,52 +1,256 @@
...
+    let worker = Worker::new(args)?;
+
+    Ok(tokio::spawn(async move {
+        tokio::select! {
+            _ = cancellation_token.cancelled() => {
+                tracing::info!("P2P worker received shutdown signal");
+            }
+            _ = worker.run() => {
+                tracing::info!("P2P worker completed");
+            }
```
jmg-duarte: You're ignoring the returned error by `worker.run()`, at the very least, do something like:
```suggestion
            result = worker.run() => {
            	match result {
            		Ok(_) => tracing::info!("P2P worker completed"),
            		Err(err) => tracing::error!("P2P failed with error: {}", err);
            	}
            }
```

storage-provider/server/src/p2p/mod.rs
cernicc: I've added a longer explanation.

storage-provider/server/src/p2p/mod.rs
cernicc: Nice one. Much prettier

storage-provider/server/src/p2p/mod.rs
aidan46: Why do we need to track the registration state in the worker? It can just assume that it needs to re-register when `REGISTRATION_TTL` is over. Maybe I don't understand what you mean

cernicc: Ask yourself, where can the registration fail and where (in code) and how should we retry the registration if it fails.

Some of the cases, when it can fail for a single rendezvous node are:
1. Failure of dialing
2. Failure of connecting
3. Failure when identifying (this is fine if we already know the external address)
4. Failure when registering.

[This](https://github.com/eigerco/polka-storage/blob/8dcdbc2277f49502e6e8247168a40abc6f75af0b/storage-provider/server/src/p2p/mod.rs#L245) register call returns an `Err` only if the registration request failed. It doesn't mean that the registration was successful or not. You can only be sure of outcome when you receive `RendezvousEvent::Registered` or `RendezvousEvent::RegisterFailed`.

storage-provider/server/src/p2p/blockstore/chaos.rs
cernicc: I've also added `#![cfg(test)]` into the module. Thanks for the suggestion.

jmg-duarte (approved):

storage-provider/server/src/p2p/mod.rs
aidan46: Thank you for the explanation :pray:

aidan46 (approved):

PR #788, opened 2025-03-01
fix(sp-server): cancel & stop tasks on failure

jmg-duarte:
### Description

Moves away from `join!` to `JoinSet` since the former blocks waiting for *all* tasks to complete, yet we (for now) should shutdown the server on error instead of limping with invalid tasks.

Notes: 
* The example that prompted this (binding to the same port) still "hangs" due to the pipeline having some tasks pending/running, so it takes a while to close everything down.
* I know the `.map(|result| ("RPC", result.map_err(ServerError::from))),` looks kind of weird but it was the way I found to distinguish tasks from each other — open to suggestions.
* I'm also not super pleased with the final error return, if anyone has ideas, please share.

### DIFF
--- a/storage-provider/server/src/main.rs
+++ b/storage-provider/server/src/main.rs
@@ -16,12 +16,13 @@ use std::{
 };
 
 use clap::Parser;
+use futures::FutureExt;
 use indexer::{
     local_index_directory::rdb::{RocksDBLid, RocksDBStateStoreConfig},
     start_indexer, IndexerMessage, IndexerState,
 };
 use libp2p::{identity::Keypair, Multiaddr, PeerId};
-use p2p::{blockstore::PiecesBlockstore, start_p2p, P2pArgs, P2pError};
+use p2p::{blockstore::PiecesBlockstore, P2pArgs, P2pError};
 use pipeline::types::PipelineMessage;
 use polka_storage_proofs::{
     porep::{self, PoRepParameters},
@@ -41,7 +42,7 @@ use storagext::{
 use subxt::{self, tx::Signer};
 use tokio::{
     sync::{mpsc::UnboundedReceiver, Mutex, Semaphore},
-    task::JoinError,
+    task::{JoinError, JoinSet},
 };
 use tokio_util::sync::CancellationToken;
 use tracing::level_filters::LevelFilter;
@@ -191,30 +192,6 @@ pub enum ServerError {
 
     #[error(transparent)]
     Lid(#[from] crate::indexer::local_index_directory::LidError),
-}
-
-/// Takes an expression that returns a `Result<Result<T, E2>, E1>`.
-/// It tries to inspect and log the first error (`E1`), otherwise,
-/// it inspects the result and tries to inspect the nested error (`E2`).
-///
-/// This macro is *roughly* equivalent to calling:
-/// ```text
-/// res // : Result<Result<T, E2>, E1>
-///     .inspect_err(|e| tracing::error!(%e))
-///     .inspect(|r| r.inspect_err(|e| tracing::error!(%e))
-/// ```
-macro_rules! inspect_and_log_nested_errors {
-    ($($task:expr),+ $(,)?) => {
-        (
-            $(
-                $task
-                    .inspect_err(|err| tracing::error!(%err))
-                    .inspect(|ok| {
-                        let _ = ok.as_ref().inspect_err(|err| tracing::error!(%err));
-                    })
-            ),+
-        )
-    };
 }
 
 /// The server arguments, as passed by the user, unvalidated.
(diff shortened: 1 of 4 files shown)


storage-provider/server/src/main.rs
```diff
@@ -390,63 +367,77 @@ impl Server {
...
+                                    tracing::error!("Cancelling remaining tasks...");
+                                    if error.is_none() {
+                                        error = Some(err);
+                                    }
+                                    cancellation_token.cancel();
+                                },
+                            }
+                        }
+                        Some(Err(err)) => {
+                            tracing::error!("Failed to join task with error: {err}");
```
th7nder, line 433: Why are we ignoring this error with logging and not saving it to `error = `?

storage-provider/server/src/main.rs
```diff
@@ -390,63 +367,77 @@ impl Server {
...
+                        Some(Ok((task_name, task_result))) => {
+                            match task_result {
+                                Ok(()) => tracing::info!("{task_name} finished successfully!"),
+                                Err(err) => {
+                                    tracing::error!("{task_name} finished with error: {err}");
+                                    tracing::error!("Cancelling remaining tasks...");
+                                    if error.is_none() {
+                                        error = Some(err);
+                                    }
+                                    cancellation_token.cancel();
```
th7nder, line 428: Which branches of the match statement are executed if one task fails and others are cancelled?
Can it be like that, that two fail, other ones are cancelled but we're saving an error only from the one?

storage-provider/server/src/p2p/mod.rs
```diff
@@ -161,6 +133,10 @@ where
                     request_registration(&mut self.swarm, &self.rendezvous_nodes);
                 }
                 event = self.swarm.select_next_some() => self.on_swarm_event(event),
+                _ = cancellation_token.cancelled() => {
+                    info!("P2P worker received shutdown signal");
```
th7nder, line 137: Hm, so when it's cancelled it's not an error even though it could be in the process of handling something?

storage-provider/server/src/pipeline/mod.rs
```diff
@@ -278,14 +278,14 @@ fn process(
 ) {
     match msg {
         PipelineMessage::AddPiece(msg) => {
-            tracker.add_piece(state.clone(), msg, token.clone());
+            tracker.add_piece(state.clone(), msg, token.child_token());
```
th7nder, line 281: Yeah, good stuff. `child_token` feels better here.

storage-provider/server/src/main.rs
cernicc: What do you mean by, ignoring the error?
Edit: I think I know what you meant

storage-provider/server/src/main.rs
```diff
@@ -390,63 +367,77 @@ impl Server {
...
+                            // This branch should run when all tasks have terminated,
+                            // thus, this is the most appropriate place to return the value
+                            match error {
+                                Some(err) => return Err(err),
+                                None => return Ok(()),
+                            }
+                        },
+                    }
+                }
+                result = tokio::signal::ctrl_c() => {
```
cernicc: I think I've added the original impl and it's wrong. The ctrl_c is listening for the SIGINT signal not SIGTERM. For the more future proof signal handling the solution below should be better. It captures SIGINT and SIGTERM. It's good to capture the SIGTERM too, so that the process is also gracefully shutdown when running in a container.
 
```suggestion
/// Wait for a shutdown signal.
async fn wait_for_shutdown_signal() {
    use tokio::signal::unix::{signal, SignalKind};

    // Infos here:
    // https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html
    let mut signal_terminate = signal(SignalKind::terminate()).expect("should be able to listen for a SIGTERM");
    let mut signal_interrupt = signal(SignalKind::interrupt()).expect("should be able to listen for a SIGINT");

    tokio::select! {
        _ = signal_terminate.recv() => tracing::debug!("Received SIGTERM."),
        _ = signal_interrupt.recv() => tracing::debug!("Received SIGINT."),
    };
}
```

storage-provider/server/src/main.rs
jmg-duarte: We're only saving the first error, yes. But we're logging them all.

If the tasks in the set finish, either successfully or with an error they should be returned by `join_next`. 
Thus, they should all be handled by the respective branches.

storage-provider/server/src/main.rs
jmg-duarte: Good catch. This should also set the error if it is None

storage-provider/server/src/p2p/mod.rs
jmg-duarte: The cancellation itself is just a signal, but the way we have libp2p implemented. But there's little to do in the libp2p case if it was handling other work

jmg-duarte: cc @cernicc @aidan46 you guys touched libp2p way more than I did, maybe you can answer properly

th7nder (approved):

storage-provider/server/src/p2p/mod.rs
cernicc: I haven't looked much at gracefully shutting down the swarm instance. I think it should be possible by storing the peers connected with their corresponding `ConnectionId`'s. When the cancelation is received, we would have to close each connection by calling [close_connection](https://docs.rs/libp2p/latest/libp2p/struct.Swarm.html#method.close_connection). Wait for all connections to be closed and notify the requestor that the swarm was shut down.

aidan46 (approved):

PR #792, opened 2025-03-03
fix: Storagext deal param types

aidan46:
### Description

Small PR to fix the types in storagext for the deal parameters and the scale file. I forgot to update those after refactoring the Offchain/Onchain deal parameters types.

This is a fix done before #771 in a separate PR to reduce the upcoming PR size.


cernicc (approved):

jmg-duarte (approved):
Happens to the best

PR #794, opened 2025-03-03
fix: set 1GiB verifying keys in Genesis Config

### LINKED ISSUE #793: Cannot set PoSt verifying key as it halts block production

th7nder:
### Description

We couldn't set 1GiB PoSt verifying keys in the `set_post_verifying_key` extrinsic as it was halting the block production (1MiB in StorageMap was tooo big).
As this is kind of one-off, I made it happen in the genesis config.
It'll allow me to continue with running the full pipeline on the worker and benchmark PoSts.

Fixes #793.

### Checklist

  - [X] If yes, which ones?


lib/polka-storage-proofs/src/groth16/mod.rs
```diff
@@ -269,6 +270,59 @@ where
...
+
+impl Serialize for VerifyingKey<Bls12> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let size = self.serialised_bytes();
+        let mut buf = Vec::with_capacity(size);
+        // SAFETY: with_capacity(size) guarantees safety of buf.set_len(size)
+        unsafe {
```
cernicc: Why did you choose the `set_len` instead of the other safe functions?

lib/polka-storage-proofs/Cargo.toml
```diff
@@ -21,7 +21,9 @@ bellperson = { workspace = true, optional = true }
 blstrs = { workspace = true, optional = true }
 filecoin-hashers = { workspace = true, optional = true }
 filecoin-proofs = { workspace = true, optional = true }
+hex = { workspace = true }
 rand = { workspace = true, default-features = false, optional = true }
+serde = { workspace = true, features = ["alloc"] }
```
cernicc, line 27: Should we enable the `alloc` only when the `std` is enabled?

th7nder: Nope, we need in in `no_std` context because of the serialization of `GenesisConfig`. 
https://forum.polkadot.network/t/genesisconfig-in-wasm-runtime-a-step-towards-a-non-native-future/2887

lib/polka-storage-proofs/src/groth16/mod.rs
th7nder: Thought it'd be faster, but now that clippy complains as well, let me make it to resize.

aidan46 (commented):
If I understand correctly, this change is a band-aid to set 1GiB verifying key. At some point the chain would need to support this too. Any ideas on how we can support the 1GiB verifying key without halting block production? Will these change be reverted after you run the pipeline on the worker?

th7nder:
> If I understand correctly, this change is a band-aid to set 1GiB verifying key. At some point the chain would need to support this too. Any ideas on how we can support the 1GiB verifying key without halting block production? Will these change be reverted after you run the pipeline on the worker?

That's true, it's kind-of a band-aid. But honestly, there is no other way to set it currently 'easily'. When we go prod, we'll generate those params off-chain and set them in the genesis block as well, setting the params is a one-off.

If we wanted to ever change big params, we'd probably to chunk the VerifyingKey across multiple extrinsics which is possible, but requires more effort.

aidan46 (dismissed):

lib/polka-storage-proofs/Cargo.toml
cernicc: Hmm. Should we write a comment explaining why this is needed? I am thinking if somebody looks at the dependencies and doesn't know this specific.

cernicc (approved):

aidan46 (approved):

PR #796, opened 2025-03-05
feat: deal_id content retrieval

cernicc:
### Description

Add ability to retrieve the content for a specific deal id.

The polka-fetch now contains two commands.

When we already know the multiaddress of the storage provider and payload cid of the content, we can use:
`cargo run --bin polka-fetch -- --output ./download.car --overwrite by-payload-cid --provider /ip4/127.0.0.1/tcp/8002 --payload-cid bafybeiefli7iugocosgirzpny4t6yxw5zehy6khtao3d252pbf352xzx5q`

If we don't know the provider and payload cid. We can use the deal_id.
`cargo run --bin polka-fetch -- --output ./download.car --overwrite by-deal-id --deal-id=0 --bootstrap-address /ip4/127.0.0.1/tcp/62649 --bootstrap-peer 12D3KooWGtBNVq3GKPQLWyyKbvrnbBZnUtQbWwwGWiAwSATAHQPN --parachain-address ws://127.0.0.1:42069`

### Checklist


examples/storage_provider_private.pem
aidan46, line 1: We shouldn't add any keys into the repo. It is better to generate them on the fly.

cernicc: These are example keys. I think we should add more keys to the repo :D e.g. If you try to execute the command in the description. It will fail, because the peer id of the bootstrap node changes each run. It's easier to give examples if the keys are pre-generated.

storage-fetch/src/peer_resolver.rs
```diff
@@ -0,0 +1,111 @@
...
+                request_response::Config::default(),
+            )
+        })?
+        .with_swarm_config(|cfg| cfg.with_idle_connection_timeout(Duration::from_secs(10)))
+        .build();
+    Ok(swarm)
+}
+
+/// Run the discovery swarm and request the peer ID to multiaddrs mapping.
+async fn run_discover(
```
aidan46: This might be a bit nit-picky but I think the `run_discover` is a bit of a misnomer. We are just requesting a mapping from the bootstrap node and not actually discovering.

examples/start_sp.sh
```diff
@@ -11,17 +11,17 @@ mkdir -p /tmp/polka-storage-provider
...
-# Generate ED25519 private key
-openssl genpkey -algorithm ED25519 -out "$P2P_PRIVATE_KEY"
-# -outpubkey is only available in OpenSSL 3.4.0 onwards
-# https://github.com/openssl/openssl/commit/6c03fa21ed4bbc9fd6d3013fdf9f4646d231f831
-openssl pkey -in "$P2P_PRIVATE_KEY" -pubout -out "$P2P_PUBLIC_KEY"
+# # Generate ED25519 private key
+# openssl genpkey -algorithm ED25519 -out "$P2P_PRIVATE_KEY"
+# # -outpubkey is only available in OpenSSL 3.4.0 onwards
+# # https://github.com/openssl/openssl/commit/6c03fa21ed4bbc9fd6d3013fdf9f4646d231f831
+# openssl pkey -in "$P2P_PRIVATE_KEY" -pubout -out "$P2P_PUBLIC_KEY"
```
jmg-duarte: Not the biggest fan of removing this, maybe at least leave a comment explaining why this is commented out?

storage-fetch/src/download.rs
```diff
@@ -59,32 +81,32 @@ impl ClientSettings {
...
+/// A Downloadclient is used to download blocks from the storage provider. Single client
 /// supports getting a single payload.
-pub struct Client {
-    settings: ClientSettings,
+pub struct DownloadClient {
+    settings: DownloadClientSettings,
 
     /// Providers of data
-    providers: Vec<Multiaddr>,
+    providers: Vec<(PeerId, Multiaddr)>,
```
aidan46: Aren't we connecting to a single storage provider? I get that a storage provider might have multiple `Multiaddr`'s but they would still have a single `PeerId`

storage-fetch/src/download.rs
```diff
@@ -10,25 +11,36 @@ use cid::Cid;
...
 /// Errors that can occur while retrieving some content.
 #[derive(Debug, Error)]
-pub enum ClientError {
-    /// Error occurred while initialing swarm
-    #[error("Swarm initialization error: {0}")]
-    InitSwarm(#[from] InitSwarmError),
+pub enum DownloadError {
+    /// Is used when the storage provider doesn't have a piece requested
+    #[error("Unknown piece")]
+    PieceUnknown,
```
jmg-duarte: This error doesn't seem to match the doc. Shouldn't it be `NoPieceRequested`? Or maybe the doc should instead state `When an unknown piece is requested` or something?

storage-fetch/src/download.rs
```diff
@@ -43,13 +55,23 @@ pub enum ClientError {
     Blockstore(#[from] blockstore::Error),
 }
 
-pub struct ClientSettings {
+/// Behaviour used by the download client.
+#[derive(NetworkBehaviour)]
+pub struct Behaviour<B>
+where
+    B: Blockstore + 'static,
```
jmg-duarte: Do we need to apply the bound here or could we only add it to the `impl`?

storage-fetch/src/download.rs
```diff
@@ -59,32 +81,32 @@ impl ClientSettings {
     }
 }
 
-/// A client is used to download blocks from the storage provider. Single client
+/// A Downloadclient is used to download blocks from the storage provider. Single client
```
jmg-duarte:
```suggestion
/// A [`DownloadClient`] is used to download blocks from the storage provider. Single client
```

Furthemore, the sentence `Single client supports getting a single payload.` feels weird. Do you mean something among the lines of `Currently, the client only supports downloading a file per "session"?`

storage-fetch/src/download.rs
```diff
@@ -112,20 +134,27 @@ impl Client {
...
     }
 
     /// Start download of some content with a payload cid.
-    pub async fn download(mut self) -> Result<(), ClientError> {
+    pub async fn download(mut self, &piece_cid: &Cid) -> Result<(), DownloadError> {
         // Dial all providers
-        for provider in self.providers.clone() {
-            self.swarm.dial(provider)?;
-        }
+        for (peer, multiaddr) in self.providers.clone() {
```
aidan46: There should be some error handling in case `self.providers` is an empty vector. Could be done here or in the new function.

storage-fetch/src/download.rs
```diff
@@ -59,32 +81,32 @@ impl ClientSettings {
...
     /// The in flight block queries. If empty we know that the client received
     /// all requested data.
     queries: HashMap<QueryId, Cid>,
     /// Blockstore used by the client to store blocks into.
     blockstore: ReadWriteBlockstore<File>,
-    /// Content roots being downloaded.
-    root: Cid,
+    /// Content roots being downloaded. The root is set when we receive a
+    /// successful response from the storage providers.
+    root: Option<Cid>,
```
jmg-duarte: * This can't be "roots" since the type is `Option<Cid>`
* "The root is set when we receive a successful response..." implies the code is written in such a way that we're constantly replacing the root

storage-fetch/src/download.rs
jmg-duarte: By this point the number of providers should have already been validated to be non-empty though

storage-fetch/src/download.rs
```diff
@@ -137,7 +166,7 @@ impl Client {
 
             // if no inflight queries, that means we received
             // everything requested. Finalize the blockstore.
-            if self.queries.is_empty() {
+            if self.root.is_some() && self.queries.is_empty() {
```
jmg-duarte: This root check here is throwing me off, maybe I'm missing something but this could probably benefit from a brief explanation

storage-fetch/src/download.rs
```diff
@@ -172,7 +201,10 @@ impl Client {
 
             FileReader::new(file)
                 .await?
-                .copy_tree(&self.root, &mut extracted_file)
+                .copy_tree(
+                    &self.root.expect("root should be known"),
```
jmg-duarte: Maybe you could check the root with a `let` guard and return a WrongNumberOfRoots or similar

examples/start_sp.sh
cernicc: I've moved the key generation to a separate script. The [answer](https://github.com/eigerco/polka-storage/pull/796#discussion_r1981051454) also partially answers this.

storage-fetch/src/download.rs
```diff
@@ -234,7 +269,41 @@ impl Client {
     }
 
     #[instrument(level = "trace", skip(self))]
-    async fn on_bitswap_event(&mut self, event: Event) -> Result<(), ClientError> {
+    fn on_request_response(
+        &mut self,
+        event: request_response::Event<PieceInfoRequest, PieceInfoResponse>,
+    ) -> Result<(), DownloadError> {
+        if let request_response::Event::Message { peer, message } = event {
+            if let Message::Response { response, .. } = message {
```
jmg-duarte: I find it sad that let chains [haven't been stabilized yet](https://github.com/rust-lang/rust/pull/132833)

examples/storage_provider_private.pem
aidan46: With this change the bootstrap peer ID would still change every time you run the node since the keys are generated in the Justfile rule before running the node, it is the storage provider info that would stay the same.
I get that it is a PITA to work with keys that change every run but we have built the tools to extract the peer ID from these keys, e.g. `polka-storage-provider-client generate-peer-id --pubkey "$P2P_BOOTSTRAP_PUBLIC_KEY"`

storage-fetch/src/peer_resolver.rs
cernicc: I agree. The peer_resolver contents were mostly copied from the example you provided. Nevertheless, I am not sure what  would be a better name. Suggestions are appreciated

storage-fetch/src/download.rs
aidan46: You're right, that mapping request will fail if the given peer ID does not match any in the bootstraps node registry.

storage-fetch/src/peer_resolver.rs
```diff
@@ -0,0 +1,111 @@
...
+    let swarm = create_swarm()?;
+
+    let peer_info = run_discover(swarm, bootstrap_address, &bootstrap_peer_id, sp_peer_id)
+        .await
+        .unwrap();
+
+    match peer_info {
+        PeerInfoResponse::NotFound(peer) => {
+            warn!(?peer, "peer is not registered");
+            bail!("{:?} is not registered", peer)
```
jmg-duarte: The `bail` in this precise spot is slightly more confusing because it adds another layer of indirection in the already final return. Normally, you'd use it for early returns instead
```suggestion
            Err(anyhow!("{:?} is not registered", peer))
```
In contrast, the bails in lines 91 and 99 are better usage examples

storage-fetch/src/peer_resolver.rs
aidan46: Something along the lines of `request_peer_id_to_mulitaddr_mapping` or `request_storage_provider_info` would be appropriate imo

storage-fetch/src/download.rs
cernicc: Technically yes. Because we are getting providers from the deal, and each deal can only have a single provider. But from the `DownloadClient` point of view. It can download content from multiple providers. It doesn't care from where the list comes from.

storage-provider/server/src/p2p/mod.rs
```diff
@@ -46,6 +53,8 @@ where
     pub listen_on: Vec<Multiaddr>,
     /// The blockstore used for content retrieval.
     pub blockstore: Arc<B>,
+    /// Index db
```
jmg-duarte:
```suggestion
    /// Piece index database.
```

storage-provider/server/src/p2p/mod.rs
```diff
@@ -193,6 +217,56 @@ where
...
+                        .behaviour_mut()
+                        .request_response
+                        .send_response(channel, response)
+                        .is_err()
+                    {
+                        error!("Failed to send piece info to {peer:?}");
+                    }
+                }
+            }
+            request_response::Event::OutboundFailure {
```
jmg-duarte: The mix between qualified and non-qualified looks weird

storage-provider/server/src/indexer/mod.rs
```diff
@@ -104,7 +105,7 @@ where
 }
 
 /// Prepares indexes of a raw piece.
-async fn piece_indexes<P>(location: P) -> Result<Vec<IndexRecord>, ServerError>
+async fn piece_indexes<P>(location: P) -> Result<(Vec<Cid>, Vec<IndexRecord>), ServerError>
```
jmg-duarte, line 109: Can you document the return type?

storage-provider/server/src/indexer/local_index_directory/rdb.rs
```diff
@@ -601,6 +601,7 @@ impl Service for RocksDBLid {
     fn add_index(
         &self,
         piece_cid: Cid,
+        roots: Vec<Cid>,
```
jmg-duarte, line 604: This diverges from the original reference, could you add the usefulness of the `roots` to the docstring?

storage-provider/server/src/indexer/local_index_directory/mod.rs
```diff
@@ -154,6 +154,9 @@ pub struct PieceInfo {
     /// Source: <https://github.com/filecoin-project/boost/blob/16a4de2af416575f60f88c723d84794f785d2825/extern/boostd-data/model/model.go#L42-L46>
     pub complete_index: bool,
 
+    /// Roots of the car encapsulated in the piece.
```
jmg-duarte:
```suggestion
    /// Roots of the CAR archive encapsulated in the piece.
```

jmg-duarte (changes requested):
I think there's a fundamental issue in the `fetch` addition — it's not an addition, it's a replacement

Meaning that if I know everything I need to know about the SP, I can no longer retrieve the piece directly, which may hurt us in the long run.

I believe these flows should *separated* from each other, for testing, user experience, etc

storage-fetch/src/download.rs
cernicc: It's required because the [beetswap::Behaviour](https://docs.rs/beetswap/latest/beetswap/struct.Behaviour.html) struct requires the bound already.

storage-fetch/src/download.rs
cernicc: Hmm. What do you mean when you say "session"?

storage-fetch/src/download.rs
cernicc: I've changed the comment

examples/storage_provider_private.pem
cernicc: That is why I wrote `I think we should add more keys to the repo`.

I don't see a connection with the command you pasted. Even if the keys are pushed to the example folder, the command can still be used to extract a peer id(?)

Following the first comment. Why is it better to generate them on the fly?

storage-fetch/src/download.rs
cernicc: 😢

storage-fetch/src/peer_resolver.rs
cernicc: Nice one. This was the thing I was asking myself when writing this

storage-provider/server/src/p2p/mod.rs
cernicc: It looks very weird. My bad.

storage-fetch/src/download.rs
cernicc: I had a feeling that someone will stumble at this 😄 I've extended the comment

examples/storage_provider_private.pem
th7nder: I think storing any form of private keys on the repo is just not a good practice. 'Industry convention' if you will.

storage-fetch/src/download.rs
```diff
@@ -10,25 +11,36 @@ use cid::Cid;
...
+use primitives::p2p::{PieceInfoRequest, PieceInfoResponse, SP_REQUEST_RESPONSE_PROTOCOL};
 use thiserror::Error;
 use tokio::{
     fs::{File, OpenOptions},
     io::AsyncSeekExt,
 };
 use tracing::{debug, error, info, instrument, trace};
 
-use crate::p2p::{new_swarm, Behaviour, BehaviourEvent, InitSwarmError};
+const MAX_MULTIHASH_LENGTH: usize = 64;
```
th7nder: I think we should extract this value to primitives, more - it's already there: called `CID_SIZE_IN_BYTES`.

storage-fetch/src/download.rs
```diff
@@ -112,20 +134,27 @@ impl Client {
...
             blockstore,
-            root,
+            root: None,
             dag: HashMap::new(),
         })
     }
 
     /// Start download of some content with a payload cid.
-    pub async fn download(mut self) -> Result<(), ClientError> {
+    pub async fn download(mut self, &piece_cid: &Cid) -> Result<(), DownloadError> {
```
th7nder:
```suggestion
    pub async fn download(mut self, piece_cid: &Cid) -> Result<(), DownloadError> {
```

storage-fetch/src/download.rs
```diff
@@ -162,6 +195,10 @@ impl Client {
             .await?;
 
         if self.settings.extract {
+            let Some(root) = &self.root else {
+                return Err(DownloadError::UnsupportedNumRoots);
```
th7nder: I think this error is misleading. I think previously we guaranteed that we have roots, right?

storage-fetch/src/download.rs
```diff
@@ -234,7 +274,41 @@ impl Client {
...
+                            Ok(piece_info.roots[0])
+                        } else {
+                            Err(DownloadError::UnsupportedNumRoots)
+                        }?;
+
+                        // Start requesting blocks
+                        self.root = Some(root);
+                        self.request_block(root);
+                    }
+                    PieceInfoResponse::NotFound(_) => return Err(DownloadError::PieceUnknown),
```
th7nder: Short-circuiting here: `self.on_request_response(event)?;` and returning the error here is a bug.

If we're sending an request to multiple storage providers and one of them responds `PieceInfoResponse::NotFound` this will fail the entire download process. Or I'm misunderstanding something?

storage-fetch/src/main.rs
```diff
@@ -37,23 +34,63 @@ struct Cli {
     #[arg(long, value_parser = parse_duration)]
     timeout: Option<Duration>,
 
-    /// Payload CID
+    /// Deal id
+    #[arg(long)]
+    deal_id: u64,
```
th7nder: nit: this should not be an arg, but just a 'value'.
```suggestion
    deal_id: u64,
```

storage-fetch/src/main.rs
```diff
@@ -37,23 +34,63 @@ struct Cli {
...
+
+    /// Bootstrap node address
```
th7nder: We should add info on how to get this.

storage-fetch/src/main.rs
```diff
@@ -37,23 +34,63 @@ struct Cli {
...
+    #[arg(long)]
+    bootstrap_address: Multiaddr,
+
+    /// Bootstrap node peerid
```
th7nder: We should add info on how to get this.

storage-fetch/src/main.rs
```diff
@@ -37,23 +34,63 @@ struct Cli {
...
     #[arg(long)]
-    payload_cid: Cid,
+    bootstrap_peer: PeerId,
+
+    /// Parachain address
```
th7nder: We should add info on how to get this.

storage-fetch/src/peer_resolver.rs
```diff
@@ -0,0 +1,88 @@
...
+        PeerInfoResponse::Found(peer_info) => Ok(peer_info.multiaddrs),
+    }
+}
+
+/// Pull the swarm until we receive peer info response or an error is observed.
+async fn wait_peer_info_response(
+    mut swarm: Swarm<request_response::cbor::Behaviour<PeerIdRequest, PeerInfoResponse>>,
+) -> Result<PeerInfoResponse, anyhow::Error> {
+    loop {
+        let Some(event) = swarm.next().await else {
```
th7nder: In which cases swarm is exhausted? Does it reconnect automatically if it loses connection?

th7nder (changes requested):

storage-fetch/src/peer_resolver.rs
cernicc: I've refactored the `peer_resolver`. The name should better represent the function now.

examples/storage_provider_private.pem
cernicc: The only problem I see with having them in examples is that a user could use them when running their provider. Anyway. I've removed them

examples/start_sp.sh
cernicc: This was unremoved

storage-fetch/src/download.rs
cernicc: This was changed back to the original code

storage-fetch/src/download.rs
cernicc: This was also removed. But good catch. This would have been a bug.

storage-fetch/src/main.rs
cernicc: The bootstrap nodes should be pre-set in the future. What kind of info do you have in mind?

storage-fetch/src/main.rs
cernicc: https://github.com/eigerco/polka-storage/pull/796#discussion_r1982907344

storage-fetch/src/main.rs
cernicc: https://github.com/eigerco/polka-storage/pull/796#discussion_r1982907344

storage-fetch/src/peer_resolver.rs
cernicc: I've changed this a little bit. But to answer your question. The swarm's stream is infinite and should never be exhausted. So this `else` branch would never be taken.

storage-fetch/src/main.rs
cernicc: I like it as a named arg. That way you know what the number represents only from reading the command.

storage-fetch/src/main.rs
```diff
@@ -36,31 +39,130 @@ struct Cli {
...
+    ByPayloadCid {
+        /// Provider multiaddres used for the data download
+        #[arg(long)]
+        provider: Vec<Multiaddr>,
+
+        /// Cid of the data being downloaded
+        #[arg(long)]
+        payload_cid: Cid,
+    },
+    #[command()]
```
jmg-duarte, line 56: This looks a bit sus, are you sure you need the `()`?

storage-fetch/src/p2p/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+
+pub mod resolvers;
+
+/// Creates a temporary P2P node. The node then connects to the specified peer
+/// and submits a request. The future resolves when the response is received or
+/// the error is observed.
+#[instrument]
+pub(crate) async fn request_from_peer_sync<Req, Resp>(
+    protocol: &'static str,
+    (peer_id, peer_multiaddr): (PeerId, Multiaddr),
```
jmg-duarte: You're not using the tuple itself for anything
```suggestion
    peer_id: PeerId, peer_multiaddr: Multiaddr,
```

storage-fetch/src/p2p/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+use serde::{de::DeserializeOwned, Serialize};
+use tracing::{info, instrument};
+
+pub mod resolvers;
+
+/// Creates a temporary P2P node. The node then connects to the specified peer
+/// and submits a request. The future resolves when the response is received or
+/// the error is observed.
+#[instrument]
+pub(crate) async fn request_from_peer_sync<Req, Resp>(
```
jmg-duarte: This `async` function with a `sync` name feels weird, what do you mean with `sync`?

storage-provider/server/src/indexer/mod.rs
```diff
@@ -104,7 +105,15 @@ where
...
 /// Prepares indexes of a raw piece.
-async fn piece_indexes<P>(location: P) -> Result<Vec<IndexRecord>, ServerError>
+///
+/// # Returns
+///
+/// Returns a Result containing:
+/// - Ok((Vec<Cid>, Vec<IndexRecord>)): A tuple of:
+///   - Vector of Content IDs (roots from v1 header)
+///   - Vector of index records containing CIDs with their offsets and sizes
+/// - Err(ServerError): If any I/O or parsing operations fail
```
jmg-duarte: The error is somewhat self-explanatory, so there's no need to explain it.
```suggestion
/// Returns the root CIDs and the index records mapping CIDs to their respective offsets and sizes. 
```

storage-provider/server/src/p2p/mod.rs
```diff
@@ -193,6 +217,56 @@ where
...
+                    if self
+                        .swarm
+                        .behaviour_mut()
+                        .request_response
+                        .send_response(channel, response)
+                        .is_err()
+                    {
+                        error!("Failed to send piece info to {peer:?}");
+                    }
+                }
```
jmg-duarte: This would provide more info
```suggestion
					let response_result = self
                        .swarm
                        .behaviour_mut()
                        .request_response
                        .send_response(channel, response);
                    if let Err(err) = response_result {
                        error!("Failed to send piece info to {peer:?} with error: {err}");
                    }
                }
```

storage-provider/server/src/p2p/mod.rs
```diff
@@ -193,6 +217,56 @@ where
...
+                request_id,
+                error,
+            } => warn!("Failed to send response with id {request_id} to {peer}: {error}"),
+            RequestResponseEvent::InboundFailure {
+                peer,
+                request_id,
+                error,
+            } => warn!("Failed to receive message with id {request_id} from {peer}: {error}"),
+            RequestResponseEvent::ResponseSent { peer, request_id } => {
+                trace!("Response with id {request_id} sent to {peer}")
```
jmg-duarte: We can discuss this later but IMO `trace` is more "yolo-logging" where you put it everywhere, this seems more like a `debug`
```suggestion
                debug!("Response with id {request_id} sent to {peer}")
```

jmg-duarte (dismissed):
Looks good! Final polish.

I do have a question, as an external client (Delia) what API am I offered to resolve a PeerID -> Multiaddress?

storage-fetch/src/main.rs
cernicc: It's required 🤷

storage-fetch/src/p2p/mod.rs
cernicc: I meant like: the future will resolve when the response is received.

I'll remove the `_sync`.

cernicc:
@jmg-duarte Currently, the only way that Delia can get a mapping is by requesting the bootstrap node. The same approach as the `polka-fetch` is doing

aidan46 (approved):

storage-fetch/src/main.rs
th7nder: Just info in the doc how to run this command and how do you get the values to specify as params.
I wouldn't know how to run this before asking you or doing dive deep in the code, I guess.

th7nder (approved):

PR #797, opened 2025-03-05
feat(sp-server): Add deal param validation

### LINKED ISSUE #771: feat: Add deal parameters check to sp-server

aidan46:
### Description

This PR adds deal parameter validation on the sp-server side. It gets the deal parameters from chain and compares the incoming deal to the set deal parameters. Also updated the scripts so that they include the deal parameters validation.

Idk if I should add a script that rejects a deal that falls outside of the set deal parameters.

### DIFF
--- a/examples/rpc_publish.sh
+++ b/examples/rpc_publish.sh
@@ -16,6 +16,8 @@ trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT
 # requires the testnet to be running!
 export DISABLE_XT_WAIT_WARNING=1
 
+mkdir -p /tmp/polka-storage
+
 CLIENT="//Alice"
 PROVIDER="//Charlie"
 
@@ -29,9 +31,10 @@ CONFIG="/tmp/config.toml"
 P2P_PUBLIC_KEY="/tmp/polka-storage/public.pem"
 P2P_PRIVATE_KEY="/tmp/polka-storage/private.pem"
 P2P_ADDRESS="/ip4/127.0.0.1/tcp/62649"
+# Deal parameters JSON location
+DEAL_PARAMS="/tmp/deal_params.json"
 
 # Generate ED25519 private key
-mkdir -p /tmp/storage-provider
 openssl genpkey -algorithm ED25519 -out "$P2P_PRIVATE_KEY"
 # -outpubkey is only available in OpenSSL 3.4.0 onwards
 # https://github.com/openssl/openssl/commit/6c03fa21ed4bbc9fd6d3013fdf9f4646d231f831
@@ -52,13 +55,16 @@ PIECE_SIZE="$(echo "$INPUT_COMMP" | jq ".size")"
 PEER_ID="$(target/release/polka-storage-provider-client generate-peer-id --pubkey "$P2P_PUBLIC_KEY")"
 
 # echo config file in the file in the /tmp folder
-echo "seal_proof = '2KiB'
-post_proof = '2KiB'
-porep_parameters = '2KiB.porep.params'
-post_parameters = '2KiB.post.params'
+echo "seal_proof = '8MiB'
+post_proof = '8MiB'
+porep_parameters = '8MiB.porep.params'
+post_parameters = '8MiB.post.params'
 rendezvous_point_address = '$P2P_ADDRESS'
 p2p_key = '@$P2P_PRIVATE_KEY'
 rendezvous_point = '$P2P_BOOTSTRAP_PEER_ID'" > "$CONFIG"
+
+# echo deal parameters in the file in the /tmp folder
+echo '{ "minimum_price_per_block": 200, "deal_duration": { "lower": 50, "upper": 1800 }}' > "$DEAL_PARAMS"
 
 
 # Setup balances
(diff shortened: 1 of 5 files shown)


storage-provider/server/src/rpc.rs
```diff
@@ -190,9 +230,8 @@ impl StorageProviderRpcServer for RpcServerState {
 
     async fn propose_deal(&self, deal: SxtDealProposal) -> Result<CidString, RpcError> {
         // TODO(@jmg-duarte,26/11/2024): proper unit or e2e testing of these validations
-        self.validate_deal_proposal(&deal)
-            .await
-            .map_err(|err| RpcError::invalid_params(err, None))?;
+        self.validate_deal_proposal(&deal).await?;
```
cernicc: Is there a case where we want to validate the deal proposal, but we don't want to check if it fails within the parameters set?

examples/rpc_publish.sh
jmg-duarte, line 1: I think we're approaching the point where moving these to Rust or some other language might be worth the effort.

storage-provider/server/src/rpc.rs
jmg-duarte: If your (end) suggestion is to merge both, I believe this is better separated

aidan46: I don't think so. I considered adding a single `validate` function where we just call these 2 functions. We have to call `validate_deal_proposal` before `validate_proposed_deal_within_parameters` to make sure that the deal duration calculation does not underflow.

storage-provider/server/src/rpc.rs
```diff
@@ -180,6 +180,46 @@ impl RpcServerState {
...
+        &self,
+        deal: &SxtDealProposal,
+    ) -> Result<(), RpcError> {
+        let Some(deal_parameters) = self
+            .xt_client
+            .retrieve_deal_parameters(self.xt_keypair.account_id())
+            .await?
+        else {
+            return Ok(());
+        };
```
jmg-duarte, line 203:
```suggestion
        // NOTE: A possible improvement would be to set up a subscription that updates the local parameters only when they changes
        let Some(deal_parameters) = self
            .xt_client
            .retrieve_deal_parameters(self.xt_keypair.account_id())
            .await?
        else {
            return Ok(());
        };
```

jmg-duarte (commented):
LGTM but I have some notes

examples/rpc_publish.sh
aidan46: Yep, I agree. I think there is a benefit in also having it in a script as it gives an example of how to run the CLI commands in a 'real world scenario'.

examples/rpc_publish.sh
```diff
@@ -29,9 +31,10 @@ CONFIG="/tmp/config.toml"
 P2P_PUBLIC_KEY="/tmp/polka-storage/public.pem"
 P2P_PRIVATE_KEY="/tmp/polka-storage/private.pem"
 P2P_ADDRESS="/ip4/127.0.0.1/tcp/62649"
+# Deal parameters JSON location
+DEAL_PARAMS="/tmp/deal_params.json"
```
th7nder: We should scope everything to `/tmp/polka-storage`.
btw, instead of `/tmp` we should use `$TMPDIR` (a recent issue I struck upon on our GPU runner).

examples/start_sp.sh
```diff
@@ -16,6 +16,8 @@ P2P_PRIVATE_KEY="/tmp/polka-storage-provider/private.pem"
 P2P_BOOTSTRAP_PUBLIC_KEY="/tmp/zombienet/charlie-public.pem"
 # Config file location
 CONFIG="/tmp/polka-storage-provider/config.toml"
+# Deal parameters JSON location
+DEAL_PARAMS="/tmp/deal_params.json"
```
th7nder: Same thing here.

examples/rpc_publish.sh
```diff
@@ -52,14 +55,17 @@ PIECE_SIZE="$(echo "$INPUT_COMMP" | jq ".size")"
...
+echo "seal_proof = '8MiB'
+post_proof = '8MiB'
+porep_parameters = '8MiB.porep.params'
+post_parameters = '8MiB.post.params'
 rendezvous_point_address = '$P2P_ADDRESS'
 p2p_key = '@$P2P_PRIVATE_KEY'
 rendezvous_point = '$P2P_BOOTSTRAP_PEER_ID'" > "$CONFIG"
 
+# echo deal parameters in the file in the /tmp folder
+echo '{ "minimum_price_per_block": 200, "deal_duration": { "lower": 50, "upper": 1800 }}' > "$DEAL_PARAMS"
```
th7nder, line 70: Sidenote: I started to wonder what's the reason for having bounds as a rational client will always choose the `minimum_price_per_block` to be minimum.

th7nder (changes requested):

examples/rpc_publish.sh
aidan46: It is true that a client will always choose the best price. Once we start adding more customisation for deal parameters like setting a price for different sector sizes and deal durations I feel like the price would make more sense.

cernicc (approved):

th7nder (approved):

PR #800, opened 2025-03-10
feat: Add deal param RPC endpoints

aidan46:
### Description

This PR adds RPC endpoints for getting all deal parameters and deal parameters tied to a specific storage provider.

### DIFF
--- a/storage-provider/common/src/rpc/mod.rs
+++ b/storage-provider/common/src/rpc/mod.rs
@@ -10,7 +10,8 @@ use primitives::{
 };
 use serde::{Deserialize, Serialize};
 use storagext::types::market::{
-    ClientDealProposal as SxtClientDealProposal, DealProposal as SxtDealProposal,
+    ClientDealProposal as SxtClientDealProposal, DealParameters as SxtDealParameters,
+    DealProposal as SxtDealProposal,
 };
 use subxt::ext::sp_core::crypto::Ss58Codec;
 
@@ -35,6 +36,26 @@ pub trait StorageProviderRpc {
     /// Retrieve a deal proposal (includes the deal status).
     #[method(name = "retrieve_deal")]
     async fn retrieve_deal(&self, deal_id: DealId) -> Result<SxtDealProposal, RpcError>;
+
+    /// Retrieve deal parameters for a given storage provider.
+    /// Returns `None` if the given storage provider has not set any deal parameters.
+    #[method(name = "retrieve_deal_parameters")]
+    async fn retrieve_deal_parameters(
+        &self,
+        sp_account_id: <storagext::PolkaStorageConfig as subxt::Config>::AccountId,
+    ) -> Result<Option<SxtDealParameters>, RpcError>;
+
+    /// Retrieves all deal parameters contained in the market pallet
+    #[method(name = "retrieve_all_deal_parameters")]
+    async fn retrieve_all_deal_parameters(
+        &self,
+    ) -> Result<
+        Vec<(
+            <storagext::PolkaStorageConfig as subxt::Config>::AccountId,
+            SxtDealParameters,
+        )>,
+        RpcError,
+    >;
 }
 
 /// Storage Provider server information, such as start time and on-chain address.
--- a/storage-provider/common/src/rpc/mod.rs
+++ b/storage-provider/common/src/rpc/mod.rs
@@ -39,15 +39,15 @@ pub trait StorageProviderRpc {
 
     /// Retrieve deal parameters for a given storage provider.
     /// Returns `None` if the given storage provider has not set any deal parameters.
-    #[method(name = "retrieve_deal_parameters")]
-    async fn retrieve_deal_parameters(
+    #[method(name = "retrieve_sp_deal_parameters_for")]
+    async fn retrieve_sp_deal_parameters_for(
         &self,
         sp_account_id: <storagext::PolkaStorageConfig as subxt::Config>::AccountId,
     ) -> Result<Option<SxtDealParameters>, RpcError>;
 
     /// Retrieves all deal parameters contained in the market pallet
-    #[method(name = "retrieve_all_deal_parameters")]
-    async fn retrieve_all_deal_parameters(
+    #[method(name = "retrieve_sp_deal_parameters")]
+    async fn retrieve_sp_deal_parameters(
         &self,
     ) -> Result<
         Vec<(
--- a/storage-provider/server/src/rpc.rs
+++ b/storage-provider/server/src/rpc.rs
@@ -199,7 +199,7 @@ impl RpcServerState {
         // and compare the passed in parameters with the ones on-chain and update the ones on chain if needed.
         let Some(deal_parameters) = self
             .xt_client
-            .retrieve_deal_parameters(self.xt_keypair.account_id())
+            .retrieve_sp_deal_parameters_for(self.xt_keypair.account_id())
             .await?
         else {
             return Ok(());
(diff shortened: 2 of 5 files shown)


storagext/cli/src/cmd/market.rs
```diff
@@ -154,6 +157,18 @@ impl MarketCommand {
...
+                let deal_parameters = client.retrieve_all_deal_parameters().await?;
+
+                match output_format {
+                    OutputFormat::Plain => {
+                        println!("Deal Parameters: {:?}", deal_parameters)
+                    }
+                    OutputFormat::Json => {
+                        println!("{}", serde_json::to_string(&deal_parameters)?)
+                    }
+                }
```
cernicc, line 175:
```suggestion
                println!("{}", output_format.format(&params)?);
```

storagext/lib/src/clients/market.rs
```diff
@@ -360,6 +373,37 @@ impl MarketClientExt for crate::runtime::client::Client {
...
+            .storage()
+            .at_latest()
+            .await?
+            .iter(deal_parameters_query)
+            .await?;
+
+        let mut params = vec![];
+
+        while let Some(Ok(kv)) = deal_params.next().await {
+            let bytes = &kv.key_bytes[(kv.key_bytes.len() - 32)..];
```
th7nder:
```suggestion
            let bytes = &kv.key_bytes[..];
```
i don't get why we need `kv.key_bytes.len()`, and there probably is a better way of doing this conversion, but I don't want to bikeshed.

th7nder (dismissed):

storage-provider/common/src/rpc/mod.rs
```diff
@@ -35,6 +36,26 @@ pub trait StorageProviderRpc {
...
+    /// Returns `None` if the given storage provider has not set any deal parameters.
+    #[method(name = "retrieve_deal_parameters")]
+    async fn retrieve_deal_parameters(
+        &self,
+        sp_account_id: <storagext::PolkaStorageConfig as subxt::Config>::AccountId,
+    ) -> Result<Option<SxtDealParameters>, RpcError>;
+
+    /// Retrieves all deal parameters contained in the market pallet
+    #[method(name = "retrieve_all_deal_parameters")]
+    async fn retrieve_all_deal_parameters(
```
jmg-duarte: Raising for awareness, not for fixing.

I'm not a fan of "all" here but I don't have a better name.

* Making the previous method "retrieve_deal_parameter" doesn't make any sense
* Making this one "retrieve_deal_parameterss" is also absolutely atrocious

Maybe making the previous method "retrieve_sp_deal_parameters" or "retrieve_deal_parameters_for" (which would make it really nice to read) would be better overall

storagext/lib/src/clients/market.rs
jmg-duarte: Because of this
https://github.com/paritytech/subxt/issues/1201

The API returns the concatenated value with the hash instead of separated

@aidan46 please add the link + an explanation so we don't lose track of this "weird" solution we found

storage-provider/server/src/rpc.rs
```diff
@@ -381,6 +384,36 @@ impl StorageProviderRpcServer for RpcServerState {
...
+
+    async fn retrieve_deal_parameters(
+        &self,
+        sp_account_id: <storagext::PolkaStorageConfig as subxt::Config>::AccountId,
+    ) -> Result<Option<SxtDealParameters>, RpcError> {
+        match self.xt_client.retrieve_deal_parameters(sp_account_id).await {
+            Ok(Some(params)) => Ok(Some(params.into())),
+            Ok(None) => Ok(None), // No need to return error on none, as the SP can choose to not set deal parameters
+            Err(err) => Err(RpcError::new(INTERNAL_ERROR_CODE, err.to_string(), None)),
+        }
```
jmg-duarte: Hmm, what about:

```suggestion
        match self.xt_client.retrieve_deal_parameters(sp_account_id).await {
            // No need to return error on none, as the SP can choose to not set deal parameters
            Ok(params) => Ok(params.map(SxtDealParameters::from)),
            Err(err) => Err(RpcError::new(INTERNAL_ERROR_CODE, err.to_string(), None)),
        }
```

storagext/cli/src/cmd/market.rs
```diff
@@ -359,7 +374,7 @@ impl MarketCommand {
     async fn publish_deal_parameters<Client>(
         client: Client,
         account_keypair: MultiPairSigner,
-        deal_parameters: SxtDealParameters,
+        deal_parameters: OffchainDealParameters,
```
jmg-duarte: The `Sxt` type family has a meaning behind it: they're `subxt` generated types

As such, I'm curious to know why you didn't use the `Sxt` family — I don't have enough context to piece this out alone, I'm just looking for a reason

storagext/lib/src/clients/market.rs
```diff
@@ -360,6 +373,37 @@ impl MarketClientExt for crate::runtime::client::Client {
...
+            let array_u8: [u8; 32] = bytes.try_into().expect("On-chain validation of accountIDs");
+            let account = <crate::PolkaStorageConfig as subxt::Config>::AccountId::from(array_u8);
+            params.push((account, kv.value))
```
jmg-duarte: [Look ma', no `expect`! (hidden panic though)](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice)

This is the approach I suggested on the call:
```suggestion
            let mut buffer = [0u8; 32];
            buffer.copy_from_slice(&kv.key_bytes[(kv.key_bytes.len() - 32)..]);
            let account = <crate::PolkaStorageConfig as subxt::Config>::AccountId::from(buffer);
            params.push((account, kv.value))
```

storagext/lib/src/types/market.rs
```diff
@@ -7,7 +7,11 @@ use crate::{
     runtime::{
         bounded_vec::IntoBoundedByteVec,
         runtime_types::pallet_market::{
-            deal_parameters::{OffchainDealDurationBound, OffchainDealParameters},
+            deal_parameters::{
+                DealParameters as PolkaStorageDealParameters,
+                OffchainDealDurationBound as PolkaStorageOffchainDealDurationBound,
+                OffchainDealParameters as PolkaStorageOffchainDealParameters,
+            },
```
jmg-duarte: Similarly to the `Sxt` comment, please review the usage of the types and ensure that it matches it's friends.

From the other imports, it seems that instead of `PolkaStorage` these should be prefixed with `Runtime`

storagext/cli/src/cmd/market.rs
aidan46: The bounds for `fn format` are `std::fmt::Display + serde::Serialize` but `Vec<(<crate::PolkaStorageConfig as subxt::Config>::AccountId, SpecializedRuntimeDealParameters,)>` (or any `Vec`) does not implement `Display`

storagext/cli/src/cmd/market.rs
aidan46: In storagext the `Sxt` prefix is not used, which makes sense as it uses its own types without the prefix. I updated the `polka-storage-provider-*` to use the `Sxt` and `Runtime` prefixes where applicable.

storagext/cli/src/cmd/market.rs
cernicc: We could implement Display for the type, so that we have the same output when retrieving a single parameter or multiple of them. I am resolving the comment because I don't want to block the pr

cernicc (approved):

jmg-duarte (commented):
LGTM

jmg-duarte (approved):

PR #801, opened 2025-03-10
fix: Add account to deal parameters

aidan46:
### Description

This PR adds the storage provider account to the `DealParameters`. This is needed because we cannot decode the key in the storage map and when querying the deal parameters we need to know the storage provider address. The storagext interface has not changed, it simply gets the account id from the key passed in when running `publish_deal_parameters`. It does not make sense for the storage provider to have to pass in the account when we can derive it from the key.


storagext/cli/src/cmd/market.rs
```diff
@@ -154,6 +157,20 @@ impl MarketCommand {
...
+            MarketCommand::RetrieveAllDealParameters => {
+                let deal_parameters = client.retrieve_all_deal_parameters().await?;
+
+                match output_format {
+                    OutputFormat::Plain => {
+                        println!("[");
+                        for param in deal_parameters.iter() {
+                            println!("{}, ", param)
+                        }
+                        println!("]");
```
th7nder, line 169:
```suggestion
                    OutputFormat::Plain => {
                        println!("{:#?}", deal_parameters);
```

I assume this didn't work here?

th7nder (approved):

PR #804, opened 2025-03-12
fix(node): multiaddr default & validation

jmg-duarte:
### Description

Adds a default multiaddress and validation for said input to the Node

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -62,7 +62,9 @@
       with pkgs;
       {
         devShells.default = mkShell {
-          inherit buildInputs;
+          buildInputs = buildInputs ++ [
+            git
+          ];
 
           OPENSSL_NO_VENDOR = 1;
           CRATE_CC_NO_DEFAULTS = lib.optionalString pkgs.stdenv.isDarwin "1";
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -146,15 +146,10 @@ pub struct RunCmd {
     #[arg(long, value_parser = keypair_value_parser, required = false)]
     pub p2p_key: Option<Keypair>,
 
-    /// TCP listen address in the P2P network of Storage Providers and Collators
-    /// that the bootstrap node binds to.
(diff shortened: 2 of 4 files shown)


node/src/cli.rs
```diff
@@ -113,6 +113,27 @@ impl RelayChainCli {
     }
 }
 
+fn default_p2p_tcp_multiaddr() -> Multiaddr {
+    Multiaddr::from_str("/ip4/127.0.0.1/tcp/62649").expect("value should be a valid Multiaddr")
+}
+
+fn validate_tcp_multiaddr(s: &str) -> Result<Multiaddr, String> {
+    const IP4_TCP: [&'static str; 2] = ["ip4", "tcp"];
+    const IP6_TCP: [&'static str; 2] = ["ip6", "tcp"];
```
aidan46: I was about to complain about the use of strings cause I never would have guessed that [libp2p actually does it this way](https://docs.rs/libp2p/latest/libp2p/multiaddr/struct.ProtoStackIter.html)

node/src/cli.rs
```diff
@@ -127,8 +148,8 @@ pub struct RunCmd {
 
     /// Listen address in the P2P network of Storage Providers and Collators
     /// that the bootstrap node binds to.
-    #[arg(long, required = false)]
-    pub p2p_listen_address: Option<Multiaddr>,
+    #[arg(long, default_value_t=default_p2p_tcp_multiaddr(), value_parser = validate_tcp_multiaddr)]
```
aidan46:
```suggestion
    #[arg(long, default_value_t = default_p2p_tcp_multiaddr(), value_parser = validate_tcp_multiaddr)]
```
Formatting to keep it consistent with the other `default_value_t`'s

cernicc (approved):

flake.nix
```diff
@@ -62,7 +62,9 @@
       with pkgs;
       {
         devShells.default = mkShell {
-          inherit buildInputs;
+          buildInputs = buildInputs ++ [
+            git
```
th7nder, line 66: It should be added in the buildInputs at the top.

th7nder (approved):

PR #809, opened 2025-03-18
Separate pallet-market benchmarks crate

Jinxit:
### Description

_Context: Benchmarking `pallet-storage-provider` requires turning on real proof verification in `polka-storage-node` (to get accurate benchmark numbers), which breaks benchmarks that currently rely on fake proofs._

Currently `pallet-market` depends on `pallet-storage-provider`, and `pallet-storage-provider` has a dev dependency on `pallet-market`. In order to move test data for benchmarking into `primitives` and reuse real proofs in other pallets there's a need to break this cycle. The solution is the same as for `pallet-storage-provider`, moving the benchmarks (and tests, since they use the same mocks) to a separate crate.

#### ~~Patch to fflonk~~

~~I hit this error at some point:~~
```
$ cargo test --workspace --exclude maat
    Updating git repository `https://github.com/w3f/fflonk`
error: no matching package named `fflonk` found
location searched: https://github.com/w3f/fflonk
required by package `ring v0.1.0 (https://github.com/w3f/ring-proof?rev=665f5f5#665f5f51)`
    ... which satisfies git dependency `ring` (locked to 0.1.0) of package `bandersnatch_vrfs v0.0.4 (https://github.com/w3f/ring-vrf?rev=0fef826#0fef8266)`
    ... which satisfies git dependency `bandersnatch_vrfs` (locked to 0.0.4) of package `sp-core v28.0.0 (https://github.com/paritytech/polkadot-sdk?tag=polkadot-stable2412#967989c5)`
    ... which satisfies git dependency `sp-core` (locked to 28.0.0) of package `polka-storage-node v0.0.0 (/Users/lucas.astrom/Code/polka-storage/node)`
```
~~Which I fixed with a [patch](https://github.com/paritytech/polkadot-sdk/issues/7653#issuecomment-2674485339), but I'm not sure if that's something we want to do. Opinions?~~

Leaving the above for posterity, but reverting the lockfile to its original state fixed the issue. I must've somehow caused it to update earlier.

### Checklist

  - [x] ~~fflonk patch, y/n?~~
  - [x] https://github.com/eigerco/polka-storage/issues/739
- ~~[ ] Have you tested this solution?~~
  - [x] At first I was thinking to just duplicate test data/code, but there's so much that could be reused that it becomes silly. Some stuff will still necessarily be duplicated (`run_to_block` comes to mind), but I aim to minimize it.


cernicc:
I haven't worked with benchmarks much. With that in mind, the pr looks good.

pallets/market/benchmarks/Cargo.toml
jmg-duarte, line 1: Did you review this to ensure all crates are used?

jmg-duarte (dismissed):

cernicc (dismissed):

pallets/market/benchmarks/Cargo.toml
Jinxit: Nope! Good catch.

cernicc (approved):

jmg-duarte (approved):

PR #814, opened 2025-03-31
fix: set mainnet values for 1GiB sectors

### LINKED ISSUE #755: Mainnet values for precommit, prove commit and submit windowed
1. Need to figure out how many partitions in a deadline at a time we can prove with 1GiB sector 
     - ~~some assumptions, like 50% of the deadline can be for calculating (because GPU)~~
        - Single Partition PoSt generation takes around ~10 secs.
     - ~~need to have times of how much time does it take to generate a Proof of SpaceTime of 1 GiB sector compared to Proof of Replication~~
         - ~20 minutes to generate a single PoSt 
     - ~~after we have those values, we can estimate how much storage we can support with the current system~~
     - ~~take into account those values will need to be recalculated after #751 (done)~~
2. Need to seed reasonable values in https://github.com/eigerco/polka-storage/blob/43105aaab60d612a32109c4e33711ffbdda977c4/runtime/src/configs/mod.rs#L311

th7nder:
### Description

It sets expected values for the 'testnet' launch in P3.
Some of those are just my guesses, feel free to challenge them.

Closes #755.

### DIFF
--- a/runtime/src/configs/mod.rs
+++ b/runtime/src/configs/mod.rs
@@ -327,8 +327,8 @@ parameter_types! {
 
     // Fault declaration/recovery functionality has been postponed, this won't make a difference until #592 is done.
     pub const FaultMaxAge: BlockNumber = 1 * DAYS;
-    /// WPoStChallengeLookBack + a tiny margin.
-    pub const FaultDeclarationCutoff: BlockNumber = WPoStChallengeLookBack + (1 * MINUTES);
+    pub const FaultDeclarationCutoffSlack: BlockNumber = 1 * MINUTES;
+    pub const FaultDeclarationCutoff: BlockNumber = WPoStChallengeLookBack + FaultDeclarationCutoffSlack;
     // <https://github.com/filecoin-project/builtin-actors/blob/8d957d2901c0f2044417c268f0511324f591cb92/runtime/src/runtime/policy.rs#L299>
     pub const AddressedSectorsMax: u64 = 25_000;
 
--- a/runtime/src/configs/mod.rs
+++ b/runtime/src/configs/mod.rs
@@ -322,7 +322,8 @@ parameter_types! {
     // The chain can verify 1 proof per block (of 25 sectors per partitions) currently.
     // It means deadline time (blocks) == max theoretical number of the partitions per deadline.
     // There is no global limit, so currently 1 Storage Provider could overload the entire chain.
-    pub const MaxPartitionsPerDeadline: u64 = WpostChallengeWindow;
+    // 1 deadline = 30 minutes, 30 * 10 blocks = 300.
+    pub const MaxPartitionsPerDeadline: u64 = 300;
 
     // Fault declaration/recovery functionality has been postponed, this won't make a difference until #592 is done.
     pub const FaultMaxAge: BlockNumber = 1 * DAYS;
--- a/runtime/src/configs/mod.rs
+++ b/runtime/src/configs/mod.rs
@@ -309,32 +309,34 @@ impl pallet_collator_selection::Config for Runtime {
 parameter_types! {
     // Storage Provider Pallet
     pub const WpostProvingPeriod: BlockNumber = DAYS;
-    // Half an hour (=48 per day)
-    // 30 * 60 = 30 minutes
-    // SLOT_DURATION is in milliseconds thats why we / 1000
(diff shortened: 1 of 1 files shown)


runtime/src/configs/mod.rs
jmg-duarte, line 1: Can we take this chance to also make the const names uniform? There's `Wpost`, `WPoSt`, etc

runtime/src/configs/mod.rs
```diff
@@ -309,32 +309,35 @@ impl pallet_collator_selection::Config for Runtime {
...
+
+    // Fault declaration/recovery functionality has been postponed, this won't make a difference until #592 is done.
+    pub const FaultMaxAge: BlockNumber = 1 * DAYS;
+    /// WPoStChallengeLookBack + a tiny margin.
+    pub const FaultDeclarationCutoff: BlockNumber = (10 * MINUTES) + (1 * MINUTES);
     // <https://github.com/filecoin-project/builtin-actors/blob/8d957d2901c0f2044417c268f0511324f591cb92/runtime/src/runtime/policy.rs#L299>
     pub const AddressedSectorsMax: u64 = 25_000;
 
+    pub const MinSectorExpiration: BlockNumber = 1 * DAYS;
+    pub const MaxSectorExpiration: BlockNumber = 365 * DAYS;
```
jmg-duarte: I think we can up this, otherwise someone that does a deal of 1 year will either complete a block of put the SP in a weird position where they can only accept deals that start and end within that first deal

runtime/src/configs/mod.rs
```diff
@@ -309,32 +309,35 @@ impl pallet_collator_selection::Config for Runtime {
...
+    // The chain can verify 1 proof per block (of 25 sectors per partitions) currently.
+    // It means deadline time (blocks) == max theoretical number of the partitions per deadline.
+    // There is no global limit, so currently 1 Storage Provider could overload the entire chain.
+    // 1 deadline = 30 minutes, 30 * 10 blocks = 300.
+    pub const MaxPartitionsPerDeadline: u64 = 300;
+
+    // Fault declaration/recovery functionality has been postponed, this won't make a difference until #592 is done.
+    pub const FaultMaxAge: BlockNumber = 1 * DAYS;
+    /// WPoStChallengeLookBack + a tiny margin.
+    pub const FaultDeclarationCutoff: BlockNumber = (10 * MINUTES) + (1 * MINUTES);
```
jmg-duarte: Maybe just use the value then?
```suggestion
    /// WPoStChallengeLookBack + a tiny margin.
    pub const FaultDeclarationCutoff: BlockNumber = WPoStChallengeLookback + (1 * MINUTES);
```

runtime/src/configs/mod.rs
jmg-duarte: We should also take the change to add proper docs (`///`) to either the consts or the traits

runtime/src/configs/mod.rs
```diff
@@ -309,32 +309,35 @@ impl pallet_collator_selection::Config for Runtime {
...
+    // PoSt for a single partition is ~20sec, it's a huge margin to start generating the proofs before deadline starts.
+    pub const WPoStChallengeLookBack: BlockNumber = 10 * MINUTES;
+
+    // Finality in polkadot is 12-60 seconds, gives us a finalized randomness to base the challenge on.
+    pub const PreCommitChallengeDelay: BlockNumber = 1 * MINUTES;
+    // With 1GiB sector size and max security params it takes ~22 minutes to create a replica and zk-SNARK on top of it.
+    pub const MaxProveCommitDuration: BlockNumber =  30 * MINUTES;
+    // The chain can verify 1 proof per block (of 25 sectors per partitions) currently.
+    // It means deadline time (blocks) == max theoretical number of the partitions per deadline.
+    // There is no global limit, so currently 1 Storage Provider could overload the entire chain.
```
jmg-duarte, line 324: Big sad

jmg-duarte (approved):

jmg-duarte (changes requested):
Previous one wasn't an approve, my bad

runtime/src/configs/mod.rs
th7nder: Current limitation :(

runtime/src/configs/mod.rs
th7nder: Docs are already there and quite reasonable.

cernicc (approved):

jmg-duarte (approved):

PR #818, opened 2025-04-04
test(primitives): Extract commonalities from market and storage-provider pallet benchmarks, and use real proof data

Jinxit:
### Description

A problem arising from using real proofs for benchmarking the storage-provider pallet is that the runtime must use the real proof verifier, which means the market pallet benchmarks break. While the setup created for storage-provider _could_ be copied across, it feels a bit more maintainable to:
1. Extract commonalities and reuse them.
2. Create a reproducible (and deterministic) way to generate the data.

This PR attempts to move test setup code, mostly for benchmarks, which is currently duplicated between the market and storage-provider pallets. I didn't try to extract _everything_, I was mostly aiming to get it to work to not make an already large PR any larger. There's a new CLI command to `storage-provider-client` called `benchmark-data` to generate code, keys and proofs which are then embedded in the benchmark binary. I've also created traits for things which are common to the respective pallets' configs, such as `CurrencyProvider`, `BalanceOf`, and traits for each pallet's config itself - since that is kind of part of the (internal) interfaces of the pallets.

**This is the big one**, so I'm very much open to comments for improvements. Going commit by commit might not _compile_, but the changes are grouped on a best-effort basis to make reviewing them easier. `be6fb5d` is a bit of a mess though, I'll admit.

There is some overlap with what this is doing and the downloading of parameters (`just download-params`), but I haven't really made any attempt to unify the two approaches.

As a side effect of creating `MarketProvider` and `StorageProviderProvider` (what a name, it's like I'm back in Java-land), the `#[pallet::constant]` tags have disappeared from the constants which are used by the other pallet, respectively. This means the constants are no longer a part of the pallet metadata, which I _think_ is fine for now. The end goal should probably be to merge the two pallets anyway and move stuff back out of `primitives`.

### Checklist

  - [x] Less pallet metadata
  - [x] #739 
  - [x] Copy-paste the setup
  - [x] Merge storage-provider and market (maybe later?) 


storage-provider/client/src/commands/proofs.rs
```diff
@@ -527,6 +597,228 @@ fn post(
...
+        input_path.display()
+    );
+    let (comm_p, padded_piece_size) = calculate_piece_commitment(&input_path).await?;
+
+    let params_root = PathBuf::from(BENCH_CACHE_DIR).join("params");
+
+    let sector_size = seal_proof.sector_size();
+    let porep_params_path = params_root.join(format!("{sector_size}.{POREP_PARAMS_EXT}"));
+    let porep_params_vk_path = params_root.join(format!("{sector_size}.{POREP_VK_EXT_SCALE}"));
+    if !tokio::fs::try_exists(&porep_params_vk_path).await? {
```
th7nder: This should fail instead of generate, we should always use those shared params that are (**should**) be stored on blob.

primitives/src/test_data.rs
```diff
@@ -0,0 +1,1662 @@
...
+            .unwrap()
+    }
+
+    pub fn load(sector_size: SectorSize) -> BenchmarkData<T> {
+        match sector_size {
+            // DO NOT MODIFY
+            // This code has been generated by `target/release/polka-storage-provider-client proofs benchmark-data --sector-size 2KiB examples/test-data-big.car`
+            SectorSize::_2KiB => BenchmarkData {
+                storage_provider_name: "//StorageProvider",
+                verifying_key: include_bytes!("../../target/bench/params/2KiB.porep.vk.scale"),
```
th7nder: We need to store verifying key as well as generated proofs in the repo, somewhere committed. If we don't, CI will fail.
Params cannot be stored on the repo, as they're too big.

I'd suggest the workflow looks like this:
- we generate the bench data for 1GiB sectors (both proofs, vks and stuff) on the beefy machine, by running the command `benchmark-data`
- upload this as a backup to our blob storage (i'm doing it now)
- commit the generate benched data to the repo (as well as the benchmarks)

storage-provider/client/src/commands/proofs.rs
```diff
@@ -527,6 +597,228 @@ fn post(
...
+        comm_p,
+        sectors: Vec::with_capacity(MAX_SECTORS_PER_CALL as usize),
+    };
+
+    let proofs_root = PathBuf::from(BENCH_CACHE_DIR)
+        .join("proofs")
+        .join(sector_size.to_string());
+    tokio::fs::create_dir_all(&proofs_root).await?;
+
+    for sector_number in 0..MAX_SECTORS_PER_CALL {
```
th7nder, line 688: This `MAX_SECTORS_PER_CALL` is probably to be adjusted. I think that with 1GiB sectors we'll be able to `ProveCommit` only 1 per call, PreCommit is much different, it's not that compute heavy, but still.

To generate 20 seperate proofs for 1GiB, it'll take like 5 hours (1 proof -> ~20minutes). 
Do we wanna do that if we'll accept one proof for prove commit anyways?

pallets/market/src/lib.rs
```diff
@@ -58,49 +57,21 @@ pub mod pallet {
...
     pub const LOG_TARGET: &'static str = "runtime::market";
 
-    /// Allows to extract Balance of an account via the Config::Currency associated type.
-    /// BalanceOf is a sophisticated way of getting an u128.
-    pub type BalanceOf<T> =
-        <<T as Config>::Currency as Currency<<T as SystemConfig>::AccountId>>::Balance;
-
     #[pallet::config]
-    pub trait Config: frame_system::Config {
+    pub trait Config: frame_system::Config + CurrencyProvider + MarketProvider {
```
th7nder, line 61: This seems fishy, but hey, as long as it works.

pallets/market/benchmarks/src/mock.rs
```diff
@@ -78,44 +81,53 @@ parameter_types! {
...
     // Storage Provider Pallet
-    pub const WPostProvingPeriod: BlockNumber = 6 * MINUTES;
-    pub const WPoStPeriodDeadlines: u64 = 3;
-    pub const WPostChallengeWindow: BlockNumber = 2 * MINUTES;
-    pub const WPostChallengeLookBack: BlockNumber = MINUTES;
+    pub const WPoStPeriodDeadlines: u64 = 10;
+    pub const WPoStProvingPeriod: BlockNumber = 40 * MINUTES;
+    pub const WPoStChallengeWindow: BlockNumber = 4 * MINUTES;
+    pub const WPoStChallengeLookBack: BlockNumber = MINUTES;
     pub const MinSectorExpiration: BlockNumber = 5 * MINUTES;
```
th7nder, line 118: Why did we change those values anyways?
'Mainnet' ones (the place where we actually need benchmarking) are a bit [different](https://github.com/eigerco/polka-storage/blob/817826c86d95aeea7ee6121c831d910b015ee9f9/runtime/src/configs/mod.rs#L316),.

Btw. the current `testnet` is more like a `devnet`.

primitives/src/test_data.rs
```diff
@@ -0,0 +1,1662 @@
+extern crate alloc;
```
th7nder, line 1: shouldn't test_data be behind some feature flag?
It'll pull a lot of stuff into the WASM runtime when when we're doing 'include_bytes'

primitives/src/test_data.rs
```diff
@@ -0,0 +1,1662 @@
...
+                }
+            })
+            .collect::<Vec<_>>()
+            .try_into()
+            .unwrap()
+    }
+
+    pub fn load(sector_size: SectorSize) -> BenchmarkData<T> {
+        match sector_size {
+            // DO NOT MODIFY
```
th7nder: Btw. shouldn't we require only one sector_size here to be loaded?
Like if we need benchmarks only for 1GiB/8MiB to not pollute the binary?

th7nder (changes requested):
Good stuff! In this first pass I just focused on the high-level functionality and concepts, not at the trait choices/code design.

primitives/src/test_data.rs
Jinxit: It's behind a `#[cfg(feature = "testing")]` in the `lib.rs` file.

primitives/src/test_data.rs
Jinxit: You mean with features or something? I mostly started like this when going back and forth between generating data and running it, it was convenient to easily swap them out.

pallets/market/benchmarks/src/mock.rs
Jinxit: I don't remember the exact error but I think I ran into timing issues with the previous values, so I matched what we have in the storage-provider pallet.

pallets/market/src/lib.rs
Jinxit: It's kinda the same thing that we're already doing by inheriting `frame_system::Config` to get access to `AccountId` and `BlockNumber` etc.

storage-provider/client/src/commands/proofs.rs
Jinxit: If we're planning on lowering the max sectors (even to 1), then yeah I agree this is a good time to do it.

primitives/src/test_data.rs
Jinxit: Sounds good, I'll adjust the paths accordingly and add it to the download script.

storage-provider/client/src/commands/proofs.rs
Jinxit: This was written before the downloading of params so it's generating a completely separate set - I'll switch it over.

Jinxit: Actually, thinking about this again, shouldn't it generate if they are missing? If we're using this as a command to run and then commit to the repo (or to the blob store), then this should be how that data is generated in the first place. This command shouldn't be run on developer machines, really, outside of developing the command itself.

storage-provider/client/src/commands/proofs.rs
```diff
@@ -237,15 +317,16 @@ async fn calculate_piece_commitment(input_path: PathBuf) -> Result<(), CliError>
...
+
+    Ok((commitment, padded_piece_size))
 }
 
 fn generate_porep_params(
-    output_path: Option<PathBuf>,
+    output_path: Option<impl AsRef<Path>>,
     seal_proof: RegisteredSealProof,
 ) -> Result<(), CliError> {
     let output_path = if let Some(output_path) = output_path {
```
cernicc: I think it is better to accept an owned type if the ownership is needed by the function  logic. The caller can then decide if another allocation is needed or if the ownership can just be transferred to the function.

In this example the owned value is immediately created from the reference. This is a good indication of the point above.

storage-provider/client/src/commands/proofs.rs
```diff
@@ -130,6 +144,59 @@ pub enum ProofsCommand {
         /// CID - CommR of a replica (output of `porep` command)
         comm_r: String,
     },
+    /// Generates PoRep params, PoSt params, proofs and commitments used for benchmarking.
```
cernicc, line 147: I have a feeling like we should hide everything required for benchmarks behind some feature flag. What do you think?

primitives/src/configs.rs
```diff
@@ -0,0 +1,46 @@
...
+use frame_system::pallet_prelude::BlockNumberFor;
+use sp_core::Get;
+use sp_runtime::traits::{IdentifyAccount, Verify};
+
+/// Allows to extract Balance of an account via the Config::Currency associated type.
+/// BalanceOf is a sophisticated way of getting an u128.
+pub type BalanceOf<T> = <<T as CurrencyProvider>::Currency as Currency<
+    <T as frame_system::Config>::AccountId,
+>>::Balance;
+pub trait CurrencyProvider: frame_system::Config {
```
cernicc, line 15: Missing new line before the trait

cernicc (commented):
This benchmarking looks like a rocket science :D Left some comments. Good job figuring this out.

storage-provider/client/src/commands/proofs.rs
th7nder: I'd agree about the generation, but to generate them properly you need to have a huge-ass machine and then update them in lots of places. Like the repo, chain, genesis, etc. 
I think this is not a trivial step and needs to be done **explicitly**.

pallets/market/src/lib.rs
th7nder: Ok

pallets/market/benchmarks/src/mock.rs
th7nder: I don't mind, i don't think those values are affecting benchmarks in any meaningful way.

primitives/src/test_data.rs
th7nder: Cool

storage-provider/client/src/commands/proofs.rs
Jinxit: I agree. At first I considered `runtime-benchmarks` but I'm not sure whether that's already too overloaded with the _actual_ benchmarking code.

storage-provider/client/src/commands/proofs.rs
Jinxit: It was less about not cloning and more about developer experience in an earlier commit where I was using a `TempDir` directly which implements `AsRef<Path>`. That changed along the way, and now it's just a `PathBuf` on both call sites.

I'm up for changing it, but I don't think performance is the right reason. The command takes minutes to run after all. But writing simpler code I can agree to.

primitives/src/test_data.rs
th7nder: I guess it was bothering me that we'll only use SectorSize 1GiB for benchmarks and have multiple sector sizes in this switch which are not necessary.

Jinxit: Yeah it's a good point, I think I'll switch it out (ha) to just returning the 1GB data as soon as I get that working.

Jinxit:
I didn't address all comments yet, but I've now updated to using the 1GB proof data and only doing 1 sector at a time. For now I've left it as a `BoundedVec` with a length of 1 to not change APIs everywhere, there's already enough in this PR.

Please give some extra attention to b587a4ed0c9545723ac986f8d7dfbfc7e82b2252 - I'm not 100% sure I replaced the right constants, but I think some of them were wrong in the first place (and happened to fit).

storage-provider/client/src/commands/proofs.rs
Jinxit: I think I prefer the `AsRef<Path>` version, but here's the less-clone version: 7acdd0e33abe11f4421adf4ae2a5ba4419314591

storage-provider/client/src/commands/proofs.rs
Jinxit: @jmg-duarte @th7nder
What do you think I should call it?

th7nder: I don't mind if this is not behind a feature flag, but `benchmarking` seems reasonable (?)

Cargo.toml
```diff
@@ -97,6 +97,7 @@ pairing = "0.23"
 polkavm = "0.9.3"
 polkavm-derive = "0.9.1"
 polkavm-linker = "0.9.2"
+proc-macro2 = { version = "1.0.93" }
```
jmg-duarte, line 100: This looks kinda sus, version unification?

pallets/market/benchmarks/src/benchmarking.rs
```diff
@@ -231,159 +162,124 @@ mod benchmarks {
...
+        let client = generate_benchmark_account::<T>(CLIENT);
+        setup_account_balance::<T>(client.0.clone());
+
+        let proposals = data.deal_proposals(&client, n);
+        let cost: u32 = proposals
+            .iter()
+            .map(|p| p.proposal.total_storage_fee().unwrap())
+            .sum::<u128>()
+            .try_into()
+            .unwrap();
```
jmg-duarte: I get why you're doing the u128 -> u32 -> Balance, but maybe you can just do u128 -> Balance through the UniqueSaturatedFrom

Because:
```
Balance: AtLeast32BitUnsigned
AtLeast32BitUnsigned: AtLeast32Bit
AtLeast32Bit: BaseArithmetic
BaseArithmetic: UniqueSaturatedFrom<u128>
```

https://paritytech.github.io/polkadot-sdk/master/sp_arithmetic/traits/trait.BaseArithmetic.html

pallets/market/benchmarks/src/benchmarking.rs
```diff
@@ -231,159 +162,124 @@ mod benchmarks {
...
+
+        let expiration_block = pre_commit_infos
+            .iter()
+            .map(|info| info.expiration)
+            .max()
+            .unwrap();
+        run_to_block::<T>(
+            expiration_block
+                .try_into()
+                .unwrap_or_else(|_| panic!("failed to convert expiration block to block number")),
```
jmg-duarte, line 282: What is the difference of this and `expect`?

pallets/market/benchmarks/src/mock.rs
```diff
@@ -78,44 +81,53 @@ parameter_types! {
 
 #[cfg(feature = "runtime-benchmarks")]
 parameter_types! {
-    pub const MinDealDuration: u64 = 5 * MINUTES;
-    pub const MaxDealDuration: u64 = 180 * MINUTES;
-    pub const PreCommitChallengeDelay: BlockNumber = 1 * MINUTES;
+    // NOTE: Changed from Testnet's 0 to 10 to match the storage-provider client's `porep` command.
+    pub const PreCommitChallengeDelay: BlockNumber = 10;
+    pub const MinDealDuration: u64 = 2 * MINUTES;
+    pub const MaxDealDuration: u64 = 30 * MINUTES;
```
jmg-duarte, line 87: We should be testing with the values that are going to make it to production. I'm not sure what those are but this comment doesn't inspire much confidence either.

We might just need to make it clear what "testnet" means for our project, etc

cc @th7nder

pallets/market/benchmarks/src/mock.rs
```diff
@@ -78,44 +81,53 @@ parameter_types! {
...
+    pub const WPoStPeriodDeadlines: u64 = 10;
+    pub const WPoStProvingPeriod: BlockNumber = 40 * MINUTES;
+    pub const WPoStChallengeWindow: BlockNumber = 4 * MINUTES;
+    pub const WPoStChallengeLookBack: BlockNumber = MINUTES;
     pub const MinSectorExpiration: BlockNumber = 5 * MINUTES;
-    pub const MaxSectorExpiration: BlockNumber = 60 * MINUTES;
+    pub const MaxSectorExpiration: BlockNumber = 360 * MINUTES;
     pub const SectorMaximumLifetime: BlockNumber = 120 * MINUTES;
-    pub const MaxProveCommitDuration: BlockNumber = 5 * MINUTES;
+    // NOTE: Changed from Testnet's 5 to 15 since the pre-commit delay is 10 blocks.
```
jmg-duarte, line 121: Same as above

primitives/src/test_data.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub storage_provider_name: &'static str,
+    pub verifying_key: &'static [u8],
+    pub seal_proof: RegisteredSealProof,
+    pub post_type: RegisteredPoStProof,
+    pub seal_randomness_height: u64,
+    pub pre_commit_block_number: u64,
+    pub comm_p: Commitment<CommP>,
+    pub sectors: Vec<SectorData>,
+    _phantom: PhantomData<T>,
+}
```
jmg-duarte: Move this to be close to the impl

pallets/storage-provider/benchmarks/src/benchmarking.rs
```diff
@@ -0,0 +1,299 @@
...
+    (sp.account_id, sectors)
+}
+
+fn check_pre_commit_sectors<T>(n: u32, sp_id: AccountId32)
+where
+    T: crate::Config<AccountId = AccountId32>,
+    BlockNumberFor<T>: From<u64> + Into<u64>,
+{
+    let state = SpPallet::<T>::storage_providers(sp_id).unwrap();
+    let balance: BalanceOf<T> = (n * 1_u32).into();
```
jmg-duarte: What is the `* 1_u32` doing?

jmg-duarte (commented):
This PR is too big for me to provide a meaningful review. I'd prefer having it split into multiple parts but I don't want to discard the reviews that were already done.

Next time, I'm rejecting.

pallets/market/benchmarks/src/benchmarking.rs
Jinxit: `expect` requires a `Debug` implementation.

pallets/storage-provider/benchmarks/src/benchmarking.rs
Jinxit: I think it used to be something else, so it's just a leftover that I forgot to remove.

Cargo.toml
Jinxit: I added it as a dependency for storage-provider-client, where I use a `proc_macro2::TokenStream` to generate the output code.

pallets/market/benchmarks/src/mock.rs
th7nder: I'd say we can leave it be in this PR, as matching those values to the mainnet ones will require changes across many things. Just create a follow-up I'd say.

pallets/storage-provider/benchmarks/src/benchmarking.rs
```diff
@@ -0,0 +1,299 @@
...
+        sectors_pre_commits
+    ));
+
+    assert_ok!(ProofsPallet::<T>::set_porep_verifying_key(
+        RawOrigin::Signed(sp.account_id.clone()).into(),
+        data.seal_proof,
+        data.verifying_key.to_vec(),
+    ));
+
+    run_to_block::<T>(15.into());
```
th7nder: those 'magic values' will bite us when we change to the 'mainnet params'

primitives/src/test_data.rs
```diff
@@ -0,0 +1,246 @@
...
+            .try_into()
+            .unwrap()
+    }
+
+    pub fn load() -> BenchmarkData<T> {
+        // DO NOT MODIFY
+        // This code has been generated by `./target/release/polka-storage-provider-client proofs benchmark-data --sector-size 1GiB examples/big_file_184k.car`
+        BenchmarkData {
+            storage_provider_name: "//StorageProvider",
+            verifying_key: include_bytes!("../../examples/benchmark/keys/1GiB.porep.vk.scale"),
```
th7nder: nit: I'd say overall the `examples/` directory stopped being useful for purposes of both bechmarking and testing.
1. The keys we're using in benchmarks are also used in genesis config, so it's a bit odd that genesis config uses something from `examples/benchmark/keys`.

`test-fixtures` or something similar would be more fitting, but this is just naming. It's not like super important, it just creates confusion.

storage-provider/client/src/commands/proofs.rs
```diff
@@ -527,6 +596,224 @@ fn post(
...
+    );
+    let (comm_p, padded_piece_size) = calculate_piece_commitment(&input_path).await?;
+
+    let params_root = PathBuf::from(PARAMS_CACHE_DIR);
+    let keys_root = &PathBuf::from(KEYS_DIR);
+
+    let sector_size = seal_proof.sector_size();
+    let porep_params_path = params_root.join(format!("{sector_size}.{POREP_PARAMS_EXT}"));
+    let porep_params_vk_path = keys_root.join(format!("{sector_size}.{POREP_VK_EXT_SCALE}"));
+    if !tokio::fs::try_exists(&porep_params_vk_path).await? {
```
th7nder:
```suggestion
    if !tokio::fs::try_exists(&porep_params_path) || !tokio::fs::try_exists(&porep_params_vk_path).await? {
```

th7nder (dismissed):
That's a piece of work which seems complete, great work! I wouldn't say any of those comments is blocking.

primitives/src/test_data.rs
Jinxit: Yeah, I agree. I kinda got lost a bit when I realized I don't currently have replacements for the _other_ proofs in `examples/` (specifically PoSt). Before then it was named `test-data`, and I wanted to get this PR done so I went with `benchmark` to distinguish the two.

pallets/storage-provider/benchmarks/src/benchmarking.rs
Jinxit: I didn't replace it in all tests etc, but at least in the benchmarks those numbers are now based on the pallet constants.

pallets/market/benchmarks/src/mock.rs
Jinxit: #824

pallets/storage-provider/benchmarks/src/benchmarking.rs
th7nder: good, not saying you should do this now - just a note - resolving

pallets/market/benchmarks/src/mock.rs
th7nder: #824

th7nder (approved):

jmg-duarte (approved):

PR #819, opened 2025-04-07
fix: change polkastore address and change keys

th7nder:
### Description

I needed to regen params and store them in the different blob store (because the old one was deleted).
Updated proofs and tests will be done by @Jinxit in #818.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -204,8 +204,8 @@ generate-proof-params sector-size:
 
 # sector-size = 8MiB, 1GiB available on the blob store
 download-params sector-size:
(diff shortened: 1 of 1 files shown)


aidan46 (approved):

aidan46:
Can this PR get merged or does it need #818 to get merged first?

th7nder:
This can be merged, #818 is coming after that one.
This is just to make sure we have a source of truth for params on the repo.

jmg-duarte (approved):

PR #821, opened 2025-04-08
feat(sp/node): resolve PeerId -> Multiaddr

jmg-duarte:
### Description

TL;DR: removes Rendezvous+Gossipsub in favour of Kadmelia

The key motivation is that we get more control over the addresses being added, though it brings more advantages.

The main issue with gossip is that it is timing dependent as it merely serves as a way to disseminate information, not to store it. A solution is simply to constantly re-gossip all info but that doesn't scale. A DHT fills that purpose much better as the information is split among multiple nodes.

While implementing this multiple issues arose:
* Kad doesn't work on the JS side
* Request/Response stopped working when the swarm building process was rewritten
* WASM libp2p works on a demo but not in the real thing

And even now I can foresee more problems rising due to the fact that this PR abuses the Kbuckets to fetch the registered multiaddresses. Currently we don't go out into the network but we need to, because we can't rely on a single peer to know everyone. With the upcoming [libp2p peerstore](https://github.com/libp2p/rust-libp2p/blob/1206fef09885d024323479d0383c37c4fe281c7c/misc/peer-store/src/lib.rs#L5), maybe we can have a slightly more solid way of keeping track of peers but for *right now*, this works.

Another thing we might try in the future is to use litep2p as a slightly higher-level approach to libp2p.

Demo:

https://github.com/user-attachments/assets/67be9bb8-08ec-4d71-92b0-4f42e64fc7b4

The HTTP error happens because we still need to resolve the other protocol (HTTP/JSON-RPC) ports, for which we need another request response.

### DIFF
--- a/node/src/service/p2p/bootstrap.rs
+++ b/node/src/service/p2p/bootstrap.rs
@@ -171,7 +171,7 @@ fn on_request_response_event(
                 log::trace!("Received a request-response request ({request_id}): {request:?}");
                 let Some(kref) = swarm.behaviour_mut().kad.kbucket(request.0) else {
                     log::trace!("KBucket query returned None, we're the node containing the peer");
-                    let peer_id = swarm.local_peer_id().clone();
+                    let peer_id = *swarm.local_peer_id();
                     let multiaddrs = swarm.external_addresses().cloned().collect();
                     let peer_info = PeerInfoResponse::Found(PeerInfo {
                         peer_id,
@@ -210,7 +210,7 @@ fn on_request_response_event(
                     return;
                 };
 
-                let peer_id = entry.node.key.preimage().clone();
+                let peer_id = *entry.node.key.preimage();
                 let multiaddrs = entry.node.value.clone().into_vec();
                 let response = PeerInfoResponse::Found(PeerInfo {
                     peer_id,
--- a/node/src/service/p2p/bootstrap.rs
+++ b/node/src/service/p2p/bootstrap.rs
@@ -193,10 +193,7 @@ fn on_request_response_event(
                 log::trace!("KBucket contains peer {}, searching for it", request.0);
                 let entry = kref
                     .iter()
-                    .filter(|entry| entry.node.key.preimage() == &request.0)
-                    .take(1)
-                    .collect::<Vec<_>>()
-                    .pop();
+                    .find(|entry| entry.node.key.preimage() == &request.0);
                 let Some(entry) = entry else {
                     log::trace!("Could not find peer {} in KBucket", request.0);
                     if swarm
--- a/node/src/service/p2p/bootstrap.rs
+++ b/node/src/service/p2p/bootstrap.rs
@@ -1,35 +1,28 @@
-use std::{
-    collections::HashMap,
-    hash::{DefaultHasher, Hash, Hasher},
-    io::{Error, ErrorKind},
-    time::Duration,
-};
+use std::time::Duration;
 
 use libp2p::{
     futures::StreamExt,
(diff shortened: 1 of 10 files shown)


node/src/service/p2p/bootstrap.rs
```diff
@@ -162,140 +135,86 @@ pub(crate) async fn bootstrap(
...
+                    let peer_info = PeerInfoResponse::Found(PeerInfo {
+                        peer_id,
+                        multiaddrs,
+                    });
+                    log::trace!("Sending response to request ({request_id}): {peer_info:?}");
+                    swarm
+                        .behaviour_mut()
+                        .request_response
+                        .send_response(channel, peer_info)
+                        .unwrap(); // TODO: remove
```
th7nder: nit

node/src/service/p2p/bootstrap.rs
```diff
@@ -162,140 +135,86 @@ pub(crate) async fn bootstrap(
...
+                log::trace!("Found entry in KBucket, sending response: {response:?}");
+                swarm
                     .behaviour_mut()
                     .request_response
                     .send_response(channel, response)
-                    .is_err()
-                {
-                    error!("Failed to send peer info to {peer:?}");
-                }
+                    .unwrap();
```
th7nder: nit

node/src/service/p2p/bootstrap.rs
```diff
@@ -162,140 +135,86 @@ pub(crate) async fn bootstrap(
...
-                        info!("Peer {id:?} found in registrations");
-                        PeerInfoResponse::Found(peer_info.clone())
-                    }
-                    None => {
-                        info!("Peer {id:?} not found in registrations");
-                        PeerInfoResponse::NotFound(id.into())
-                    }
+                log::trace!("Received a request-response request ({request_id}): {request:?}");
+                // When we receive a message from
+                let Some(kref) = swarm.behaviour_mut().kad.kbucket(request.0) else {
```
th7nder, line 172: What if the current peer doesn't contain the information, but some other one does?
How would we query the network or the info **should be** in kademlia?

If I understand correctly, now we're just exposing addresses of servers which directly connected to the Collator.

th7nder (changes requested):

node/src/service/p2p/bootstrap.rs
jmg-duarte: Great catch. I did that on purpose to minimize the amount of moving parts *right now*.

On failure to find we need to perform a get_closest_peers, which *should* add the peers it connects to, to the table. Once that query finishes, we search again

node/src/service/p2p/bootstrap.rs
```diff
@@ -162,139 +135,95 @@ pub(crate) async fn bootstrap(
...
+                    return;
+                };
+
+                log::trace!("KBucket contains peer {}, searching for it", request.0);
+                let entry = kref
+                    .iter()
+                    .filter(|entry| entry.node.key.preimage() == &request.0)
+                    .take(1)
+                    .collect::<Vec<_>>()
+                    .pop();
```
aidan46: Isn't this just [find](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) with extra steps?

jmg-duarte: It very much is, yes. I most likely tried to get the autocomplete, didn't work, resorted to this one. Great catch, clippy should be able to catch this though

th7nder (approved):

aidan46 (approved):

PR #823, opened 2025-04-10
feat(sp): add services req/resp protocol

jmg-duarte:
### Description

Adds a request/response libp2p service providing information about the non-libp2p services on the Storage Provider.

Delia demo:
https://github.com/user-attachments/assets/2d0a3cc4-7d72-43e1-a6d9-ec8c8ae2c6df

### Checklist

### DIFF
--- a/primitives-p2p/src/request_response/services.rs
+++ b/primitives-p2p/src/request_response/services.rs
@@ -1,0 +1,150 @@
+use std::collections::HashMap;
+
+pub const PROTOCOL_NAME: &str = "/polka-storage/rr-services/1.0.0";
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))]
+pub struct ServiceInfo {
+    pub port: u16,
+}
+
+/// Holds a map between protocol names to service information
+/// (e.g. `"ws": { port: 7008 }`).
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))]
+pub struct Services(pub HashMap<String, ServiceInfo>);
+
+impl Services {
+    pub fn get(&self, service: &String) -> Services {
+        Services(match self.0.get(service) {
+            Some(info) => {
+                let mut m = HashMap::new();
+                m.insert(service.clone(), info.clone());
+                m
+            }
+            None => HashMap::new(),
+        })
+    }
+
+    pub fn get_n<'s, I>(&self, services: I) -> Services
+    where
+        I: Iterator<Item = &'s String>,
+    {
+        let mut m = HashMap::new();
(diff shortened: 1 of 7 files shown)


primitives-p2p/src/request_response/services.rs
```diff
@@ -0,0 +1,150 @@
...
+            Some(info) => {
+                let mut m = HashMap::new();
+                m.insert(service.clone(), info.clone());
+                m
+            }
+            None => HashMap::new(),
+        })
+    }
+
+    pub fn get_n<'s, I>(&self, services: I) -> Services
```
Jinxit: I think a simpler API would be to _always_ request _all_ services and have it return essentially `self.0`. Is there a ever a reason to only request a single service?

Jinxit: I just noticed the `All` request 😬 But still, is there any reason for having the others? Just in the name of "less code is less problems"

storage-provider/server/src/main.rs
```diff
@@ -550,6 +551,22 @@ impl Server {
...
+                    ServiceInfo {
+                        port: self.rpc_listen_address.port(),
+                    },
+                );
+                hm.insert(
+                    "http".to_string(),
+                    ServiceInfo {
+                        port: self.upload_listen_address.port(),
+                    },
+                );
```
th7nder: this naming is a bit confusing here, why would `ws` mean `rpc` and `http` upload?

primitives-p2p/src/request_response/services.rs
```diff
@@ -0,0 +1,114 @@
...
+mod tests {
+    use std::collections::HashMap;
+
+    use serde_json::{json, to_string};
+
+    use super::ServiceInfo;
+    use crate::services::{Request, Response, Services};
+
+    #[test]
+    fn service_info_shape() {
```
th7nder, line 52: bit odd that we're using Cbor but testing json here. I guess this isn't that important as it **should work** anyhow.

th7nder (dismissed):

primitives-p2p/src/request_response/services.rs
jmg-duarte: I get you, but here's about the "shape". The cbor crates (AFAIK) don't provide a similar API to the JSON one

storage-provider/server/src/main.rs
jmg-duarte: I used those because they're the protocols underlying those services. I can change the naming

th7nder: i mean it should say what those are about functionally, not the underlying protocol.
on the other hand I don't remember what those are doing under the hood, so i'm fine. leaving it up to you.

th7nder (approved):

Jinxit (approved):

aidan46 (approved):

PR #825, opened 2025-04-14
feat(node): expose libp2p listen addrs

jmg-duarte:
### Description

This change enables Delia (with additional changes) to connect to the collator well known address and then the libp2p one.

### DIFF
--- a/node/src/rpc.rs
+++ b/node/src/rpc.rs
@@ -7,11 +7,15 @@
 
 use std::sync::Arc;
 
+use jsonrpsee::core::RpcResult;
+use libp2p::Multiaddr;
 use polka_storage_runtime::{opaque::Block, AccountId, Balance, Nonce};
 use sc_transaction_pool_api::TransactionPool;
 use sp_api::ProvideRuntimeApi;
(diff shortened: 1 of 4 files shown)


th7nder (approved):

node/src/rpc.rs
```diff
@@ -44,9 +51,47 @@ where
...
+    }
     Ok(module)
 }
+
+#[jsonrpsee::proc_macros::rpc(client, server)]
+trait PolkaStorageServicesApi {
+    /// Exposes the local listen multiaddresses. It will usually be either `0.0.0.0` or `127.0.0.1`.
+    ///
+    /// This is used by Delia to resolve the libp2p addresses for the request/response protocols.
+    #[method(name = "polkaStorage_getP2pMultiaddrs")]
```
aidan46, line 74: Why is this mixing camelCase and snake_case?

jmg-duarte: It's JSON-RPC nomeclature (it seems)
image

aidan46 (approved):

Issue #827, opened 2025-04-16
Storage Provider benchmarks: Run all relevant extrinsics on multiple sectors


Jinxit:
For now, we only have 1 example of a proven sector for the production 1GiB sector size, which means certain extrinsics' weights (`declare_faults`, `declare_faults_recovered`, `terminate_sectors`) aren't accurately scaling with the number of sectors.

When we have test data for more sectors (or figure out a good way to test it without them), the benchmarks need to be updated with `n: Linear<1, N>` parameters and re-run.

As the code is written right now we'd need to run separate PoRep proofs of the 1GiB sector data, which each takes 20 minutes, and right now `DECLARATIONS_MAX == 3000` and `MAX_TERMINATIONS_PER_CALL == 32` which means about 3 solid years of proving.

Once we figure out what would be reasonable values for those constants and how much would even fit in a block (using non-scaled weights as an initial estimate), then we can decide on a good limit to run it to and scale it linearly from there.

PR #828, opened 2025-04-16
test(sp): Benchmark `declare_faults`

Jinxit:
### Description

Requires #826 to be merged first.

This PR adds a benchmark for `declare_faults`, written in a way that it could eventually handle multiple sectors (when we have the data for it), but for now it just does a single sector and the benchmark doesn't scale at all.

I had to make the `error` and `fault` modules in `pallet-storage-provider` public, I hope that's ok.

When implementing the benchmark I got a rather unhelpful "GeneralPalletError" error, with no additional contents, so I wrote the function `assert_ok_sp` to decode that error into something more useful. Is this the way to do it, or does it already exist somewhere as a function in polkadot?

### Checklist

  - [x] Two previously private modules are now public.
  - [x] A question about pallet errors.
  - [x] #827
  - [x] No

### DIFF
--- a/pallets/storage-provider/benchmarks/src/benchmarking.rs
+++ b/pallets/storage-provider/benchmarks/src/benchmarking.rs
@@ -1,26 +1,32 @@
 #![cfg(feature = "runtime-benchmarks")]
 
-use alloc::vec;
+use alloc::{collections::BTreeSet, vec, vec::Vec};
 use core::ops::Add;
 
-use codec::Encode;
+use codec::{Decode, Encode};
 use frame_benchmarking::v2::*;
 use frame_support::{
     assert_ok,
-    pallet_prelude::{ConstU32, One},
+    pallet_prelude::{ConstU32, DispatchError, One},
     traits::{Currency, Get, Hooks},
     BoundedVec,
 };
 use frame_system::{pallet_prelude::BlockNumberFor, RawOrigin};
+use itertools::Itertools;
 use pallet_market::Pallet as MarketPallet;
 use pallet_proofs::Pallet as ProofsPallet;
-use pallet_storage_provider::{pallet::Call, Pallet as SpPallet};
+use pallet_storage_provider::{
+    error::GeneralPalletError,
+    fault::{DeclareFaultsParams, FaultDeclaration},
+    pallet::Call,
+    Pallet as SpPallet,
+};
 use primitives::{
     configs::BalanceOf,
     deals::ClientDealProposalOf,
     sector::{ProveCommitSector, SectorPreCommitInfo},
     test_data::{generate_benchmark_account, BenchmarkData},
-    MAX_SECTORS_PER_CALL, PEER_ID_MAX_BYTES,
+    MAX_SECTORS_PER_CALL, MAX_TERMINATIONS_PER_CALL, PEER_ID_MAX_BYTES,
 };
 use sp_runtime::{AccountId32, MultiSignature, MultiSigner};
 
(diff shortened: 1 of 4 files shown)


pallets/storage-provider/benchmarks/src/benchmarking.rs
```diff
@@ -84,7 +88,25 @@ mod benchmarks {
         #[extrinsic_call]
         _(RawOrigin::Signed(sp_id.clone()), prove_sectors.clone());
 
-        check_prove_commit_sectors::<T>(sp_id, prove_sectors, total_fee, 1);
+        check_prove_commit_sectors::<T>(sp_id, prove_sectors, total_fee);
+    }
+
+    /// `n`: number of submitted faulty sectors
+    // TODO(@Jinxit,15/04/2025): Use `n: Linear<1, DECLARATIONS_MAX * MAX_TERMINATIONS_PER_CALL>`
+    //                           when we have more proven sectors to use.
```
jmg-duarte: ~What's missing to have more proven sectors?~ https://github.com/eigerco/polka-storage/issues/827

But the issue would benefit from more info

pallets/storage-provider/benchmarks/src/benchmarking.rs
```diff
@@ -286,18 +359,14 @@ where
...
-    assert_ok!(SpPallet::<T>::register_storage_provider(
-        RawOrigin::Signed(provider.account_id.clone()).into(),
-        provider.peer_id.clone(),
-        post_proof,
-    ));
+/// Asserts that the result was Ok, and additionally decodes the error if it was a GeneralPalletError
+/// to get the inner error as a user-friendly string.
+fn assert_ok_sp(res: Result<(), DispatchError>) {
+    if let Err(sp_runtime::DispatchError::Module(module)) = &res {
+        if module.message == Some("GeneralPalletError") {
```
jmg-duarte:
```suggestion
        if let Some("GeneralPalletError") = module.message {
```

jmg-duarte (dismissed):

pallets/storage-provider/benchmarks/src/benchmarking.rs
Jinxit: With how we're doing it right now we'd need to run separate PoRep proofs of the 1GiB sector data, which each takes 20 minutes, and right now `DECLARATIONS_MAX == 3000` and `MAX_TERMINATIONS_PER_CALL == 32` which means about 3 solid years of proving 😅

Once we figure out what would be reasonable values for those constants and how much would even fit in a block (using these non-scaled weights as an initial estimate), then we can decide on a good limit to run it to and scale it linearly from there.

Jinxit: Added it to the issue.

pallets/storage-provider/benchmarks/src/benchmarking.rs
```diff
@@ -271,3 +358,15 @@ where
...
 }
+
+/// Asserts that the result was Ok, and additionally decodes the error if it was a GeneralPalletError
+/// to get the inner error as a user-friendly string.
+fn assert_ok_sp(res: Result<(), DispatchError>) {
+    if let Err(sp_runtime::DispatchError::Module(module)) = &res {
+        if let Some("GeneralPalletError") = module.message {
+            let variant = GeneralPalletError::decode(&mut [module.error[1]].as_slice()).unwrap();
+            panic!("{variant:#?}");
+        }
```
aidan46, line 369:
```suggestion
            if let Ok(variant) = GeneralPalletError::decode(&mut [module.error[1]].as_slice()) {
            panic!("{variant:#?}");
        }
```
Would this work, or is the `&str` check required for the decoding?

Jinxit: I think that _could_ decode without an error even if it wasn't a `GeneralPalletError`.

aidan46: That was what I was afraid of. Resolving this then

aidan46 (approved):

pallets/storage-provider/benchmarks/src/benchmarking.rs
```diff
@@ -85,6 +91,24 @@ mod benchmarks {
         check_prove_commit_sectors::<T>(sp_id, prove_sectors, total_fee);
     }
 
+    /// `n`: number of submitted faulty sectors
+    // TODO(@Jinxit,#827,15/04/2025): Use `n: Linear<1, DECLARATIONS_MAX * MAX_TERMINATIONS_PER_CALL>`
+    //                                when we have more proven sectors to use.
```
jmg-duarte, line 96:
```suggestion
    // when we have more proven sectors to use.
```

jmg-duarte (approved):

pallets/storage-provider/benchmarks/src/benchmarking.rs
Jinxit: I'll make sure any future TODOs are formatted as such.

PR #831, opened 2025-04-17
fix: Move constants back to pallets and fix module references

Jinxit:
### Description

When some of the pallet constants were moved to a trait in #818, which broke anything dependent on that metadata. I also forgot to regenerate the SCALE metadata file, which was fixed in #829, but that broke compilation due to outdated module references (deal structs were moved to `primitives`) and CI didn't run for the PR, so it ended up in develop.

This PR moves all except one constant back to the pallets, and instead exposes them as functions in the `Provider` traits. It also updates the SCALE metadata file and fixes the module references.

Still building on my machine, but looks good so far.

I've left the `MaxDeals` constant in `MarketProvider` because it's used in a const context (which trait functions cannot be used in) and I figured we'd solve the `develop` issue first and maybe move it back to the pallet later, if even necessary.

### Checklist

  - [x] Didn't run all build and tests yet, I'll let CI do it.
  - [x] `MaxDeals` remains where it was.
- ~~[ ] If there are follow-ups, have you created issues for them?~~
  - [x] Constant types in traits as it was before, which didn't work out in the end.


primitives/src/configs.rs
```diff
@@ -29,19 +29,21 @@ pub trait MarketProvider: frame_system::Config {
     type OffchainPublic: IdentifyAccount<AccountId = <Self as frame_system::Config>::AccountId>;
 
     /// How many deals can be published in a single batch of `publish_storage_deals`.
+    // TODO(@Jinxit,17/04/2025): Turn this into a function same like `min_deal_duration` and re-add
```
aidan46, line 32: ~~How come this is the only one not moved back?~~ I see, this value is only used for bounding array types

aidan46 (approved):

jmg-duarte (approved):

PR #837, opened 2025-04-24
test(sp): Benchmark `terminate_sectors`

Jinxit:
### Description

This PR adds benchmarks for the `terminate_sectors` extrinsic.

The benchmark requires exposing the `expiration_queue` and `sector` modules from the storage-provider pallet.

#827 applies here too.

### Checklist

  - [x] Two more modules become `pub`.
  - [x] See #827
  - [x] #827
  - [x] No

### DIFF
--- a/pallets/storage-provider/benchmarks/src/benchmarking.rs
+++ b/pallets/storage-provider/benchmarks/src/benchmarking.rs
@@ -17,9 +17,11 @@ use pallet_market::Pallet as MarketPallet;
 use pallet_proofs::Pallet as ProofsPallet;
 use pallet_storage_provider::{
     error::GeneralPalletError,
+    expiration_queue::ExpirationSet,
     fault::{
         DeclareFaultsParams, DeclareFaultsRecoveredParams, FaultDeclaration, RecoveryDeclaration,
     },
+    sector::{TerminateSectorsParams, TerminationDeclaration},
     Pallet as SpPallet,
 };
 use primitives::{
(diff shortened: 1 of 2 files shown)


pallets/storage-provider/benchmarks/src/benchmarking.rs
```diff
@@ -418,6 +438,80 @@ where
...
+    let (sp_id, prove_sectors, _) = prepare_prove_commit_sectors::<T>(n);
+
+    assert_ok_sp(SpPallet::<T>::prove_commit_sectors(
+        RawOrigin::Signed(sp_id.clone()).into(),
+        prove_sectors.clone(),
+    ));
+
+    let terminations = prove_sectors
+        .into_iter()
+        .chunks(MAX_TERMINATIONS_PER_CALL as usize)
```
th7nder, line 460: if we currently have 1 termination and only benchmark it for 1, shouldn't we change MAX_TERMINATIONS_PER_CALL to 1 as well?

Jinxit: I think I'll leave that up to you, I'm not 100% sure the logic is even correct in the first place but it's just to have _something_. I've got some upcoming work for actually scaling it which doesn't rely on having real proofs.

jmg-duarte (approved):
I'm a bit out of the loop for the sector termination (it has been a while) but the test looks ok, it's the logic I'd expect:
* Setup sectors
* Terminate them
* Test terminations

I'd say we can move forward with this right now

th7nder (approved):

PR #855, opened 2025-05-09
docs: add delia dapp docs

jmg-duarte:
Covers the Delia dApp "internals"


cernicc (approved):

docs/src/architecture/delia.md
aidan46, line 1: I think it is important to mention that both a storage provider and storage client need to deposit balance to the market pallet as this was an issue that Sergey ran into.

jmg-duarte: I can add, but notice that this IS NOT a guide on how to use Delia. For that, people should refer to Delia

aidan46 (approved):

PR #858, opened 2025-05-12
test(sp): Generate PoSt benchmark data

Jinxit:
### Description

PR 6 of many to introduce benchmarking of the storage-provider pallet's `submit_windowed_post` extrinsic.

This PR adds code for generating PoSt benchmark data and embeds it for 8MiB.

### Checklist

  - [x] None.
  - [x] This and upcoming PRs are a part of #739.
- ~~[ ] Were there any alternative implementations considered?~~

### DIFF
--- a/storage-provider/client/src/commands/proofs.rs
+++ b/storage-provider/client/src/commands/proofs.rs
@@ -550,7 +550,7 @@ fn post(
     sector_number: u32,
     comm_r: String,
     replica_path: PathBuf,
-    cache_directory: impl AsRef<Path>,
+    cache_directory: PathBuf,
     proof_parameters_path: PathBuf,
     post_type: RegisteredPoStProof,
 ) -> Result<(), CliError> {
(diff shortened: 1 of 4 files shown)


storage-provider/client/src/commands/proofs.rs
```diff
@@ -541,19 +544,16 @@ fn generate_post_params(
...
 fn post(
-    signer_key: MultiPairArgs,
+    signer: &MultiPairSigner,
     challenge_block: u64,
     output_path: Option<PathBuf>,
     sector_number: u32,
     comm_r: String,
     replica_path: PathBuf,
-    cache_directory: PathBuf,
+    cache_directory: impl AsRef<Path>,
```
cernicc: I think this was fine as a PathBuf. You are creating a PathBuf `cache_directory.as_ref().to_path_buf(),` bellow, so the allocation is there but hidden from the caller of the function. When you require a PathBuf in the signature. The caller is required to allocate a PathBuf so the allocation is not hidden.

What do you think?

Jinxit: The reason was [this](https://github.com/eigerco/polka-storage/pull/858/files/e322a408b28d6ac5aa6f20f1bf729df07bf6aa79#diff-358758f32b7b2a1a8319e10e612480d4626364d9b4a520f311ca83d3a964075dR793). When using a TempDir it requires `cache_directory.path().to_path_buf()`. Also just to match the `porep` signature for consistency.

But I don't think it matters either way, so I changed it back.

cernicc (approved):

jmg-duarte (approved):

PR #861, opened 2025-05-14
fix(sp): remove cli args

jmg-duarte:
### Description

Removes the CLI args, using only the config.

There's still a follow up coming to add a command to dump a default configuration.

### Checklist

### DIFF
--- a/storage-provider/common/src/config/sealing.rs
+++ b/storage-provider/common/src/config/sealing.rs
@@ -89,7 +89,6 @@ where
 
 /// Configuration for the sealing process.
 #[derive(Debug, Clone, Deserialize, Serialize)]
-// #[cfg_attr(feature = "clap", derive(::clap::Args))]
 pub struct SealingConfiguration {
     /// The percentage above which a sector is considered "full", defaults to 95% of the registered
     /// sector size.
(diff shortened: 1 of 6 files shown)


storage-provider/common/src/config/sealing.rs
```diff
@@ -89,7 +89,7 @@ where
 
 /// Configuration for the sealing process.
 #[derive(Debug, Clone, Deserialize, Serialize)]
-#[cfg_attr(feature = "clap", derive(::clap::Args))]
+// #[cfg_attr(feature = "clap", derive(::clap::Args))]
```
cernicc: Comment instead of removing. Intentional?

storage-provider/server/src/config.rs
```diff
@@ -107,35 +96,28 @@ pub struct ConfigurationArgs {
...
-    #[arg(long, required = false)]
     pub(crate) rendezvous_point_address: Multiaddr,
 
     /// PeerID of the rendezvous node used.
     #[serde(deserialize_with = "deserialize_string_to_peer_id")]
-    #[arg(long, required = false)]
     pub(crate) rendezvous_point: PeerId,
 
-    #[clap(flatten)]
+    // #[clap(flatten)]
```
cernicc: Same here

storage-provider/common/src/config/sealing.rs
jmg-duarte: No, I forgot to delete here

cernicc (approved):

aidan46 (approved):

PR #868, opened 2025-05-27
feat: calc collateral in the pallet

### LINKED ISSUE #835: Collateral being calculated client-side
Currently provider collateral is calculated client side and can be manipulated in Delia leading to some economic cheating incentives.
It should be calculated on the chain, in a pallet, not in Delia.

cernicc:
### Description

Currently, the user specifies the collateral. This change makes it so that the collateral is calculated by the pallet.

### Checklist


primitives/src/deals/deal_proposal.rs
```diff
@@ -85,4 +81,9 @@ where
         let commitment = Commitment::from_cid_bytes(&self.piece_cid[..])?;
         Ok(commitment)
     }
+
+    // The current provider collateral is specified as double the storage fee
+    pub fn provider_collateral(&self) -> Option<u128> {
```
Jinxit, line 86: Could we add a `Balance: From<BlockNumber>` requirement and have both `total_storage_fee` and `provider_collateral` return `Balance`s instead of `Option<u128>`s?

pallets/storage-provider/src/tests/submit_windowed_post.rs
```diff
@@ -34,7 +34,7 @@ fn setup() {
     // Add balance to the market pallet
     assert_ok!(Market::add_balance(
         RuntimeOrigin::signed(account(storage_provider)),
-        60
+        101
```
Jinxit: Doesn't match comment above.

Jinxit (changes requested):

primitives/src/deals/deal_proposal.rs
cernicc: I've created it like that, because I thought we'll slowly remove the generic types for BlockNumber and Balance. This was a step in that direction. The `u128` was used because the `total_storage_fee` was already returning it.

pallets/storage-provider/src/tests/submit_windowed_post.rs
cernicc: Good catch. Fixed

pallets/market/src/lib.rs
```diff
@@ -433,7 +433,11 @@ pub mod pallet {
             let caller = ensure_signed(origin)?;
 
             BalanceTable::<T>::try_mutate(&caller, |balance| -> DispatchResult {
-                ensure!(balance.free >= amount, Error::<T>::InsufficientFreeFunds);
+                if balance.free < amount {
+                    log::error!(target: LOG_TARGET, "withdraw_balance: not enough free balance {:?} < {:?}", balance.free, amount);
+                    fail!(Error::<T>::InsufficientFreeFunds);
+                }
+
```
jmg-duarte: Instead you could do the other hack:
```suggestion
				ensure!(balance.free >= amount, {
					log::error!(target: LOG_TARGET, "withdraw_balance: not enough free balance {:?} < {:?}", balance.free, amount);
					Error::<T>::InsufficientFreeFunds
				});
```

pallets/market/src/lib.rs
```diff
@@ -1474,9 +1497,20 @@ pub mod pallet {
                             proposal.provider,
                             deal_id
                         );
+
+                        let Some(provider_collateral) = proposal.provider_collateral() else {
+                            log::error!(target: LOG_TARGET, "on_finalize: invariant violated cannot calculate provider_collateral, deal {}", deal_id);
```
jmg-duarte: This should never be violated though, right? Shouldn't we panic on testnet builds here at least?

cernicc: I've handled the error here because the same one was handled above when using `total_storage_fee`. I think this can fail, if runtime is misconfigured and smaller than u128 number is used as a Currency.

primitives/src/deals/deal_proposal.rs
Jinxit: Honestly I hadn't considered just committing to a specific BlockNumber and Balance, but that's a great idea. With that in mind I'm fine with it as it is.

Jinxit (dismissed):

Jinxit (approved):

jmg-duarte (approved):

PR #869, opened 2025-05-28
chore: separate storage pallet logic to the modules

cernicc:
### Description

Move pallet logic from `lib.rs` to the modules. This will result in a cleaner and leaner pallet module, which will enable us to merge the pallets more easily.

### Checklist


Jinxit (approved):
Anything that gets code out of the pallets' `lib.rs` is a good change to me.

jmg-duarte (approved):

Issue #871, opened 2025-05-30
Use polkadot-omni-node instead of custom node


Jinxit:
If possible, we should be using the official [polkadot-omni-node](https://docs.polkadot.com/develop/toolkit/parachains/polkadot-omni-node/) to run our parachain runtime. This would allow us to completely delete the `node` folder from our repo, saving our IDEs and compilers a bunch of work.

This would necessitate moving the p2p bootstrap node to be a separate binary, which should be fine - there's no incentive to run the bootstrap node regardless of whether it's built into the collator or not.

PR #872, opened 2025-05-30
feat(runtime): Re-sync repo with stable2412 parachain template

Jinxit:
### Description

This brings in almost all (except [this](https://github.com/paritytech/polkadot-sdk/issues/6659)) changes from the parachain template we're using. Amongst those changes is a move of the template chain specs / config from the node (which would require a specialized node for each parachain) to the runtime (which means polkadot-omni-node can use it generically).

I haven't verified that we can run on polkadot-omni-node yet, if I remember correctly I hit a snag with the VRF, but I'll get back to that later.

### Checklist

  - [x] Polkadot-omni-node _maybe_ on the way, but not yet.
  - [x] #871
- [ ] ~~Were there any alternative implementations considered?~~


runtime/src/lib.rs
```diff
@@ -2,19 +2,19 @@
 // `construct_runtime!` does a lot of recursion and requires us to increase the limit to 256.
 #![recursion_limit = "256"]
 
-// Make the WASM binary available.
+extern crate alloc; // Make the WASM binary available.
```
cernicc:
```suggestion
extern crate alloc;

// Make the WASM binary available.
```

It's confusing if the comment is in the same line.

cernicc (commented):
nit

runtime/src/lib.rs
Jinxit: It isn't in my source locally, not sure what's going on there.

Jinxit: Re-fetched and got it locally now, pushing a fix.

Justfile
jmg-duarte, line 1: Why do you need the `-p polka-storage-node`? Did something break?

Cargo.toml
```diff
@@ -30,6 +30,30 @@ members = [
   "storagext/cli",
   "storagext/lib",
 ]
+# Same as above except no `node`.
```
jmg-duarte, line 33: Why is this needed?

jmg-duarte (approved):
lets ship this, and if anything comes up, we fix it later

Justfile
Jinxit: It's because as of stable2412, the [parachain template](https://github.com/paritytech/polkadot-sdk-parachain-template/commit/ba81070bf638426ac24c99083aa13f50c83fdc6a#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R9-R14) moved the node out of the workspace default-members list. This means any command invoking the node binary needs to explicitly state the crate it is located in, and it won't be built by default.

As for why it was removed in the first place, it's to save on compile time. Most parachains won't need their own custom node (they can use the omni-node) and as such have no need to build the node, but they left it in just in case someone wants to create their own node, I suppose.

Cargo.toml
Jinxit: See https://github.com/eigerco/polka-storage/pull/872#discussion_r2120996188.

cernicc (approved):

PR #883, opened 2025-06-09
feat(sp-client): deprecate wallet commands and remove sp-cli

jmg-duarte:
Deprecates the wallet command inherited from sc-cli — use subkey instead

~400 more deps out of the build process


cernicc (commented):
Let's add a `subkey` to the `buildInputs` in the `flake.nix`. I needed it for the paseo and this approach worked.

We've made a full circle. The wallet cli commands were one of the first things added to the project :D

cernicc (approved):

Jinxit (approved):

Issue #840, opened 2025-04-30
Add AutoNAT support to the Node and SP


jmg-duarte:
While the node is supposed to be publicly known, the same doesn't apply to the SP though it needs to be publicly *dialable*.

Adding AutoNAT should provide us with information on "is this address publicly dialable" as such, we can then add said address to the external addresses, which should be advertised to the rest of the network.

Given we're already running a separate libp2p swarm in the node, we can simply add the AutoNAT behaviour to the node and the SP. The idea being that when the SP connects to the node, they'll exchange AutoNAT "dialability" and then setup the DHT addresses correctly.

This should replace the "hack" introduced in #839

jmg-duarte:
If we're to move the bootstrap out of the node, AutoNAT should be added "as a server" to the bootstrap and "as a client" to the SP

jmg-duarte:
No longer relevant

PR #885, opened 2025-06-09
feat(bootstrap): add bootstrap node

jmg-duarte:
### Description

Adds the new bootstrap node so we can (in the future) move the P2P logic out of the collator node and leverage the omni node instead.

This brings more complexity to the network in favor of (hopefully) faster compilation times and a more standard setup for the parachain.


chain_spec.json
Jinxit, line 1: In my (unmerged) Paseo experiments I've put the new chain specs as `chainspecs/paseo.raw.json` and `chainspecs/paseo.plain.json`, but they are not committed since they are build artifacts rather than inputs. Is this chain spec intended to be committed?

Regardless, the Justfile is missing the command used to generate it, and it would be nice to have a similar naming structure to make it identifiable.

Justfile
```diff
@@ -252,3 +252,12 @@ generate-weights pallet steps="5" repeat="1":
         --repeat "{{repeat}}" \
         --template node/benchmark_template.hbs \
         --output "pallets/{{pallet}}/src/weights.rs"
+
+run-bootstrap:
+    # openssl genpkey -algorithm ED25519 -out /tmp/zombienet/charlie-private.pem
+    # openssl pkey -in /tmp/zombienet/charlie-private.pem -pubout -out /tmp/zombienet/charlie-public.pem # Generate public key so script can get the Peer ID
+    RUST_LOG=info,polka_storage_bootstrap=trace cargo r -r -p polka-storage-bootstrap -- \
```
Jinxit, line 259: Want to keep `trace` or drop it a bit?

storage-bootstrap/src/main.rs
```diff
@@ -0,0 +1,53 @@
...
+fn main() {
+    tracing_subscriber::registry()
+        .with(
+            fmt::layer().with_filter(
+                EnvFilter::builder()
+                    .with_default_directive(if cfg!(debug_assertions) {
+                        LevelFilter::DEBUG.into()
+                    } else {
+                        LevelFilter::INFO.into()
+                    })
```
Jinxit, line 35: Ooh, I like this.

Justfile
jmg-duarte: For this, it's ok for now

chain_spec.json
jmg-duarte: This wasn't supposed to be here, I'll remove

Jinxit (approved):

storage-bootstrap/src/cli/bootstrap.rs
```diff
@@ -0,0 +1,196 @@
...
+                SwarmEvent::ConnectionClosed { peer_id, .. } => {
+                    info!("Disconnected from {}", peer_id);
+                }
+                SwarmEvent::Behaviour(crate::behaviour::BehaviourEvent::Identify(event)) => {
+                    match event {
+                        identify::Event::Received {  peer_id, info, .. } => {
+                            tracing::trace!("Received an identify received event");
+                            for addr in info.listen_addrs {
+                                tracing::debug!("Adding address to kademlia: peer_id={peer_id}, addr={addr:?}");
+                                swarm.behaviour_mut().kad.add_address(&peer_id, addr);
```
aidan46, line 97: As far as I can tell this will always append the received peer information. Do we need to clear the old addresses before updating? Should there be some periodic garbage collection of old peer information? I realise that this is probably out of scope for this PR.

storage-bootstrap/src/behaviour.rs
```diff
@@ -0,0 +1,93 @@
...
+    public_key: PublicKey,
+}
+
+#[derive(NetworkBehaviour)]
+pub struct Behaviour {
+    pub identify: identify::Behaviour,
+    pub request_response: request_response::Behaviour<LpCbor<PeerIdRequest, PeerInfoResponse>>,
+    // TODO: replace the memory store with a persistent one and add caching
+    pub kad: kad::Behaviour<kad::store::MemoryStore>,
+    // TODO: add AutoNAT, this is the perfect node for that
```
aidan46, line 32: I see there is an open issue for adding autoNAT to the node #840. Is there an issue for replacing the memory store with a persistent one and adding caching too?

jmg-duarte: No, because caching may be implemented by default inside Kademlia, the persistent memory store though, would mean the peer is more resilient to faults like crashing since it can simply restart from it's last state

storage-bootstrap/src/cli/bootstrap.rs
jmg-duarte: I'll need to dig inside Kad to answer your question in full, but it should replace it in the routing table yes

aidan46 (approved):

PR #882, opened 2025-06-09
chore: merge market pallet into the storage-provider

cernicc:
### Description

This PR merges market pallet and storage-provider into a single pallet. The PR is huge. I am soo sorry 🙈 The change is tightly integrated with multiple parts of the system, so it was not possible to split it into multiple PRs. 

### Checklist

- [X] Make sure that you described what this change does.
- [X] Have you tested this solution?

PR #891, opened 2025-06-20
chore: merge market and sp pallets

jmg-duarte:
### Description

Supersedes #882, same goals, it was just easier to open a separate PR to ensure I didn't break any of the original PR


pallets/market/DESIGN.md
aidan46, line 1: Should we update and move this design doc or is it okay to delete? I guess the flow is documented in the book too

aidan46 (approved):

PR #896, opened 2025-06-26
fix(runtime): Fix duplicate balances in genesis

Jinxit:
### Description

After merging the market balances with pallet_balances, there's some duplicate accounts in the genesis. With this change, the accounts are stored in a BTreeMap, meaning the later (more specific) values override the earlier.

### Checklist

  - [x] No
  - [x] No
  - Still running locally, but wanted to put up the PR in the meantime.


aidan46 (approved):

aidan46 (changes requested):
The SCALE file changed after I generated it on this branch, breaking storagext in the process.

PR #899, opened 2025-06-27
fix(sp/client): move propose_deal from rpc to http

jmg-duarte:
### Description

As the title states


storage-provider/server/src/storage.rs
```diff
@@ -86,6 +101,7 @@ fn configure_router(state: Arc<StorageServerState>) -> Router {
                 )),
         )
         .route("/download/:cid", get(download))
+        .route("/api/v0/propose_deal", post(propose_deal))
```
Jinxit, line 104: This is new, should we standardize?

Jinxit (approved):

PR #900, opened 2025-06-30
feat: Remove BalanceEntry remains

Jinxit:
### Description

This PR removes the last remains of the BalanceEntry struct, and removes the deprecated market extrinsics from the storage-provider pallet.

Since transaction fees are now deducted from the same balance as we are using for our escrow and storage payments, I've opted to just give the test accounts the usual test default to keep things simple (2^60). It's significantly more than necessary, but I'd rather overshoot than try to work out the exact amount needed, which is pretty brittle anyways.

### Checklist

  - [x] Test genesis balances changed.
  - [ ] Using the Fungible traits. Starting on it straight away, so skipping the issue if that's ok.
  - [x] Working out exact amounts for balances felt too brittle.


pallets/storage-provider/src/lib.rs
jmg-duarte, line 1: Question for future reference, in production scenarios the indexes don't change, do they?

storagext/lib/src/clients/storage_provider.rs
```diff
@@ -146,10 +126,10 @@ pub trait StorageProviderClientExt {
     >;
 
     /// Retrieve the balance for a given account (includes the `free` and `locked` balance).
-    fn retrieve_balance(
+    fn retrieve_free_balance(
```
jmg-duarte, line 129: Can you double check if this is used and if not, delete it instead?

Let's remove whatever is not being used and we'll re-add if needed later

jmg-duarte (dismissed):

storagext/lib/src/clients/storage_provider.rs
Jinxit: Let's keep it in this PR, and it can get deleted in #899.

pallets/storage-provider/src/lib.rs
Jinxit: It's a breaking change to do so. Permitted, but requires a correct change to the runtime version.

aidan46 (approved):

jmg-duarte (approved):

PR #901, opened 2025-07-01
fix(sp/client): move publish_deal from rpc to http

jmg-duarte:
### Description

Move the publish_deal rpc to http, effectively removing JSON-RPC from the storage provider

Now that the SP has a single address, we can start removing other components that were added to solve for that issue specifically


storage-provider/server/src/main.rs
```diff
@@ -573,13 +548,6 @@ impl Server {
...
                 let mut hm = HashMap::new();
-                hm.insert(
-                    "rpc".to_string(),
-                    ServiceInfo {
-                        port: self.rpc_listen_address.port(),
-                        secure_url: self.public_secure_rpc_url,
-                    },
-                );
                 hm.insert(
                     "upload".to_string(),
```
Jinxit, line 552: I'm guessing this is next up for deletion?

jmg-duarte: Yes, right now it's still needed

Jinxit (approved):

PR #902, opened 2025-07-01
feat: prevent deregistration with pre-committed sectors

aidan46:
### Description

This PR adds a safety check to the `deregister_storage_provider` extrinsic to prevent an SP from deregistering while it still has pre-committed sectors.

Previously, deregistration only checked for active sectors with deals. However, an SP with pre-committed sectors has pending commitments in the system that must be resolved before deregistration is 'safe'.

### Checklist

  - [x] Adds a new error variant; please check naming and consistency.
  - [x] Ensures no deregistration with pending commitments.
  - [x] Added a new unit test.


jmg-duarte (changes requested):
New event -> new SCALE

jmg-duarte (approved):
LGTM

PR #903, opened 2025-07-01
fix: replace the PeerId on chain with the SP Multiaddr

jmg-duarte:
### Description

There are more changes to be done, the DHT isn't gone yet

Fetch is temporarily disabled to make this build without insane changes


Jinxit (dismissed):

Jinxit (dismissed):

aidan46 (commented):
I am not sure why this doesn't show up in the GitHub UI but when I check out the branch it does not include the change that adds a pre-committed sectors check for deregistration introduced in #902. The removal of this check is also shown in the diff when running `git diff origin/develop`.

aidan46 (approved):

PR #904, opened 2025-07-02
feat: remove the DHT and all related infra

jmg-duarte:
As the title states


examples/start_sp.sh
```diff
@@ -59,9 +59,9 @@ echo "seal_proof = '8MiB'
 post_proof = '8MiB'
 porep_parameters = 'target/params/8MiB.porep.params'
 post_parameters = 'target/params/8MiB.post.params'
-rendezvous_point_address = '$P2P_ADDRESS'
-p2p_key = '@$P2P_PRIVATE_KEY'
-rendezvous_point = '$P2P_BOOTSTRAP_PEER_ID'
+# rendezvous_point_address = '$P2P_ADDRESS'
+# p2p_key = '@$P2P_PRIVATE_KEY'
+# rendezvous_point = '$P2P_BOOTSTRAP_PEER_ID'
```
Jinxit: Leftovers

maat/src/lib.rs
```diff
@@ -113,15 +57,13 @@ pub fn local_testnet_config(temp_dir_path: &std::path::Path) -> NetworkConfig {
             parachain
                 .with_id(1000)
                 .cumulus_based(true)
+                .with_chain_spec_path("../chain_spec.json")
```
Jinxit, line 60: I think you can get zombienet to generate the chain spec for you (using the runtime), with one of the other `with_chain_spec_*` variants.

maat/src/lib.rs
```diff
@@ -149,10 +91,3 @@ where
     let keypair = Pair::from_string(s, None).unwrap();
     PairSigner::<PolkaStorageConfig, P>::new(keypair)
 }
-
-fn generate_pem_file<P: AsRef<Path>>(path: P) {
-    let signing_key = SigningKey::from([0; 32]);
-    let pem = signing_key.to_pkcs8_pem(LineEnding::default()).unwrap();
-    let mut file = File::create(path).unwrap();
-    write!(file, "{}", *pem).unwrap();
-}
```
Jinxit, line 158: Good riddance, honestly this part was the most confusing to get right when testing parachain deployment.

storage-provider/server/src/config.rs
```diff
@@ -29,15 +26,8 @@ fn default_node_address() -> Url {
...
-fn default_p2p_multiaddrs() -> Vec<Multiaddr> {
-    vec![
-        Multiaddr::from_str("/ip4/0.0.0.0/tcp/8002").expect("value should be a valid Multiaddr"),
-        Multiaddr::from_str("/ip4/0.0.0.0/tcp/8003/ws").expect("value should be a valid Multiaddr"),
-    ]
-}
-
 #[derive(Debug, Clone, Deserialize)]
-#[serde(deny_unknown_fields)]
+// #[serde(deny_unknown_fields)]
```
Jinxit: Leftover

Jinxit (changes requested):
Great work, just some leftovers to get rid of.

maat/src/lib.rs
Jinxit: I think it might already be doing that by default because of the `.cumulus_based(true)` 🤔

jmg-duarte: Without that, it doesn't work and I don't want to figure that out right now

Jinxit (approved):

PR #905, opened 2025-07-02
feat: Migrate from Currency to Fungible traits

Jinxit:
### Description

This is gonna need a rebase but I wanted to get it out for review anyways. I recommend reviewing the commits one at a time, they don't overlap.

This PR switches from the semi-deprecated [Currency](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/token/currency/index.html) traits to the newer [Fungible](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/token/fungible/index.html) traits. Functionality should be basically the same, with the only notable difference being that holds now have their reasons specified with individual buckets that can only be drained for the same reason as it was first held. The only exception is slashing which I've set to [forcibly remove](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/token/enum.Fortitude.html#variant.Force), i.e. also take from adjacent holds and freezes.

All the events that we briefly had from using Currency are now gone, if we want them back we have to implement the corresponding [`done_` events](https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_frame/token/fungible/trait.Mutate.html#method.done_transfer) - but we already have our own StorageProvider events so it feels unnecessary to me.

I also deleted the wrapper functions we had for locking/unlocking, since they just delegated to a single function call in the end (which are better documented than ours).

### Checklist

  - [x] Some events are now gone.
  - [x] No more wrapping functions for balance manipulation.
  - [x] None.
  - [x] Keeping the previous implementation.


jmg-duarte (dismissed):

aidan46 (commented):
Implementation LGTM. Shouldn't the scale file be updated since the pallet config changed?

Jinxit:
> Implementation LGTM. Shouldn't the scale file be updated since the pallet config changed?

Good catch!

aidan46 (approved):

PR #909, opened 2025-07-04
fix: unify storage provider benchmarks

jmg-duarte:
### Description

As the title states, unifies the storage provider benchmarks under the same pallet


pallets/storage-provider/src/benchmarks/benchmarking.rs
```diff
@@ -1,5 +1,7 @@
 #![cfg(feature = "runtime-benchmarks")]
 
+extern crate alloc;
+
```
Jinxit, line 4: This is probably not necessary now that they're merged, see if it works without.

pallets/storage-provider/src/benchmarks/test_data/sector_timeline.rs
```diff
@@ -220,8 +212,4 @@ where
         // Must be after the deadline, but there is no strict delay.
         self.deadline_start() + Relative::one()
     }
-
-    pub fn deadline_close(&self) -> Absolute<BlockNumber> {
-        self.deadline_start() + self.challenge_window()
-    }
```
Jinxit: This one has some logic so I'd like to keep it, even if we're not using it right now.

pallets/storage-provider/src/lib.rs
```diff
@@ -11,6 +11,9 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+#[cfg(any(feature = "runtime-benchmarks", test))]
```
Jinxit: Is this the right condition? It's only `#[cfg(feature = "runtime-benchmarks")]` in the other pallets.

pallets/storage-provider/src/benchmarks/benchmarking.rs
jmg-duarte: Still needed

Jinxit (approved):

PR #913, opened 2025-07-10
fix(sp): replace Arc<PathBuf> with PathBuf

jmg-duarte:
### Description

Definitely not one that @Jinxit will be a fan of but I spotted this while I was looking into adding SQLite

There's no point in adding the Arc indirection here since we're always allocating a new path with the `.join`


Jinxit (approved):

Jinxit:
I don't mind honestly, it has its place but this ain't it.

PR #914, opened 2025-07-11
fix(sp): refactor config and startup

jmg-duarte:
### Description

One of the reasons we had two configurations for the server configurations had to do with random path defaults, this PR addresses that and tries to simplify the setup ceremony

### DIFF
--- a/storage-provider/server/src/config.rs
+++ b/storage-provider/server/src/config.rs
@@ -1,15 +1,18 @@
 use std::{
+    env::temp_dir,
     net::{IpAddr, Ipv4Addr, SocketAddr},
     num::NonZero,
-    path::PathBuf,
+    path::{Path, PathBuf},
+    sync::OnceLock,
 };
 
 use polka_storage_provider_common::config::sealing::SealingConfiguration;
 use primitives::proofs::{RegisteredPoStProof, RegisteredSealProof};
+use rand::Rng;
 use serde::Deserialize;
 use url::Url;
 
-use crate::DEFAULT_NODE_ADDRESS;
+use crate::ServerError;
 
 /// Default address to bind the RPC server to.
 const fn default_upload_listen_address() -> SocketAddr {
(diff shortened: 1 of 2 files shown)


storage-provider/server/src/config.rs
```diff
@@ -22,28 +25,62 @@ const fn default_parallel_prove_commits() -> NonZero<usize> {
...
 fn default_node_address() -> Url {
     Url::parse(DEFAULT_NODE_ADDRESS).expect("DEFAULT_NODE_ADDRESS must be a valid Url")
 }
 
+static RANDOM_PATH: OnceLock<PathBuf> = OnceLock::new();
+
+/// Generates a random temporary directory.
+///
+/// This function will always yield the same random path for a *single program execution*.
+fn default_random_directory() -> PathBuf {
```
Jinxit, line 40: I think we should use use [tempdir](https://docs.rs/tempdir/latest/tempdir/) instead of reinventing the wheel.

Jinxit: (I know this is just moving code around, but might as well touch it up)

storage-provider/server/src/main.rs
```diff
@@ -202,361 +151,321 @@ pub struct ServerCli {
...
+
+fn main() -> Result<(), ServerError> {
+    // Metrics initialization.
+    let builder = PrometheusBuilder::new();
+    let exporter_handle = builder
+        .install_recorder()
+        .expect("Failed to install metrics recorder");
+
+    // Logger initialization.
+    let file_appender = tracing_appender::rolling::daily("logs", "sp_server.log");
```
Jinxit, line 433: Should probably initialize the logger before the metrics.

storage-provider/server/src/config.rs
aidan46: We don't even need to import the `tempdir` library. We can use `tempfile` (already imported) which already includes a [tempdir](https://docs.rs/tempfile/latest/tempfile/fn.tempdir.html) function.

storage-provider/server/src/main.rs
jmg-duarte: Doesn't matter in this case, they're not related

Jinxit (approved):

PR #922, opened 2025-07-21
feat(mater): support UnixFS wrapping

jmg-duarte:
### Description

Closes #678 

Wraps files in UnixFS information — when using (better) libraries, it allows the extractor to know the filename beforehand and write the file to the FS correctly.

I didn't add extraction support because it requires further re-architecture of the extraction work and maybe of the library API — in other words, I don't think it's worth the effort.

### DIFF
--- a/mater/cli/src/main.rs
+++ b/mater/cli/src/main.rs
@@ -29,6 +29,10 @@ enum MaterCli {
         /// If enabled, the output will overwrite any existing files.
         #[arg(long, action)]
         overwrite: bool,
+
+        /// If enabled, the file will be wrapped with UnixFS information.
+        #[arg(long, action)]
+        wrap: bool,
     },
     /// Convert a CARv2 file to its original format
     Extract {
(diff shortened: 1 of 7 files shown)


mater/cli/src/main.rs
```diff
@@ -29,6 +29,10 @@ enum MaterCli {
         /// If enabled, the output will overwrite any existing files.
         #[arg(long, action)]
         overwrite: bool,
+
+        /// If enabled, the file will be wrapped with UnixFS information.
+        #[arg(long, action)]
+        wrap: bool,
```
Jinxit, line 35: Should this be the default?

jmg-duarte: `wrap` should be false by default to keep the previous behavior — `action` example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3a281a51dfa650d6ab8635445d1ff381

Jinxit: That part I understand, but my question was more like: Do we want to keep the previous behaviour as the default or is this a better default?

jmg-duarte: Doesn't really matter right now to be honest. This was more of a "I know how to solve this issue" solution. 

Plus I prefer this flag to the `no-wrap` that go-car has

Jinxit (approved):

PR #923, opened 2025-07-22
chore: remove redundant crates & unify common benchmark primitives

jmg-duarte:
### Description

* Removes a bunch of crates that weren't being used or could be replaced with existing libraries (like hyper)
* Merges the benchmark common code


lib/polka-storage-proofs/src/porep/sealer.rs
aidan46, line 1: Why did you change the `From/.into()` conversions to explicit struct literals (`filecoin_proofs::UnpaddedBytesAmount(...)` and `filecoin_proofs::PieceInfo { … }`)? Was there an issue with the blanket implementations?

jmg-duarte: They were pulling in the whole FC crate just for those conversion functions.

aidan46 (approved):

